PR #42, opened 2015-09-11
Fault handling and recovery improved

AlessandroA:
This PR suggests improvements for the fault handling and recovery in uVisor.

**The Problem**
Fault handling is hardware specific when it comes to memory accesses, since faults on devices with the ARM MPU result in MemManage Faults, which are generally easy to recover, while the Freescale MPU triggers bus faults, which can be of different classes.

**The Solution**
- Different cases for bus faults introduced for K64F:
    - `vmpu_fault_recovery_mpu(pc, sp)` for Freescale MPU faults (hw-specific)
    - `vmpu_fault_recovery_bus(pc, sp)` for special registers that cannot be given unprivileged access in any other way (observed on MK64F, even if not documented)
- Bus/MemManage faults re-structured to systematically test all cases and halt on failure
- *Special* bus fault recovery now supports imprecise bus faults

**Additional Notes**
The `vmpu_fault_recovery_bus(pc, sp)` function allows to recover from precise and imprecise bus faults if they occurred in a format that follows a very precise pattern:
```asm
str r1, [r0]
dsb
nop
nop
nop
```
where the `str` instruction can be replaced by `strb`, `strh`, `ldr`, `ldrb`, `ldrh` and the `nop`s are only used if imprecise bus faults are enabled. The code is written in a way that little or no overhead is added wrt to the precise-fault-only implementation.

`FIXME`: ACL checks to be added to the `vmpu_fault_recovery_bus` function
`TODO`: `vmpu_fault_recovery_mpu(pc, sp)` to be implemented for both targets

### DIFF
--- a/core/mbed/uvisor-lib/secure_access.h
+++ b/core/mbed/uvisor-lib/secure_access.h
@@ -17,6 +17,8 @@
 #ifndef __SECURE_ACCESS_H__
 #define __SECURE_ACCESS_H__
 
+#include "uvisor-lib/vmpu_exports.h"
+
 /* the switch statement will be optimised away since the compiler already knows
  * the sizeof(type) */
 #define ADDRESS_WRITE(type, addr, val) \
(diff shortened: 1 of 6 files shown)


miklis:

PR #43, opened 2015-09-15
core: avoid bashisms like '<<<'

jserv:
When building with GNU make, it uses sh as the shell, which is dash
in Ubuntu. It is quite likely to be bash in other versions of GNU/Linux.
As a result, we should avoid bashisms (like <<<) and only use strict
POSIX syntax.


miklis:

AlessandroA:
Hello @jserv , thank you for your contribution!

Could you please sign the [contributor agreement](https://developer.mbed.org/contributor_agreement/)?

jserv:
@AlessandroA : got it. Signed.

AlessandroA:
Thank you!
+1 
Merging.

PR #49, opened 2015-09-17
mpu: use clz to find highest bit

jserv:
For ARMv7-M, GCC's __builtin_clz generates efficient clz instruction
without loops.


miklis:

PR #50, opened 2015-09-21
Improved generality and portability of SVCall macros

AlessandroA:
This PR is a preliminary set of changes that will enable the upcoming re-writing of inline assembly code. Changelog:

- A single SVCall macro introduced, which is easier to maintain for portability and can be used for all cases (direct SVC or secure gateways)
- All macros that were used to enable macro overloading are now made general and can be used in similar situations (for example, in `uvisor-lib/box-config.h`)
- Secure gateway changed to use these new macros

Next steps:

- apply all macros where relevant (especially for all SVCalls in uvisor-lib and when custom macro overloading was duplicated)


miklis:

PR #51, opened 2015-09-21
simplified vmpu_bits

meriac:
The special case of comparing zero is not necessary
- it's the exception under normal operations, so a speed-up is not justified for size==0. Instead the overhead of the comparison is applied on all the more common cases.
- size==zero has 32 leading zeros: (32-32)==0 which correctly results in zero bits.

### DIFF
--- a/core/system/inc/mpu/vmpu_exports.h
+++ b/core/system/inc/mpu/vmpu_exports.h
@@ -148,7 +148,7 @@ typedef struct
 
 static inline int vmpu_bits(uint32_t size)
 {
-    return (size == 0) ? 0 : (32 - __builtin_clz(size));
+    return 32 - __builtin_clz(size);
 }
 
 #endif/*__VMPU_EXPORTS_H__*/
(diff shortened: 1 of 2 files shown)


miklis:

meriac:
@AlessandroA, @jserv - simplified vmpu_bits

jserv:
`__builtin_clz(0)` is undefined and architecture-specific. But, as far as the current implementation is dedicated to ARM Cortex-M series, I agree with @meriac . Let's simplify.

AlessandroA:
+1 for the simplification. Probably a comment on the fact that `__builtin_clz(0)` is arch-specific could help

meriac:
I will add some tests that will cover these assumptions in form of debug-only asserts

miklis:

miklis:

miklis:

AlessandroA:
+1

miklis:

PR #52, opened 2015-09-22
Replace all inline asm in uvisor-lib

AlessandroA:
This PR replaces all occurrences of inline `asm` in uvisor-lib with macros that allow the assembly to be defined, changed and ported to other toolchains more easily.

In particular:
- All SVCall-based APIs (`vIRQ`, benchmarking, error) bodies have been replaced with the `UVISOR_SVC(...)` variadic macro
    - in addition `vIRQ` APIs have been simplified to remove one level of indirection
- The low-level functions for special memory accesses (see https://github.com/ARMmbed/mbed-hal-k64f/pull/6) have been replaced with assembly templates (`UVISOR_ASM_MEMORY_ACCESS(...)`) that are maintained in a separate header file
- Existing support variadic macros have been updated to be more flexible (you can specify the type of the inline `register asm(...)` constructs)

Next step: port the (now very few) inline `asm`s to armcc

### DIFF
--- a/core/mbed/source/benchmark.cpp
+++ b/core/mbed/source/benchmark.cpp
@@ -20,39 +20,22 @@
 void uvisor_benchmark_configure(void)
 {
     /* in sequence:
-     *   - configure the DWT (default overhead = 0)
+     *   - initialize the benchmarking unit (default overhead = 0)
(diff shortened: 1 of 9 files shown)


miklis:

miklis:

PR #56, opened 2015-10-01
Release process generates object file

AlessandroA:
This PR changes the `uvisor` ---> `uvisor-lib` release process. In particular, the asm + binary blob mechanism is replaced with a single object file, `uvisor.o`:
- the file is generated taking the asm + binary blob inputs (like in the old mechanism) and compiling them into an object file; the object file is the released product and gets published to `uvisor-lib`
- `uvisor.o` is toolchain-independent (target-specific support is still needed, for example for scatter file symbols aliasing in armcc), but still hardware-specific (we could reduce as a next step the hardware dependency to a platform dependency)


miklis:

miklis:

AlessandroA:
@meriac

meriac:
Can you please squash the identical commit messages together ?

AlessandroA:
Done @meriac

miklis:

Issue #34, opened 2015-07-30
Method to detect whether current code is being executed as a result of an IRQ


bremoran:
For some library functions, it is important, for fault detection, to know whether the API has been called as part of the call tree of an IRQ.

Examples:
* Memory allocators
* Some scheduler functions, e.g. ```postCallbackAgain()```

It may be adequate to document that these should not be done, but runtime checks are more direct.

AlessandroA:
Implemented in https://github.com/ARMmbed/uvisor/issues/57

PR #57, opened 2015-10-02
Add API to tell if user code is in active IRQn

AlessandroA:
This introduces a new user API:

`int vIRQ_GetLevel(void)`

that returns:
- `-1`  if no interrupt is active
- `>= 0` if an interrupt is active and it is being served;
       the actual return value indicates the priority level of the IRQn
       as set by default or with `vIRQ_SetPriority(IRQn_Type irqn)`


miklis:

AlessandroA:
@meriac 
@bremoran please tell me if this implements what you need, and if it solves https://github.com/ARMmbed/uvisor/issues/34

AlessandroA:
@hugovincent adding you as you were involved in the original discussion. Feel free to comment

core/mbed/source/interrupts.cpp
```diff
@@ -106,3 +106,38 @@ uint32_t vIRQ_GetPriority(uint32_t irqn)
...
+ * a preprocessor define in a uniform way it will be deleted from here */
+#define UVISOR_NVIC_IRQn_OFFSET 16
+
+int vIRQ_GetLevel(void)
+{
+    /* if uVisor is disabled we use the standard priority levels */
+    if (__uvisor_mode == 0) {
+        /* check if an IRQn is active (an ISR is being served) */
+        uint32_t ipsr = __get_IPSR();
+        int irqn = (int) (ipsr & 0x1FF) - UVISOR_NVIC_IRQn_OFFSET;
```
meriac: Ensure that mask is appropriate for selected target. The NVIC->IABR register (as use in NVIC_GetActive) is only 8 DWORD wide - which leaves only 256 interrupts. 0x1FF would reult in a overflow.

core/system/src/unvic.c
```diff
@@ -196,6 +196,31 @@ uint32_t unvic_irq_priority_get(uint32_t irqn)
...
 
+int unvic_irq_level_get(void)
+{
+    /* get and validate the unprivileged exception stack frame */
+    uint32_t *box_sp = svc_cx_validate_sf((uint32_t *) __get_PSP());
+
+    /* gather IPSR from exception stack frame */
+    /* the currently active IRQn is the one of the SVCall, while instead we want
+     * to know the IRQn at the time of the SVCcall, which is saved on the stack */
+    uint32_t ipsr = box_sp[7];
```
meriac: please drop svc_cx_validate_sf and use unprivileged read instead

core/system/src/unvic.c
```diff
@@ -196,6 +196,31 @@ uint32_t unvic_irq_priority_get(uint32_t irqn)
...
+    uint32_t irqn = (ipsr & 0x1FF) - IRQn_OFFSET;
```
meriac: 0x1FF results in an overflow of NVIC_GetActive

miklis:

AlessandroA:
@meriac 

- removed use of `svc_cx_validate_sf()`
- added check on `irqn` (fail if `>= HW_INTERRUPTS`)
    - note: this check is included in `unvic_check_default()` in uVisor, and is added explicitly in the mbed-facing fallback version (for when uVisor is disabled)

PR #61, opened 2015-10-06
ensure that MMARVALID is kept in case no known region has been hit

meriac:
@AlessandroA


miklis:

AlessandroA:
+1

PR #62, opened 2015-10-07
Add uvisor_{read,write}{32,16,8} to fallback APIs

AlessandroA:
`ADDRESS_{READ,WRITE}` macro are less meaningful and will only be used for massive header files (see the mighty PR for K64F macros); generally code should use these functions (`uvisor_{read,write}{32,16,8}`) to access special registers.

**Note**: this is temporary; when register-level access will be enabled there won't be need for these APIs


AlessandroA:
This fixes the errors on the unsupported platforms. @bogdanm @0xc0170 
@meriac

miklis:

bogdanm:
+1

PR #80, opened 2015-10-27
Verbose debug messages commented out

AlessandroA:
FIXME: will be re-introduced with verbosity levels for uVisor debugging


AlessandroA:
@meriac

PR #85, opened 2015-10-28
Explicitly test for YOTTA_CFG_UVISOR_PRESENT

AlessandroA:
This is safer as `SYMBOL` re-definition would always lead `#if defined(SYMBOL)` to be resolved in "true"

@meriac 
@bogdanm


bogdanm:
+1

PR #86, opened 2015-10-28
Fixing Error on "path/to/uvisor/release/uvisor-lib" does not exist.

JaredCJR:
When the current directory is at "path/to/uvisor/stm32f4/uvisor",the command "make release" will output the error messages like "/release/uvisor-lib" does not exist.

### DIFF
--- a/core/Makefile.rules
+++ b/core/Makefile.rules
@@ -186,6 +186,7 @@ release: $(MBED_ASM_INPUT) $(PROJECT).bin
 	rm  -f $(RELEASE_SRC)/*.cpp
 	rm -rf $(RELEASE_SRC_HW)
 	mkdir -p $(MBED_SRC_HW)
+	mkdir -p $(RELEASE_INC)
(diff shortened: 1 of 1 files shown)


miklis:

miklis:

meriac:
Thanks a lot for spotting and fixing!

AlessandroA:
@JakeCJR please keep in mind that while this PR fixes the issue, the released module will still be incomplete and unusable as a yotta module. `uvisor-lib`, in fact, is made by both code that we generate from `uvisor` (`make clean release`) and by yotta-specific files (like `CMakeLists.txt`, `module.json`, etc.). These latter are not automatically generated (yet) by the `release` rule.

The solution to this is to always clone `uvisor` recursively:
```
git clone --recursive git@github.com:ARMmbed/uvisor.git
```
which also clones `uvisor-lib` as a submodule, creating the release folder and populating it. The `make clean release` rule will then add the auto-generated files to the existing yotta module skeleton.

Issue #96, opened 2015-10-29
Bus fault recovery does not clear the BBFAR register


AlessandroA:
It is a similar issue to what was previously solved for the MMFAR register, for MemManage faults.

rainierwolfcastle:
ARM Internal Ref: IOTSFW-1129

PR #97, opened 2015-10-29
Fixed bug for bus fault acknowledgment

AlessandroA:
Acknowledging the bus fault it is possible to recover from it and be ready for a new one in the future -- otherwise the syndrome register retains the old value.

Fixes https://github.com/ARMmbed/uvisor/issues/96
@meriac

### DIFF
--- a/core/system/src/mpu/vmpu.c
+++ b/core/system/src/mpu/vmpu.c
@@ -204,7 +204,7 @@ uint32_t vmpu_register_gateway(uint32_t addr, uint32_t val)
     return 0;
 }
 
-int vmpu_fault_recovery_bus(uint32_t pc, uint32_t sp)
+int vmpu_fault_recovery_bus(uint32_t pc, uint32_t sp, uint32_t fault_addr, uint32_t fault_status)
 {
     uint16_t opcode;
     uint32_t r0, r1;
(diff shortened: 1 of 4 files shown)


AlessandroA:
@meriac updated after review

core/system/src/mpu/vmpu.c
```diff
@@ -216,11 +218,18 @@ int vmpu_fault_recovery_bus(uint32_t pc, uint32_t sp)
...
        HALT_ERROR(NOT_ALLOWED, "This is not the PC (0x%08X) your were searching for", pc);
 
-    /* if the bus fault is imprecise we will seek back for ldrX/strX opcode */
-    if(SCB->CFSR & (1 << 10))
-        cnt_max = UVISOR_NOP_CNT;
-    else
+    /* check fault register; the following two configurations are allowed:
+     *   0x04 - imprecise data bus fault, no stacking/unstacking errors
+     *   0x82 - precise data bus fault, no stacking/unstacking errors */
+    if (SCB_BFSR == 0x82) {
```
meriac: As both branches of the if use SCB_BFSR : it might be more readable to use a switch statement ?

core/system/src/mpu/vmpu.c
```diff
@@ -317,6 +326,9 @@ int vmpu_fault_recovery_bus(uint32_t pc, uint32_t sp)
     /* note: we assume the instruction is 16 bits wide and skip possible NOPs */
     vmpu_unpriv_uint32_write(sp + (6 << 2), pc + ((UVISOR_NOP_CNT + 2 - cnt) << 1));
 
+    /* clear BFSR only - reset BFARVALID bit to acknowledge fault */
+    SCB_BFSR = 0x80;
```
meriac: SCB_BFSR needs to be cleared even in case the function returns - we might want to consider in this function to clear it right after checking it at the top.

AlessandroA:
@meriac Updated PR. Changelog now:
- moved check on fault registers right at the fault handler: create backup copy, clear flag, use backup copy in recovery functions
- removed checks from other functions (these resulted in removing one function, which is not needed anymore)

Issue #99, opened 2015-10-30
Use of unvic_acl_check for interrupt context switching is too restrictive


AlessandroA:
In https://github.com/ARMmbed/uvisor/blob/master/core/system/src/unvic.c#L265 the check on ACLs for the context switch is too restrictive. The implementation of the function blocks cross-box interrupts since it requires the destination box to be the current box.

To fix this:
- add a `box_id` arg to `unvic_acl_check()`
- use `dst_id` as trigger for the ACL check in interrupt context switch. `g_active_box` is used for all other cases

PR #100, opened 2015-10-30
Add box id information for unvic_acl_check

AlessandroA:
The `unvic_acl_check` function (`unvic.c` scope only) checks if accesses to the virtual NVIC module is allowed. It always used `g_active_box` to tell if the access is allowed, which is too restrictive for the cross-box interrupt context switch.

In that case the information for the context switch comes from trusted internal uVisor tables, for which only a basic ACL check is needed.

Fixes https://github.com/ARMmbed/uvisor/issues/99

@meriac

### DIFF
--- a/core/system/src/unvic.c
+++ b/core/system/src/unvic.c
@@ -82,7 +82,7 @@ static void unvic_acl_check(int irqn)
      *       box claimed exclusive ownership for it. So, another box can claim it
      *       if it is currently un-registered (that is, if the registered handler
      *       is NULL) */
-    if(uv->id != svc_cx_get_curr_id() && (uv->id || uv->hdlr))
+    if(uv->id != g_active_box && (uv->id || uv->hdlr))
     {
         HALT_ERROR(PERMISSION_DENIED,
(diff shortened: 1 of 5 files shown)


core/system/src/unvic.c
```diff
@@ -94,7 +94,7 @@ static void unvic_acl_check(int irqn)
 void unvic_isr_set(uint32_t irqn, uint32_t vector, uint32_t flag)
 {
     /* verify IRQ access privileges */
-    unvic_acl_check(irqn);
+    unvic_acl_check(irqn, g_active_box);
```
meriac: What's the reason for adding a g_active_box parameter versus directly accessing g_active_box in unvic_acl_check?

AlessandroA: Because although most of the time it is `g_active_box`, there are special cases in which it is not. An important example: cross-box IRQ handling (see https://github.com/ARMmbed/uvisor/pull/100/files#diff-8079bf2b7ed17a83b95fadc1b10f7625R268), where `dst_id` is used instead. It is related to https://github.com/ARMmbed/uvisor/issues/99

AlessandroA:
@meriac updated after review

PR #101, opened 2015-10-30
Add bitbanding support

AlessandroA:
This PR introduces macros to translate bitbanded aliases to their physical address counterpart.

The macros are then used in fault recovery to use the physical address as argument to search for ACLs

@meriac


AlessandroA:
@meriac switched `PERIPH` and `SRAM` bitband checks in recovery functions

PR #105, opened 2015-11-03
Move fault status register clearing to end of handler

AlessandroA:
This allows the fault-related registers to be still valid for the rest of the handler execution, where
they could be used for debugging purposes. The flags are always cleared right before returning from
the exception handler.

@meriac


AlessandroA:
Bump

PR #111, opened 2015-11-27
Export uvisor_init with C-compatible binding

Patater:
When we replaced all inline assembly with macros, we accidentally removed
UVISOR_EXTERN from the unsupported-target implementation of uvisor_init.
This broke the build of uvisor on unsupported targets. Restore
UVISOR_EXTERN on the uvisor_init implementation to ensure uvisor_init is
exported with a C-compatible binding.

Fixes: 314207ac03f3 ("All explicit inline assembly replaced with macros")


AlessandroA:
+1

Issue #87, opened 2015-10-28
Context switch bug


meriac:
- Fix the interrupt bug related to tailchained IRQ that happen during a context switch.
- Affects EFM32GG

Acceptance criteria:
- create test to reproduce the behaviour
- verify the bug if fixed in the context of the test

rainierwolfcastle:
ARM Internal Ref: IOTSFW-1124

AlessandroA:
This bug has been addressed with https://github.com/ARMmbed/uvisor/issues/113.

The problem was not related to interrupts tailchaining as it was thought at the beginning. Instead, a particular interrupt configuration allowed an IRQ to be enabled (`vIRQ_EnableIRQ`) even before being registered (`vIRQ_SetVector`).

Although this is a pattern that is not recommended for NVIC interrupts, some missing checks in uVisor made it hard to detect and debug. With https://github.com/ARMmbed/uvisor/issues/113 those checks are now in place, and an error is issued when NVIC flags for an IRQ are changed before registering the corresponding ISR.

@akselsm As for the EFM32GG, the problem was a missing interrupt registration in the GPIO HAL initialization. As described in the `uvisor-lib` [documentation](https://github.com/ARMmbed/uvisor-lib#interrupt-management), interrupts need to be registered before being enabled, as without registration uVisor cannot ensure ownership and protection from other secure boxes.
```C
vIRQ_SetVector(SOME_IRQn, (uint32_t) &SOME_IRQHandler);
vIRQ_EnableIRQ(SOME_IRQn);
```

PR #113, opened 2015-11-27
Improve security checks on vIRQ APIs

AlessandroA:
The `unvic_acl_check(irqn)` function now returns whether `irqn` is registered or not.

This information is used by specific APIs to decide whether a state can be changed
for a certain interrupt. In particular:
- setting/clearing pending status for IRQs is only possible when they are registered
- enabling/disabling IRQs is possible even if IRQn is unregistered but only if
  this does not change the current state (enabled/disabled); this loosens a bit
  the requirement on function calls order for vIRQ APIs
- setting the priority of unregistered IRQs will always fail

The consequences of these changes are that, as a best practice, all IRQs should always registered (`vIRQ_SetVector(irqn, isr)`) before being enabled/disabled/set as pending/etc. Common patterns as clearing the pending bit before enabling an interrupt are still valid (and encouraged), but they require the vector to be set first (i.e., registered).

@meriac

### DIFF
--- a/core/system/inc/unvic.h
+++ b/core/system/inc/unvic.h
@@ -20,6 +20,9 @@
 
 #include "svc.h"
 
+#define UNVIC_IS_IRQ_ENABLED(irqn) (NVIC->ISER[(((uint32_t) ((int32_t) (irqn))) >> 5UL)] & \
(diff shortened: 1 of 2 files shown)


meriac:
This pull requests fixes bug https://github.com/ARMmbed/uvisor/issues/87

AlessandroA:
@patater

PR #114, opened 2015-12-01
changed uvisor compiler settings to speed-optimization

meriac:
(comes at ~9% size overhead currently)

@AlessandroA @patater


Patater:
Please make the first line of the commit message shorter. Move the 9% comment to the commit description. Please also describe the rationale for the change in the commit message, answering the question why are we changing to O2?

PR #117, opened 2015-12-09
Fix interrupt priority offset in vIRQ_GetPriority

AlessandroA:
The `vIRQ_GetPriority` API did not return the virtualised interrupt
priority, defined as:
`virtualised_priority = physical_priority - UVISOR_MIN_PRIORITY`.
The physical interrupt priority (from NVIC) was returned instead.

@meriac 
@Patater

### DIFF
--- a/core/system/src/unvic.c
+++ b/core/system/src/unvic.c
@@ -226,10 +226,7 @@ void unvic_irq_priority_set(uint32_t irqn, uint32_t priority)
     /* FIXME add check for maximum priority */
 
     /* set priority for device specific interrupts */
-    if(priority < UNVIC_MIN_PRIORITY)
-        NVIC_SetPriority(irqn, UNVIC_MIN_PRIORITY);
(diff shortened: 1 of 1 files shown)


Patater:
This implementation looks good to me, but I'm not sure if the change has any unintended side-effects.

AlessandroA:
@meriac Updated after your comment.

meriac:
@AlessandroA - i realised another bug :-) imagine "priority" is equal to (0xFFFFFFFFUL-UNVIC_MIN_PRIORITY) in unvic_irq_priority_set

AlessandroA:
@meriac yup, I think that was covered by the comment `/* FIXME add check for maximum priority */`, which is even more restrictive than `0xFFFFFFFFUL`. I'll fix it now. Shall we put it in a different PR?

meriac:
@AlessandroA - yes, makes sense

PR #118, opened 2015-12-11
Background ACLs for AIPS periph added to each box

AlessandroA:
When a new ACL (AIPS) was added to a box, a `memcpy` was always performed to copy
box 0 ACLs, overriding the previously-scanned ACLs for that box. This commit
fixes that bug by removing the `memcpy` and or-ing all boxes ACLs with the ones
from box 0, when box 0 is parsed.

@meriac 
@Patater


AlessandroA:
@meriac Updated after your comments: now memcpy removed and replaced by a selective or-ing: when box 0 ACLs are parsed all other boxes are initialized with the same ACLs.

PR #119, opened 2015-12-11
Added check for max priority in vIRQ APIs

AlessandroA:
Since we offset all interrupt priorities by `UNVIC_MIN_PRIORITY` we
want to make sure that the priority field does not overflow.

Added:
- sanity check on available number of interrupts priority levels
- default setting for priority grouping (this is set by default
  after reset but it is needed anyway as it might happen that
  `SystemInit()` changed it before `uvisor_init()`
- runtime check for `vIRQ_SetPriority()`

Note: the same runtime check is not needed for `vIRQ_SetVector()`
as the minimum priority is set by default and the `unvic_init()`
routine already checked that this is legit.

@meriac 
@Patater

### DIFF
--- a/core/system/src/unvic.c
+++ b/core/system/src/unvic.c
@@ -223,7 +223,11 @@ void unvic_irq_priority_set(uint32_t irqn, uint32_t priority)
     /* verify IRQ access privileges */
     is_irqn_registered = unvic_acl_check(irqn);
 
-    /* FIXME add check for maximum priority */
+    /* check for maximum priority */
+    if (priority >= ((1 << __NVIC_PRIO_BITS) - UNVIC_MIN_PRIORITY)) {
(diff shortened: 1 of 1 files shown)


core/system/src/unvic.c
```diff
@@ -223,7 +223,11 @@ void unvic_irq_priority_set(uint32_t irqn, uint32_t priority)
     /* verify IRQ access privileges */
     is_irqn_registered = unvic_acl_check(irqn);
 
-    /* FIXME add check for maximum priority */
+    /* check for maximum priority */
+    if ((UNVIC_MIN_PRIORITY + priority) > ((1 << __NVIC_PRIO_BITS) - 1)) {
```
meriac: if (priority >= ((1 << __NVIC_PRIO_BITS) - UNVIC_MIN_PRIORITY)) ?

AlessandroA:
@meriac Update after your suggestion

PR #121, opened 2015-12-15
vIRQ_SetVector turned into an actual function

AlessandroA:
This allows the function to exist as an actual symbol and to be used accordingly (for example, in a secure gateway).

@meriac 
@patater

### DIFF
--- a/core/mbed/source/interrupts.cpp
+++ b/core/mbed/source/interrupts.cpp
@@ -26,6 +26,11 @@ void vIRQ_SetVectorX(uint32_t irqn, uint32_t vector, uint32_t flag)
     }
 }
 
+void vIRQ_SetVector(uint32_t irqn, uint32_t vector)
+{
+    vIRQ_SetVectorX(irqn, vector, 0);
+}
+
 uint32_t vIRQ_GetVector(uint32_t irqn)
 {
     if(__uvisor_mode == 0) {
--- a/core/mbed/uvisor-lib/interrupts.h
+++ b/core/mbed/uvisor-lib/interrupts.h
@@ -18,6 +18,7 @@
 #define __UVISOR_LIB_INTERRUPTS_H__
 
 UVISOR_EXTERN void vIRQ_SetVectorX(uint32_t irqn, uint32_t vector, uint32_t flag);
+UVISOR_EXTERN void vIRQ_SetVector(uint32_t irqn, uint32_t vector);
 UVISOR_EXTERN uint32_t vIRQ_GetVector(uint32_t irqn);
 UVISOR_EXTERN void vIRQ_EnableIRQ(uint32_t irqn);
 UVISOR_EXTERN void vIRQ_DisableIRQ(uint32_t irqn);
@@ -27,8 +28,5 @@ UVISOR_EXTERN uint32_t vIRQ_GetPendingIRQ(uint32_t irqn);
 UVISOR_EXTERN void vIRQ_SetPriority(uint32_t irqn, uint32_t priority);
 UVISOR_EXTERN uint32_t vIRQ_GetPriority(uint32_t irqn);
 UVISOR_EXTERN int vIRQ_GetLevel(void);
-
-/* this definition is kept for backward compatibility with NVIC_SetVector */
-#define vIRQ_SetVector(irqn, vector) vIRQ_SetVectorX(irqn, vector, 0)
 
 #endif /* __UVISOR_LIB_INTERRUPTS_H__ */


core/mbed/source/interrupts.cpp
```diff
@@ -26,6 +26,11 @@ void vIRQ_SetVectorX(uint32_t irqn, uint32_t vector, uint32_t flag)
     }
 }
 
+void vIRQ_SetVector(uint32_t irqn, uint32_t vector)
```
Patater: Consider putting this in the header file and using the "inline" keyword.

AlessandroA: I still need it as an addressable symbol, though. If I define it as static inline in the header file it should not work, right?

Patater: Ah, yes. That's the whole point of this change.

Is there a reason we can't call gate vIRQ_SetVectorX? or rename vIRQ_SetVectorX to vIRQ_SetVector?

AlessandroA: Yes that is an option! `vIRQ_SetVector` keeps backward-compatibility with the corresponding NVIC API, meaning that it uses the same format `(irqn, vector)`. `vIRQ_SetVectorX` is somehow an extension, since it includes an additional flag -- although currently unused.

@meriac what do you think? Are we willing to keep the flag of `vIRQ_SetVectorX` or do we prefer sticking with the classic API?

For reference:
```C
void vIRQ_SetVector(uint32_t irqn, uint32_t vector);
void vIRQ_SetVectorX(uint32_t irqn, uint32_t vector, uint32_t flag);
```

Patater: I didn't notice they have different signatures. It would be a dumb idea to break backwards compatibility just to avoid a function call in non-critical path code.

What is the reason why a call gateway for vIRQ_SetVector is needed? Could the gateway be just as effective at the vIRQ_SetVectorX call?

On 15 Dec 2015, at 21:17, Alessandro Angelino <notifications@github.com<mailto:notifications@github.com>> wrote:


In core/mbed/source/interrupts.cpp<https://github.com/ARMmbed/uvisor/pull/121#discussion_r47699617>:

> @@ -26,6 +26,11 @@ void vIRQ_SetVectorX(uint32_t irqn, uint32_t vector, uint32_t flag)
>      }
>  }
>
> +void vIRQ_SetVector(uint32_t irqn, uint32_t vector)


Yes that is an option! vIRQ_SetVector keeps backward-compatibility with the corresponding NVIC API, meaning that it uses the same format (irqn, vector). vIRQ_SetVectorX is somehow an extension, since it includes an additional flag -- although currently unused.

@meriac<https://github.com/meriac> what do you think? Are we willing to keep the flag of vIRQ_SetVectorX or do we prefer sticking with the classic API?

For reference:

void vIRQ_SetVector(uint32_t irqn, uint32_t vector);
void vIRQ_SetVectorX(uint32_t irqn, uint32_t vector, uint32_t flag);

-
Reply to this email directly or view it on GitHub<https://github.com/ARMmbed/uvisor/pull/121/files#r47699617>.

IMPORTANT NOTICE: The contents of this email and any attachments are confidential and may also be privileged. If you are not the intended recipient, please notify the sender immediately and do not disclose the contents to any other person, use it for any purpose, or store or copy the information in any medium. Thank you.

AlessandroA: Yes, since it was for testing purposes having a call gateway to `vIRQ_SetVectorX` is fine as well. The point of the PR was that at the moment `vIRQ_SetVector` is much more widely used but it is not an explicit symbol, which might cause problem in some cases (secure gateway was just an example; debugging is probably another).

In general I'm happier to avoid pre-processors redefinitions, especially for a widely used API.

As for the specific secure gateway example with `vIRQ_SetVector`, it would make sure that a box has exclusive access to that IRQ, for example when running an initialization function:

```C
void box1_init(void)
{
    secure_gateway(box1, vIRQ_SetVector, TRNG_IRQn, (uint32_t) trng_handler);

    /* no need for the rest of the function to run in the context of box 1 */
    ...
}
```
Without the symbol one has to write a wrapper function that calls `vIRQ_SetVector`. Not a big deal, it's just inconsistent with the other APIs, which are available for direct gatewaying.

Patater: Please add this information to the commit message.

I guess this is a mainly a usability feature, as the knowledge to make the gateway with vIRQ_SetVectorX instead of vIRQ_SetVector is not widely known. In which case, after the "why" goes into your commit message, the commit seems good to me.

AlessandroA: @patater I updated the commit message with an additional comment on the reason behind the change

AlessandroA:
Bump @meriac

PR #123, opened 2015-12-24
allow acl declaration to be optional for UVISOR_BOX_CONFIG

meriac:
As a result of https://github.com/ARMmbed/uvisor-lib/issues/22 I've added the option for UVISOR_BOX_CONFIG to avoid declaration of box ACLs:
- UVISOR_BOX_CONFIG (now aliased to UVISOR_BOX_CONFIG_CTX) still behaves the same for backward compatibility in the 4 parameter case. For three parameters the ACL is dropped.
- UVISOR_BOX_CONFIG_ACL behaves the same for the 4 parameter case as UVISOR_BOX_CONFIG_CTX. For three parameters the context is dropped.

For both cases above for parameters both the context and the acl declaration are dropped.
@AlessandroA @Patater


Patater:
LGTM

PR #126, opened 2016-01-14
Export symbol of max priority level allowed (users)

AlessandroA:
With the classic NVIC module, a user would typically use `__NVIC_PRIO_BITS` to
determine the max priority level available. This commit makes sure that a
similar macro is exported for vIRQ users. Currently the symbol still relies on
`__NVIC_PRIO_BITS`, but this could change in the future if more complex
virtualisation is implemented.

Note that the minimum NVIC priority must be exported as well, but the name
has been prefixed with a double underscore to avoid that users use it
explicitly (it would make no sense as the NVIC module cannot be accessed
directly).

### DIFF
--- a/core/mbed/uvisor-lib/uvisor-lib.h
+++ b/core/mbed/uvisor-lib/uvisor-lib.h
@@ -28,6 +28,7 @@
 #include "uvisor-lib/halt_exports.h"
 #include "uvisor-lib/svc_exports.h"
 #include "uvisor-lib/svc_gw_exports.h"
+#include "uvisor-lib/unvic_exports.h"
 
 /* conditionally included header files */
 #if YOTTA_CFG_UVISOR_PRESENT == 1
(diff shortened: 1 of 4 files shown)


AlessandroA:
@meriac 
@Patater

Patater:
"to avoid that users use it" --> "to discourage users from using it"

Was `UNVIC_MIN_PRIORITY` purely internal, or might this break outside code?

AlessandroA:
(updated, thanks!)

It was internal in the sense that there is no reason to export it. Users cannot really make a use of it as NVIC priority levels are not the ones they directly set/get with `vIRQ_{S,G}etPriority(irqn)`.

It is now being exported otherwise the useful macro cannot be computed (`UVISOR_VIRQ_MAX_PRIORITY`)

Patater:
If we are sure we won't break outside code, I'm good with this change.

Issue #125, opened 2016-01-14
[K64F] A secure box with an empty ACL list does not get the background ACLs


AlessandroA:
In `vmpu_load_boxes()` we have a `for` loop that scans ACLs and calls the function `vmpu_acl_add()` for each of them.

In K64F this function further calls `vmpu_aips_add()`, which
* adds the relevant AIPS slot and
* **also adds ACLs from box 0 to the ACL list of the currently parsed box**

This last point creates the bug: If the ACL list for a secure box is empty, then the `vmpu_acl_add()` will never be called and hence (for the K64F) the box 0 ACLs for AIPS slots are never applied to other boxes.

For reference:
* [Call to `vmpu_acl_add()`](https://github.com/ARMmbed/uvisor/blob/master/core/system/src/mpu/vmpu.c#L169)
* [Box 0 ACLs are added to other boxes in `vmpu_aips_add()`](https://github.com/ARMmbed/uvisor/blob/master/core/system/src/mpu/vmpu_freescale_k64_aips.c#L104)

---

Possible **solution**:
* Create a new function that gets called after ACLs parsing: `vmpu_add_common_acls()`
    * Implemented differently for ARM and Freescale MPUs
    * All shared box 0 ACLs can be accounted for after private ACLs have been parsed
    *  *sharing box 0 ACLs* is decoupled from *having own ACLs*

@meriac 
@Patater

ciarmcom:
ARM Internal Ref: IOTSFW-1819

avri-schneider:
See my comment here: https://github.com/ARMmbed/uvisor/pull/127#discussion-diff-49997361

AlessandroA:
Fixed in https://github.com/ARMmbed/uvisor/pull/128

PR #127, opened 2016-01-15
Fix background ACL bug for boxes with no ACLs

AlessandroA:
This solves https://github.com/ARMmbed/uvisor/issues/125
@meriac @Patater 

**Problem**
Background ACLs are added automatically to each box's ACL list. On the K64F
this happens when an AIPS-slot ACL is being added. This means that if no
AIPS-slot ACL exists for a box, background ACLs will never be included for
that box and accesses captured by uVisor will be denied.

**Solution**
Defer the inclusion of box 0 ACLs after the box N ACLs parsing. This happens
independently from the fact that box N has ACLs. Currently this applies only
to K64F, AIPS-slots.

**For reference**
https://github.com/ARMmbed/uvisor/issues/125


meriac:
@AlessandroA I believe this breaks future intended behaviour: Once we have dynamic ACL's, box zero ACLs are not applied any more to all the other boxes as in the existing code.

AlessandroA:
@meriac Yes, in the future the feature would be removed completely. The problem as for now is that the feature is there but has unintended behaviour when the ACL list is empty.

I would suggest we at least fix the bug for now. The proposed solution also decouples the shared ACLs from the regular ACLs so it's easier to remove it afterwards.

AlessandroA:
@meriac 

Closing as the bug is not in the modified code, but rather in box enumeration. 

When [this line](https://github.com/ARMmbed/uvisor/blob/master/core/system/src/mpu/vmpu_freescale_k64_aips.c#L104) is executed, boxes have not been fully enumerated yet (`g_vmpu_box_count` is incremented step by step [here](https://github.com/ARMmbed/uvisor/blob/master/core/system/src/mpu/vmpu.c#L151)).

A new PR coming soon!

core/system/src/mpu/vmpu_freescale_k64_aips.c
```diff
@@ -117,6 +109,16 @@ int vmpu_aips_add(uint8_t box_id, void* start, uint32_t size, UvisorBoxAcl acl)
     return 1;
 }
 
+void vmpu_aips_common_add(void)
+{
+    int box, slot;
+    for (box = 1; box <= g_vmpu_box_count; box++) {
```
avri-schneider: `g_vmpu_box_count` can be `UVISOR_MAX_BOXES` - See: `if((box_id = g_vmpu_box_count++)>=UVISOR_MAX_BOXES)` in `/core/system/src/mpu/vmpu.c`

Should change condition to `box < g_vmpu_box_count`, so not to overflow `g_aipsx_box` buffer.

Alternatively change `if((box_id = g_vmpu_box_count++)>=UVISOR_MAX_BOXES)` to `if((box_id = ++g_vmpu_box_count)>=UVISOR_MAX_BOXES)`

AlessandroA: Thank you @avri-schneider ! I'm now sending a new PR that also fixes the issue you reported.

avri-schneider: Thanks @AlessandroA , the fix looks great! Want to be sure regarding this comment:
> `/* contains the total number of boxes`
> `* boxes are enumerated from 0 to (g_vmpu_box_count - 1) and the following`
> `* condition must hold: g_vmpu_box_count < UVISOR_MAX_BOXES */`

If `g_vmpu_box_count ==  UVISOR_MAX_BOXES - 1` (`g_vmpu_box_count < UVISOR_MAX_BOXES` condition), boxes are enumerated from `0` to `UVISOR_MAX_BOXES - 2`. There are `UVISOR_MAX_BOXES` allocated. Is this expected behavior?

PR #128, opened 2016-01-18
Fix box enumeration

AlessandroA:
With this PR boxes are enumerated more consistently.
Boxes IDs go from `0` to `(g_vmpu_box_count - 1)` and the following condition must hold:
```g_vmpu_box_count < UVISOR_MAX_BOXES```

The PR also fixes the bug for which ACLs were parsed before boxes were fully enumerated;
in that case box 0 ACLs were not applied to all boxes, but only to the boxes enumerated
so far.

@meriac 
@Patater

### DIFF
--- a/core/system/inc/mpu/vmpu.h
+++ b/core/system/inc/mpu/vmpu.h
@@ -125,6 +125,9 @@ extern void vmpu_init_post(void);
 
 extern void vmpu_sys_mux_handler(uint32_t lr, uint32_t msp);
 
+/* contains the total number of boxes
+ * boxes are enumerated from 0 to (g_vmpu_box_count - 1) and the following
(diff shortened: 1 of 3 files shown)


AlessandroA:
@avri-schneider

meriac:
@AlessandroA - very nice piece of code

avri-schneider:
@AlessandroA wrote:
> Boxes IDs go from 0 to (g_vmpu_box_count - 1) and the following condition must hold:
> g_vmpu_box_count < UVISOR_MAX_BOXES

Are you sure the expected behavior is to only iterate from `0` to `UVISOR_MAX_BOXES - 2` ?
If there are `UVISOR_MAX_BOXES` allocated, it is safe to iterate from `0` to `UVISOR_MAX_BOXES - 1`.

PR #129, opened 2016-01-20
Add explicit vIRQ_SetVector API to override.h (mbed)

AlessandroA:
In this previous commit:
https://github.com/ARMmbed/uvisor/commit/f83a82d881207f28ce01bd34e67d8df9dd0b9822
`vIRQ_SetVector` has been turned in an actual API instead of a preprocessor macro
redefinition.

This commit mirrors that change to the `override.h` file, which is used in mbed
to override NVIC symbols when one does not want to replace them manually in the
source code.

An example use case:
https://github.com/ARMmbed/mbed-hal-st-stm32f4/blob/master/source/hal_tick.c#L36.

@Patater 
@meriac

### DIFF
--- a/core/mbed/uvisor-lib/override.h
+++ b/core/mbed/uvisor-lib/override.h
@@ -27,6 +27,7 @@
 #include <stdint.h>
 
 extern void vIRQ_SetVectorX(uint32_t irqn, uint32_t vector, uint32_t flag);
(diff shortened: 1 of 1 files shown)


Patater:
In the commit message, add the name of the commit after the commit hash like so.
f83a82d ("vIRQ_SetVector turned into an actual function")

LGTM

AlessandroA:
@Patater Thanks, done!

PR #136, opened 2016-01-29
The huge refactoring

AlessandroA:
This is a uVisor refactoring.

Please look at the single commits to see what changed. The following criteria have been used:
- **Renaming** happened in dedicated commits. You can still follow the history for a given file before the renaming took place with `git log --follow filename`.
- **Deleting** happened in dedicated commits.
- **Refactoring** followed this scheme:
    - **NO** hw-specific code in `core/`
    - A dedicated folder `platform/` with all the possible hw-specific code/headers
    - Platform-specific header files kept to a minimum
    - The port is ready for extension to multi-targets (that is, building a uvisor-lib binary that supports multiple chips at the same time)
        - A separate PR will be created with more details)

**Rationale**
uVisor only relies on a few symbols that are hardware-specific. As it turns out, there is absolutely no need to use hardware-specific header files, as those definitions can be wrapped up in a custom header file (see file [`symbols.h`](https://github.com/ARMmbed/uvisor/blob/the_huge_refactoring/platform/kinetis/inc/symbols.h)). This file can be even autogenerated from the platform SDK.

Note: In some cases it still necessary to keep the header files. For example, for external MPUs (NXP).

The refactoring introduced here has the following advantages:
- To port a platform to uVisor, only files in one single folder need changing (`platform/`)
- The effort of porting 1 platform can be easily extended to port a whole family of devices
- The build system is made to compile all targets/configurations at the same time, for a given family

@meriac 
@Patater


AlessandroA:
**DO NOT MERGE**

The refactoring is being rebased to include latest offline discussions.

@Patater 
@meriac

AlessandroA:
New PR coming.

PR #138, opened 2016-02-02
Remove need of SRAM_LENGTH and FLASH_LENGTH

AlessandroA:
The 2 symbols `SRAM_LENGTH` and `FLASH_LENGTH` are now fetched from Flash at boot time, and are provided by the host linker script.

Sanity checks have been updated accordingly. A new sanity check verifies that the values read from flash are the same as expected from the hard-coded uVisor symbols, when available.

**Note**: The symbols `FLASH_LENGTH` and `SRAM_LENGTH` are not removed. They are used only once to pre-process the uVisor linker script and to determine if the Flash/SRAM usage fits into the target device:
```C
#define FLASH_MAX (FLASH_LENGTH - RESERVED_FLASH)
#define SRAM_MAX  (SRAM_LENGTH - RESERVED_SRAM)
```
When a single uVisor port will support multiple chips at a time (ideally, a whole family), they will represent the **minimum** Flash/SRAM length needed to have uVisor fit into the target memories for that family. 

Requires a PR with [these changes](https://github.com/ARMmbed/target-kinetis-k64-gcc/compare/master...AlessandroA:master?expand=1) (mbed linker script).

@meriac 
@Patater

### DIFF
--- a/core/system/inc/mpu/vmpu.h
+++ b/core/system/inc/mpu/vmpu.h
@@ -20,9 +20,23 @@
 #include "vmpu_exports.h"
 #include "vmpu_unpriv_access.h"
 
-#define VMPU_FLASH_ADDR_MASK  (~(((uint32_t)(FLASH_LENGTH)) - 1))
-#define VMPU_FLASH_ADDR(addr) ((VMPU_FLASH_ADDR_MASK & (uint32_t)(addr)) == \
-                               (FLASH_ORIGIN & VMPU_FLASH_ADDR_MASK))
+/* Check if the address is in Flash/SRAM. */
+/* Note: Instead of using the '<' check on
+ *          __uvisor_config.{flash, sram}_end
+ *       we use the '<=' check on
+ *          __uvisor_config.{flash, sram}_end - 4
+ *       because unaligned accesses at a physical memory boundary have undefined
+ *       behavior and must be avoided. */
+static inline int vmpu_flash_addr(uint32_t addr)
+{
+    return (((uint32_t) addr >= FLASH_ORIGIN) &&
+            ((uint32_t) addr <= (FLASH_ORIGIN + (uint32_t) __uvisor_config.flash_end - 4)));
+}
+static inline int vmpu_sram_addr(uint32_t addr)
+{
+    return (((uint32_t) addr >= SRAM_ORIGIN) &&
+            ((uint32_t) addr <= (SRAM_ORIGIN + (uint32_t) __uvisor_config.sram_end - 4)));
+}
 
 #define VMPU_REGION_SIZE(p1, p2) ((p1 >= p2) ? 0 : \
                                              ((uint32_t) (p2) - (uint32_t) (p1)))
--- a/core/system/inc/mpu/vmpu_freescale_k64_map.h
+++ b/core/system/inc/mpu/vmpu_freescale_k64_map.h
@@ -18,7 +18,6 @@
 #define __VMPU_FREESCALE_K64_MAP_H__
 
 #define MEMORY_MAP_SRAM_START       ((uint32_t) SRAM_ORIGIN)
-#define MEMORY_MAP_SRAM_END         ((uint32_t) (SRAM_ORIGIN + SRAM_LENGTH))
 #define MEMORY_MAP_PERIPH_START     ((uint32_t) 0x40000000)
 #define MEMORY_MAP_PERIPH_END       ((uint32_t) 0x400FEFFF)
 #define MEMORY_MAP_GPIO_START       ((uint32_t) 0x400FF000)
--- a/core/system/src/mpu/vmpu.c
+++ b/core/system/src/mpu/vmpu.c
@@ -59,16 +59,20 @@ static int vmpu_sanity_checks(void)
     assert(vmpu_bits(0x8001UL) == 16);
     assert(vmpu_bits(1) == 1);
 
-    /* verify that __uvisor_config is within valid flash */
-    assert( ((uint32_t) &__uvisor_config) >= FLASH_ORIGIN );
-    assert( ((((uint32_t) &__uvisor_config) + sizeof(__uvisor_config))
-             <= (FLASH_ORIGIN + FLASH_LENGTH)) );
+    /* Verify that the known hard-coded symbols are equal to the ones taken from
+     * the host linker script. */
+    assert((uint32_t) __uvisor_config.flash_start == FLASH_ORIGIN);
+    assert((uint32_t) __uvisor_config.sram_start == SRAM_ORIGIN);
+
+    /* Verify that the uVisor binary blob is positioned at the flash offset. */
+    assert(((uint32_t) __uvisor_config.flash_start + RESERVED_FLASH) ==
+           (uint32_t) __uvisor_config.main_start);
+
 
     /* verify if configuration mode is inside flash memory */
-    assert((uint32_t)__uvisor_config.mode >= FLASH_ORIGIN);
-    assert((uint32_t)__uvisor_config.mode <= (FLASH_ORIGIN + FLASH_LENGTH - 4));
-    DPRINTF("uvisor_mode: %u\n", *__uvisor_config.mode);
-    assert(*__uvisor_config.mode <= 2);
+    assert(vmpu_flash_addr((uint32_t) __uvisor_config.mode));
+    assert(*(__uvisor_config.mode) <= 2);
+    DPRINTF("uVisor mode: %u\n", *(__uvisor_config.mode));
 
     /* verify SRAM relocation */
     DPRINTF("uvisor_ram : @0x%08X (%u bytes) [config]\n",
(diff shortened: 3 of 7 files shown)


core/system/inc/mpu/vmpu.h
```diff
@@ -20,9 +20,17 @@
 #include "vmpu_exports.h"
 #include "vmpu_unpriv_access.h"
 
-#define VMPU_FLASH_ADDR_MASK  (~(((uint32_t)(FLASH_LENGTH)) - 1))
-#define VMPU_FLASH_ADDR(addr) ((VMPU_FLASH_ADDR_MASK & (uint32_t)(addr)) == \
-                               (FLASH_ORIGIN & VMPU_FLASH_ADDR_MASK))
+/* Check if address in is Flash/SRAM. */
```
Patater: `the address`

core/system/inc/mpu/vmpu_freescale_k64_map.h
```diff
@@ -18,7 +18,7 @@
 #define __VMPU_FREESCALE_K64_MAP_H__
 
 #define MEMORY_MAP_SRAM_START       ((uint32_t) SRAM_ORIGIN)
-#define MEMORY_MAP_SRAM_END         ((uint32_t) (SRAM_ORIGIN + SRAM_LENGTH))
+#define MEMORY_MAP_SRAM_END         ((uint32_t) (__uvisor_config.sram_end))
```
Patater: Consider removing the extra parenthesis here.

core/system/src/mpu/vmpu.c
```diff
@@ -59,16 +59,24 @@ static int vmpu_sanity_checks(void)
     assert(vmpu_bits(0x8001UL) == 16);
     assert(vmpu_bits(1) == 1);
 
+    /* Verify that the known hard-coded symbols are equal to the ones taken from
+     * the host linker script. */
+    assert((uint32_t) __uvisor_config.flash_start == FLASH_ORIGIN);
+    assert((uint32_t) __uvisor_config.flash_offset == FLASH_ORIGIN + RESERVED_FLASH);
+    assert((uint32_t) __uvisor_config.sram_start == SRAM_ORIGIN);
+    assert((uint32_t) __uvisor_config.sram_offset == SRAM_ORIGIN + RESERVED_SRAM);
```
Patater: This seems like an odd thing to do. It's good for now, for testing that the new technique is equivalent, but these asserts will be invalid when we begin to run code for which the likes of `__uvisor_config.flash_start` were added. If this weren't the case, then we wouldn't need to add the new config variables. Consider removing the asserts since they will break soon.

core/system/src/mpu/vmpu.c
```diff
@@ -59,16 +59,24 @@ static int vmpu_sanity_checks(void)
...
+
+    /* Verify that the uVisor binary blob is positioned at the flash offset. */
+    assert(__uvisor_config.flash_offset == __uvisor_config.main_start);
+
     /* verify that __uvisor_config is within valid flash */
-    assert( ((uint32_t) &__uvisor_config) >= FLASH_ORIGIN );
-    assert( ((((uint32_t) &__uvisor_config) + sizeof(__uvisor_config))
-             <= (FLASH_ORIGIN + FLASH_LENGTH)) );
+    assert(VMPU_FLASH_ADDR((uint32_t) &__uvisor_config));
+    assert(VMPU_FLASH_ADDR(((uint32_t) &__uvisor_config) + sizeof(__uvisor_config)));
```
Patater: I would prefer leaving the comparison operators here rather than moving them inside the `VMPU_FLASH_ADDR` macro, or renaming the macro to `VMPU_FLASH_ADDR_VALID`.

core/system/src/mpu/vmpu.c
```diff
@@ -59,16 +59,24 @@ static int vmpu_sanity_checks(void)
...
 
     /* verify if configuration mode is inside flash memory */
-    assert((uint32_t)__uvisor_config.mode >= FLASH_ORIGIN);
-    assert((uint32_t)__uvisor_config.mode <= (FLASH_ORIGIN + FLASH_LENGTH - 4));
-    DPRINTF("uvisor_mode: %u\n", *__uvisor_config.mode);
-    assert(*__uvisor_config.mode <= 2);
+    assert(VMPU_FLASH_ADDR((uint32_t) __uvisor_config.mode));
+    assert(*(__uvisor_config.mode) <= 2);
+    DPRINTF("uVisor mode: %u\n", *(__uvisor_config.mode));
```
Patater: Doesn't this mode get printed somewhere else already? Do we want to print it twice?

core/system/src/mpu/vmpu.c
Patater: Actually, since those hard-coded defines are now the `minimum` space required, change the asserts to test that there is enough room instead of testing that the room available is exactly equal.

AlessandroA: I think I didn't explain those symbols correctly. `__uvisor_config.flash_start` is supposed to be exactly `FLASH_ORIGIN` and so on, because they represent exactly the same thing.

Here I'm basically performing sanity checks between what I hardcoded in uVisor (`FLASH_ORIGIN`) and what I instead receive from the host linker script (`__uvisor_config.flash_origin`) in the configuration table. This table is populated by mbed so we need to check that it was done properly. 

If the misunderstanding was caused by `RESERVED_{FLASH, SRAM}`, they actually mean "space to reserve before uVisor is placed", so in a sense they should be called `{FLASH, SRAM}_OFFSET` just like their counterparts in the configuration table. (Spoiler alert: they will be renamed, in a subsequent PR)

core/system/src/mpu/vmpu.c
AlessandroA: It only gets printed here that I'm aware of

core/system/inc/mpu/vmpu.h
AlessandroA: Fixed

core/system/inc/mpu/vmpu_freescale_k64_map.h
AlessandroA: Done

core/system/src/mpu/vmpu.c
Patater: Yup, this was where it was getting originally printed. I didn't notice the removal of the old print in the diff for some reason.

core/system/src/mpu/vmpu.c
AlessandroA: Update: I changed the `{flash, sram}_offsets` to represent the space reserved of Flash/SRAM before uVisor. They now map exactly to `RESERVED_{FLASH, SRAM}'.

Patater: Based on your forward looking comment that these _will_ represent the minimum _at some future time_, I was criticizing the asserts as currently written. I'm completely fine with the asserts as is for now, and, as I later realized, changing the asserts to ensure sure there is enough room is a natural thing to do at the _future_ time where we need that behavior.

core/system/src/mpu/vmpu.c
AlessandroA: @meriac @patater what about `VMPU_IS_ADDR_IN_FLASH(addr)` (and correspondingly `VMPU_IS_ADDR_IN_SRAM(addr)`)

Patater:
```
Note: The symbols FLASH_LENGTH and SRAM_LENGTH are not removed. They now
represent the **minimum** Flash/SRAM length needed to have uVisor fit into the
target memories.
```

Revise the note in the commit message to not say "now", because they don't yet represent the **minimum**. Because, in fact, we still have asserts in the code that make sure FLASH_LENGTH and SRAM_LENGTH are not the **minimum**, but are actually exactly the amount both required and available.

core/system/src/mpu/vmpu.c
Patater: That sounds good to me.

AlessandroA:
@Patater Wait, are we mixing `FLASH_ORIGIN` and `FLASH_LENGTH` in the previous comment? Or maybe I'm not following you.

The assert in the code only check for `FLASH_ORIGIN` and `SRAM_ORIGIN`, not for `*_LENGTH`. The `*_LENGTH` ones will be calculated as the minimum of all the possible available lengths for a certain chip family.

As of now, they still are but in a corner case: `min(all lenghts) = min(length of the only chip supported)`. The `*_LENGTH` parameter is only used by the linker to assess if there is enough space: `max space that uVisor can take < min(all lenghts)`.

Patater:
Yes, I was confused. I had misunderstood `RESERVED_FLASH` as being `RESERVED_FLASH_LENGTH`. and not `RESERVED_FLASH_START` and `flash_offset` as being `flash_end`. So, never mind, and remind me not to code review while sick.

Do whatever makes sense for the asserts (your old way or the new way), ignoring my confusion.

AlessandroA:
No worries. Actually your comments on this triggered a fix so they were very useful -- thanks! But yes, don't work and get well soon! :)

AlessandroA:
Bump @meriac @Patater

Patater:
The build fails on stm32f4. Could you look into it?

```
In file included from ../../core/system/src/mpu/vmpu.c:17:0:
../../core/system/src/mpu/vmpu.c: In function 'vmpu_sanity_checks':
../../core/system/src/mpu/vmpu.c:67:54: error: 'RESERVED_SRAM' undeclared (first use in this function)
```

AlessandroA:
Yes, sorry. I did not realize [this change](https://github.com/ARMmbed/uvisor/pull/138/files#diff-348e4d41162717e340febb0ed05aaa3fR22) was not staged. It now compiles for both targets.

Patater:
LGTM

AlessandroA:
@meriac Removed `__uvisor_config.flash_offset` and `__uvisor_config.sram_offset` as from our offline discussion -- they are duplicate of other symbols (`__uvisor_config.main_start`, where the uVisor code is located, and `__uvisor_config.bss_main_start`, where the uVisor first SRAM section is located).

AlessandroA:
@meriac Are we ready to merge?

PR #139, opened 2016-02-02
Replace hardware-specific system.{c, h} files with global ones

AlessandroA:
This PR removes the need of having hardware-specific `system.{c, h}` files. It gets us one step closer to make uVisor more hardware-agnostic.

@Patater 
@meriac


AlessandroA:
Bump @meriac @Patater

Patater:
The build fails on stm32f4. Could you look into it?

```
../../core/system/src/mpu/vmpu_armv7m.c: In function 'vmpu_sys_mux_handler':
../../core/system/src/mpu/vmpu_armv7m.c:176:49: error: 'IRQn_OFFSET' undeclared (first use in this function)
     int ipsr = ((int) (__get_IPSR() & 0x1FF)) - IRQn_OFFSET;
                                                 ^
```

core/system/inc/system.h
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2013-2015, ARM Limited, All Rights Reserved
```
Patater: Should this be 2016 only? If not, at least include 2016.

core/system/inc/system.h
```diff
@@ -0,0 +1,41 @@
...
+/* System IRQs */
+extern void NonMaskableInt_IRQn_Handler(void);
+extern void HardFault_IRQn_Handler(void);
+extern void MemoryManagement_IRQn_Handler(void);
+extern void BusFault_IRQn_Handler(void);
+extern void UsageFault_IRQn_Handler(void);
+extern void SVCall_IRQn_Handler(void);
+extern void DebugMonitor_IRQn_Handler(void);
+extern void PendSV_IRQn_Handler(void);
+extern void SysTick_IRQn_Handler(void);
```
Patater: Do we need to declare these System IRQ handlers in this header file? (I know we previously did, but I'm not sure we need to. At least, uVisor builds fine if we don't declare these here.)

core/system/src/system.c
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright (c) 2013-2015, ARM Limited, All Rights Reserved
```
Patater: Should this be 2016 only? If not, at least include 2016.

core/system/src/system.c
```diff
@@ -0,0 +1,84 @@
...
+        "push {lr}\n"
+        "blx vmpu_sys_mux_handler\n"
+        "pop {pc}\n"
+    );
+}
+
+void UVISOR_NAKED UVISOR_NORETURN isr_default_handler(void)
+{
+    /* Handle NVIC IRQ with an unprivileged handler.
+     * Serving an IRQn in unprivileged mode is achieved by mean of two SVCalls;
```
Patater: Replace `;` with `:`.

core/system/src/system.c
```diff
@@ -0,0 +1,84 @@
...
+     * the first one de-previliges execution, the second one re-privileges it. */
+    /* Note: NONBASETHRDENA (in SCB) must be set to 1 for this to work. */
```
Patater: Should we add an assertion someplace to check that NONBASETHRDENA is set to 1? (It's probably not good for performance reasons [even for a debug build] to put the assertion here in the interrupt handler, so hopefully there is a better place.)

core/uvisor.h
```diff
@@ -68,7 +68,7 @@ typedef void (*UnprivilegedBoxEntry)(void);
 /* IOT-OS specific includes */
 #include <linker.h>
 
-/* system specific declatiions and IRQ vector table */
-#include <system.h>
+/* System IRQ vector table and default ISRs */
+#include "system.h"
```
Patater: This seems like a purely cosmetic change. Probably it shouldn't be part of this commit.

Patater:
I'd recommend squashing these two commits together, since there isn't really a practical benefit in doing the deletion separate from the creation of the new system.c and system.h files; they are also both part of a single, coherent change. There is a practical downside for having these commits separate: the build will fail after the remove commit, and we'd prefer each commit to be buildable (so we can, for example, automate running tests on our history with git bisect).

core/system/inc/system.h
AlessandroA: Yup, it should be 2016 (but still including the start date. Most of the file is copy-pasted by a file previously copyrighted 2013-2015).

Patater: OK. I wasn't sure if it was a derivative work of the old file or a totally new thing.

core/system/inc/system.h
AlessandroA: I would keep them because the reason they are not needed is a "collateral". By [aliasing them](https://github.com/ARMmbed/uvisor/pull/139/files#diff-eadeeb38ba6620c41387e36e1f53d6adR20) we declare an `extern` symbol which is then referenced in the same file. At link time the declaration is solved and no error is thrown.

But if the alias is not there the same code in `system.c` fails without the `extern`s of `system.h`. As a general practice I'd always externally declare symbols that might be referenced somewhere else.

core/system/src/system.c
AlessandroA: Yup, it should be 2016 (but still including the start date. Most of the file is copy-pasted by a file previously copyrighted 2013-2015).

core/system/src/system.c
AlessandroA: :+1:

core/system/src/system.c
AlessandroA: It's set by default at startup time [here](https://github.com/ARMmbed/uvisor/blob/master/core/system/src/mpu/vmpu.c#L349) but as it turns out I never explicitly wrote `NONBASETHRDENA` in the comments, which makes it un`grep`able. I'll create a quick PR for that.

It does not need continuous runtime assertions, anyway. If it's not set the IRQ de-privileging fails with a verbose blue screen that reports an `EXC_RETURN` mismatch error (in the `CFSR`).

AlessandroA:
The commits were squashed! Thanks for the review.

core/uvisor.h
Patater: I should've read the comment more closely. The system.h header is no longer system-specific, so it makes sense to change this comment at the same time the system.h header is becoming no longer system-specific. I am OK with this change staying in.

Patater:
LGTM

PR #140, opened 2016-02-04
Improve comment on NONBASETHRDENA bit

AlessandroA:
The bit is set but it's name does not appear explicitly, which makes it
difficult to `grep`. It is important for this information to stand out as the
feature is vital for interrupts de-privileging.

@Patater @meriac


Patater:
LGTM

PR #142, opened 2016-02-09
uVisor refactoring

AlessandroA:
This is a uVisor refactoring. The net effects of the refactoring are:
* A new folder `platform` contains hardware-specific configurations, divided by device families (currently `kinetis`, `stm32`).
* No hardware-specific code is needed any more:
    * LED blinking patterns have been removed.
    * Memory maps have been removed.
* No hardware-specific headers are needed any more. A `core_generic.h` file defines all that is needed to replace the missing hw-specific CMSIS headers.
* Platform-specific Makefiles iterate over all the possible configurations and generate multiple uVisor binaries (as many as the configurations of Flash/SRAM origin, NVIC number, etc.).

Tested on FRDM-K64F and STM32F4 Discovery Board.

@meriac @Patater


AlessandroA:
Note: The new `PORTING.md` guide that results from this PR is already being modified but will be included in a separate PR after details of the refactoring have been accepted.

Patater:
Here are my review comments for the commit, "Remove all hardware-specific files and code".

This is great. I love all the deleted code.

Replace "we do not include devices memory maps" with "we do not include any device memory maps".

Is the change in HALT_ERROR for the debugging case from `halt_line(__FILE__, __LINE__, reason, __VA_ARGS__)` to `halt_line(__FILE__, __LINE__, reason, ##__VA_ARGS__)` necessitated by the deletion of hardware specific code? If not, then consider doing a separate commit for the change first. It seems like we ought to have been able to deal with empty args even before we deleted all the hardware specific code.

Hookable uVisor fault handlers would be nice, if a uVisor user wanted to restore the previous "blink an LED on fault" on their platform, or especially for unit testing. Is this feature request already being tracked someplace?

Patater:
Here is my review for the commit, "Add NXP Kinetis MPU header file".

I'm not sure what the changes in `core/linker` and `core/system/src/mpu/vmpu_armv8m.c` and `core/uvisor-config.h` have to do with adding this Kinetis MPU header file. Should those changes be part of this commit or part of the delete hardware specific stuff commit, or some other commit?

AlessandroA:
* 'halt_line' bug now fixed in https://github.com/ARMmbed/uvisor/pull/144 (so it can be applied before we finish reviewing this, which will take longer).
* Fixed comment in 'memory_map.c'.

As for the hookable uVisor fault handlers I agree. I'll create an issue/enhancement. This PR removes the memory map and the LED blinking patterns, as they are hw-specific. Since they are very useful for debugging we should track the missing feature.

AlessandroA:
You're right, all the files in that commit but the MPU header belonged to a different commit which had been squashed by mistake. It has now restored. As for the MPU header file, I'd still leave it in its own commit has it needs more visibility (it is derived from a partner's code).

Patater:
> As for the MPU header file, I'd still leave it in its own commit has it needs more visibility (it is derived from a partner's code).
That's fine with me. Place the commit right before the commit it that needs it ("Add core_generic.h to select correct CMSIS headers"), at least.

"Remove all hardware-specific files and code"
I see we added `#include "system.h"` to core/system/src/main.c. Is this needed due to removal of hardware-specific files? or should it be a part of a separate commit?

"Add core_generic.h to select correct CMSIS headers"
`Verify that the priority bits read at runtime are correct.` could more accurately say `Verify that the priority bits read at runtime are realistic.`, since we actually can't know what is truly correct.

"Fix conditional macros definition in exported file"
Can we do this before the refactor? I think it makes sense to fix this sooner and separately from the refactor.

"Add experimental support for NXP Kinetis and STM32 families"
Did the NXP MK64FN12MVLL12 and STM32F429ZI work before this commit? If not, consider splitting out the work to add those targets into a commit, and then making a new commit to add support for the families. This way, it becomes clear what is needed to support the old targets after the refactor, and what is needed to additionally support the families beyond the old targets. For instance, I'd expect the Makefile.rules changes necessary for the refactor to work for the previous targets to be in a separate commit from any changes (even if just renaming) needed to support new targets.

"Add experimental support for NXP Kinetis and STM32 families"
I see the file "platform/stm32/docs/memory_layout.svg" show up here. Did we delete that in the remove hardware specific commit? Is this file related to the old one? Also, this seems like a hardware specific file. Would we be better off having this someplace else? Also, platform/kinetis doesn't have any "docs" folder.

AlessandroA:
First 3 points in last comment addressed.

As for those regarding "Add experimental support for NXP Kinetis and STM32 families" her are my comments:

* I would only reword the commit to ("Re-add support for K64F and STM32F429 after refactoring"), but still keep the name of the families as subfolders in the `platform` dir (`kinetis`, `stm32`). This is because the whole point of the refactoring is to be able to port families, and not single devices, to uVisor. In the particular case of the commit we are talking about, we would be adding just one device per family. This will be extended in a next commit. The documentation will still highlight the limitations -- that is, we currently support only two devices.

* We could probably remove the docs folder altogether and plan on how to replace it with something more general. We need anyway an updated picture and that would be a good occasion to do that. Also, after the refactoring we will have at least 2 figures, one for each supported MPU. The figure for the ARMv7M MPU should also highlight the different structures in case the SRAM uVisor is placed into is different from the SRAM mbed uses (for example, the in the STM32F429 the uVisor uses the CCM, mbed uses the actual SRAM).

Patater:
Rewording the commit sounds good. If there were any extra work required to enable family-adding, then I'd expect that to go in a different commit. I'm fine with just rewording because there doesn't seem to be any extra family-adding work here aside from just re-adding support for the previously supported targets.

It may make things clearer to add support for each previously supported target in separate commits. Then, when somebody wants to port to a new family, they could look at one of these commits and see exactly what would be involved. With both families added at once in one commit, it's a bit less clear.

I agree we need to discuss further what to do with the svgs and docs in general. Maybe adding one svg (for the ARMv7M MPU) is all we need. I'd imagine the svg would be added with the commit that adds support for STM32F429.

AlessandroA:
* Changed last commit into 3 separate commits:
    ** The first one updated the Makefile to the new folder structure
    ** The second and third one add support for the old devices, respectively

* I left out the `docs` folder. We should open an issue for that and re-draw them updated

Patater:
Please wrap commit message descriptions to 72 characters. Try to keep the first line below 50 characters long.

"Update build system after refactoring"
$(SYSTEM_DIR)/src/mpu/vmpu.c didn't need to move, did it?

AlessandroA:
:+1:

AlessandroA:
> "Update build system after refactoring"
> $(SYSTEM_DIR)/src/mpu/vmpu.c didn't need to move, did it?

Well, I wanted it so be where all the `$(SYSTEM_DIR)` guys are :smile:

Patater:
> Well, I wanted it so be where all the $(SYSTEM_DIR) guys are :smile:
If there are any other simple organization and clean up things like this, we can do them right now in their own commit before this giant refactor.

AlessandroA:
That's the only one I spotted -- I wouldn't spend more time on it anyway as I prefer moving forward. It will unblock you for your port as well.

Patater:
Yeah, that's fine. It's the only one I spotted as well.

@meriac I'm done with my review of this. Could you give it a review?

Patater:
"Re-add support for K64F after refactoring"
This commit deletes "`STM32F429xx/*` from core/mbed/source/.gitignore. The "Re-add support for STM32F429 after refactoring" commit should do this instead, since you really just intend to rename `MK64FN1M0XXX12/*` to `kinetis/*` and `STM32F429xx/*` to `stm32/*` not to delete anything.

PR #144, opened 2016-02-10
Fix halt_line bug (empty __VA_ARGS__)

AlessandroA:
@Patater


Patater:
Could you add some description in the commit message of when one might run into this problem, and how this change fixes that?

Thanks

AlessandroA:
Done

Patater:
LGTM

PR #146, opened 2016-02-11
Fix conditional macros definition in exported file

AlessandroA:
`vmpu_export.h` is exported to mbed, which means that the definitions `ARCH_MPU_*`
do not apply. The wrong macros for rounding were selected.

Now the file checks for either the uvisor-facing definitions or for the uvisor-lib-facing ones.

@Patater @meriac


Patater:
LGTM

AlessandroA:
(Tested)

PR #147, opened 2016-02-12
Fix disabled.cpp bugs / missing features

AlessandroA:
Missing features added here:
* Box context switching for IRQs
* Check on call stack pointer over/underflow

Bugs fixed:
* Call stack pointer had off-by-one error

@meriac 
@Patater

### DIFF
--- a/core/mbed/source/disabled.cpp
+++ b/core/mbed/source/disabled.cpp
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 #include "uvisor-lib/uvisor-lib.h"
+#include "cmsis-core/cmsis_nvic.h"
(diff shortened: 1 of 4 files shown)


Patater:
Add "Fixes" tag for "Call stack pointer had off-by-one error"

AlessandroA:
@meriac Right! I put one `const` back. Now the declaration style should also be fixed. 

@Patater do you mean you'd like a "Fixes: $SHA ("$COMMIT_MSG")" in the commit that fixes the bug?

Patater:
> do you mean you'd like a "Fixes...
Yup.

AlessandroA:
Done

PR #148, opened 2016-02-15
Fix conditional macro def. (unsupported device)

AlessandroA:
The changes introduced in the previous commit (see below) introduced a
further bug for unsupported devices, for which an "Unknown MPU" error
message was thrown.

For unsupported devices the rounding macros are bypassed. No rounding is
applied.

Fixes: 143c256f5ba6 ("Fix conditional macros definition in exported file")

@Patater 
@meriac

### DIFF
--- a/core/system/inc/mpu/vmpu_exports.h
+++ b/core/system/inc/mpu/vmpu_exports.h
@@ -52,7 +52,7 @@
 #define UVISOR_TACL_SUBREGIONS_POS  24
 #define UVISOR_TACL_SUBREGIONS_MASK (0xFFUL<<UVISOR_TACL_SUBREGIONS_POS)
 #define UVISOR_TACL_SUBREGIONS(x)   ( (((uint32_t)(x))<<UVISOR_TACL_SUBREGIONS_POS) & UVISOR_TACL_SUBREGIONS_MASK )
-#endif /* defined(ARCH_MPU_ARMv7M) || (defined(YOTTA_CFG_UVISOR_PRESENT) && !defined(TARGET_LIKE_KINETIS)) */
+#endif
 
 #define UVISOR_TACLDEF_SECURE_BSS   (UVISOR_TACL_UREAD          |\
                                      UVISOR_TACL_UWRITE         |\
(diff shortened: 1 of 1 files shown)


core/system/inc/mpu/vmpu_exports.h
```diff
@@ -88,18 +88,23 @@
...
+#elif !defined(YOTTA_CFG_UVISOR_PRESENT) || YOTTA_CFG_UVISOR_PRESENT == 0
+#define UVISOR_REGION_ROUND_DOWN(x) (x)
+#define UVISOR_REGION_ROUND_UP(x)   (x)
+#define UVISOR_STACK_SIZE_ROUND(x)  (x)
 #else
 #error "Unknown MPU architecture. uvisor: Check your Makefile. uvisor-lib: Check if uVisor is supported"
 #endif /* defined(ARCH_MPU_ARMv7M) || (defined(YOTTA_CFG_UVISOR_PRESENT) && !defined(TARGET_LIKE_KINETIS)) or
-          defined(ARCH_MPU_KINETIS) || (defined(YOTTA_CFG_UVISOR_PRESENT) && defined(TARGET_LIKE_KINETIS)) */
+          defined(ARCH_MPU_KINETIS) || (defined(YOTTA_CFG_UVISOR_PRESENT) && defined(TARGET_LIKE_KINETIS)) or
+          !defined(YOTTA_CFG_UVISOR_PRESENT) || YOTTA_CFG_UVISOR_PRESENT == 0 */
```
Patater: This comment doesn't appear to match the `#if`. I'd recommend deleting the comment. I would rather we not add these sorts of comments and not require them in our style guide, since they are a pain to maintain and any decent text editor can indicate which `#endif` goes with which `#if`.

AlessandroA: Yeah, I agree. I'm removing it.

AlessandroA:
* Removed closing comments for `#endif`.
* Applied the same fix to other conditional macro definition in same file.

Patater:
Could you swap the order of these commits? It'd be better to do stylistic and refactoring things first before your new change.

AlessandroA:
Done

Patater:
LGTM

PR #150, opened 2016-02-15
Implement Box ID API

Patater:
XXX Don't merge, yet. This needs a bit more work and testing. I expect to make changes as a result of this PR.

Provide the ability to query:
- Numerical box ID of one's own box
- Numerical box ID of an inbound (across box boundary) caller
- String ID for box via numerical box ID

Provide the ability to set a box name. This change breaks binary compatibility.

@AlessandroA
@meriac


AlessandroA:
Added comments to specific commits.

Patater:
FYI, GitHub is showing the order of commits weird. The order is supposed to be as follows.

```
c5edc6a Add ability to create and retrieve box names
c9a623e Add vmpu_unpriv_uint8_[read/write]
354db0d Use [ld/st]rht for vmpu_unpriv_uint16_[read/write]
df99c14 Add ability to get calling box ID
e2faf64 Add ability to read current box ID
471870a vmpu_exports: Add missing header dependency
3838ccf vmpu: Add missing header dependency
2981324 Rename `box_name` to `box_prefix`
3609486 Allow ACL to be optional with unsupported targets
```

Patater:
- Maintain source-level backwards compatibility (don't force the adding of a name)
- Return a locally-unique identifier for boxes with no name
- Make sure box-id is non-negative

meriac:
@Patater yes - the main box is consistent with other unnamed boxes: we don't have to explain a different story here.

Patater:
- Removed default name from Box 0

Patater:
- Fix Obiwan error
- Replace "flash" with "public flash" in halt message
- Use hex string representation of box struct address as locally unique identifier

Patater:
- Refactor box name retrieving into more functions

Patater:
Moving to fork-based PR.

PR #151, opened 2016-02-16
Fix the Kinetis Makefile for debug releases

AlessandroA:
The `OPT=` configuration was not properly passed to the sub-make hence
generating a release binary even with the `kinetis_d` make rule.

Fixes: 7d60c75 ("Re-add support for K64F after refactoring ")

@Patater


Patater:
LGTM

Issue #152, opened 2016-02-16
Not possible to release multiple configurations for a platform


akselsm:
With the current Makefile.rules, it seems that it isn't possible to release multiple configurations. If I have a Makefile like this for my platform:

```
CONFIGURATIONS:= \
    MY_CONFIG_1 \
    MY_CONFIG_2 \
    MY_CONFIG_3

my_platform_r:
	$(foreach CONFIGURATION, \
	          $(CONFIGURATIONS), \
	          CONFIGURATION=$(CONFIGURATION) make -f ../../core/Makefile.rules clean release && ) true
```

Then only the last configuration will be preserved in `release/source/<platform>/`, since the `release` make target performs `rm -rf $(RELEASE_SRC_HW)`.

If I understand the comments on the refactoring pull request + comments in source code correctly, it should be possible to call `make release` multiple times to output multiple binaries/configurations for the same platform.

AlessandroA:
Thanks a lot for spotting this! It seems pretty simple to fix, I'll issue a PR.

ciarmcom:
ARM Internal Ref: IOTSFW-2055

PR #153, opened 2016-02-16
Fix Makefiles for multiple platform configurations

AlessandroA:
A single Makefile rule should allow to build all available uVisor configurations for a given platform. This PR fixes 2 bugs that prevent that from happening:
* The release script deleted the release folder each time after a single make
* The `foreach` construct in the platform-specific Makefile pasted several make instructions together instead of running them sequentially.

Fixes: #152 ("Not possible to release multiple configurations for a platform")
(Thanks @akselsm)

@meriac 
@Patater


AlessandroA:
**Update**:

This has now been updated with new commits to generalize and improve the build process.

* No need any more to have custom rules in platform-specific Makefiles.
    * You only need to define the `CONFIGURATIONS`.
* The main Makefile provides two default rules, `debug` and `release`, that use the make-friendly dynamic rule creation to iterate over all configurations.
    * It also solves the problem with `foreach` (https://github.com/ARMmbed/uvisor/issues/152).

@meriac @Patater

PR #165, opened 2016-02-24
Add support for EFM32 families

akselsm:
Add support for Giant Gecko, Leopard Gecko, Wonder Gecko, Jade Gecko and Pearl Gecko.

The published versions of cmsis-core-*, mbed-hal-* and target-efm32*-* in the yotta registry all have uvisor support, but the config parameter "uvisor/present" is currently set to false, in order for things not to break until support is merged and published in uvisor-lib. Will need to be overridden on the yotta build command line, or in config.json, in order to test before merge.


akselsm:
All our Cortex-M3s are r2p1 or newer, so we have the SCnSCB->ACTRL register. I have updated the PR.

AlessandroA:
@akselsm Thanks for the quick update! Also, apologies for the change-of-plans but after an offline discussion with @meriac we realized we actually still need to differentiate between Cortex-M3 and M4. 

Could you please re-introduce the double configuration M3/M4? I'll create a PR today that reflects the changes for the CMSIS header file inclusion.

Finally, if you can squash the two commits together and make sure the commit message stays in the [50/72 limits](http://chris.beams.io/posts/git-commit/#seven-rules) we are ready to merge :)

akselsm:
No problem :) Now it's done.

platform/efm32/inc/configurations.h
```diff
@@ -0,0 +1,60 @@
...
+/*******************************************************************************
+ * Hardware-specific configurations
+ *
+ * Configurations are named after the parameter values, in this order:
+ *   - CORE
+ *   - PLATFORM
+ ******************************************************************************/
+#if defined(CONFIGURATION_EFM32_M3_P1)
+    /* ARM core selection */
+#   define CORE_CORTEX_M3
```
AlessandroA: Could you move `NVIC_VECTORS` and `FLASH_OFFSET` under "Universal configurations"? Even if there might be future configurations that require to have differentiated values for those symbols, I'd like to have current configurations only rely on the differentiating one.

As a general approach, I think `NVIC_VECTORS` will always be one single value across a family (defined as the max of all values across the family). An exception might be families with such a large variance that might benefit in having two configurations (for wasting less memory). For example:

> In the family, the following number of NVIC vectors are available: 17, 19, 21, 60, 76, 84. I'd have two configurations with `NVIC_VECTORS` set to 21 (for {17, 19, 21}) and 84 (for {60, 76, 84}), respectively (local max on a cluster).

platform/efm32/inc/configurations.h
```diff
@@ -0,0 +1,60 @@
...
+/*******************************************************************************
+ * Universal configurations
+ ******************************************************************************/
+
+/* Memory boundaries */
+#define FLASH_ORIGIN 0x00000000
+
+/* Memory boundaries */
+#define SRAM_ORIGIN  0x20000000
+#define SRAM_OFFSET  0x0
```
AlessandroA: If I remember correctly, in your mbed port you don't allow vector relocation to SRAM -- and `SRAM_OFFSET` here is 0 for that reason. Could you confirm this?

platform/efm32/inc/configurations.h
akselsm: I see what you mean, and yeah, it's done for future-proofing for the exact reason you specify. If I make these defines universal now, I'd expect to have to revert that in the not-too-distant future. But if that makes it cleaner in the mean time, I'm open to doing that.

platform/efm32/inc/configurations.h
akselsm: We allow vector relocation to SRAM, but we're relocating to a position _after_ the uvisor bss (`__start_vector_table__` in the .data section). This is because we need the DMA descriptors to end up before the vector table, since the ARM UDMA IP has strict alignment requirements.

platform/efm32/inc/configurations.h
meriac: I would leave it as it is - it's probably more readable as its now.

platform/efm32/inc/configurations.h
AlessandroA: Thanks for the clarification!

platform/efm32/inc/configurations.h
AlessandroA: Yes @akselsm, if possible please move it. It helps with consistency and it states more clearly what is that make every configuration unique (in this case, only the core version).

I think the revert will be very easy when the time comes. In that occasion, it will be natural to move a universal define to a config-specific one.

meriac: @akselsm I agree with @AlessandroA now - for the time being it's important for us to establish best practice for future ports. The patch should reflect the best case for existing platforms. Once you're requirements change, it's easy to change back. The readers of these future changes will understand better the reasoning.

akselsm: Done.

AlessandroA:
@meriac LGTM

PR #167, opened 2016-02-24
Redefine SVCall imm8 field bits

AlessandroA:
This commit changes the way the 8 bit immediate field in an `svc` opcode
is used to store information about the SVCall.

The 8 bits are now used as follows:

```
 For fast APIs:

   7   6   5   4   3   2   1   0
 .---.---.---.---.---.---.---.---.
 | 1 | N | N | N | h | h | h | h |
 '---'---'---'---'---'---'---'---'
   |   |_______|   |___________|
   |       |         |
   |       |         |
   |       |         |
   |       |        (h) Handler index in the hardcoded table (max 16)
   |       |
   |      (N) Number of arguments to pass to the handler (max 4)
   |
   SVCall mode: Fast

 For slow APIs:

   7   6   5   4   3   2   1   0
 .---.---.---.---.---.---.---.---.
 | 0 | c | c | c | c | c | c | c |
 '---'---'---'---'---'---'---'---'
   |   |_______________________|
   |               |
   |               |
   |              (c) Handler index in the custom table (max 128)
   |
   SVCall mode: Slow
```

This increases the number of available entries for the custom SVCall
handler table, from 16 to 128.

@meriac @Patater


core/system/inc/svc_gw.h
```diff
@@ -39,8 +39,7 @@ static inline void svc_gw_check_magic(TSecGw *svc_pc)
         HALT_ERROR(SANITY_CHECK_FAILED,
                    "secure gateway not in flash (0x%08X)", svc_pc);
 
-    if((svc_pc->opcode & ~((uint16_t) UVISOR_SVC_CUSTOM_MSK)) !=
-        UVISOR_SVC_GW_OPCODE)
+    if((svc_pc->opcode & ~((uint16_t) UVISOR_SVC_FAST_NARGS_MASK)) != UVISOR_SVC_GW_OPCODE)
```
meriac: @AlessandroA why the uint16_t cast ?

core/system/src/svc_cx.c
```diff
@@ -95,7 +95,7 @@ uint32_t __svc_cx_switch_in(uint32_t *svc_sp, uint32_t svc_pc,
     uint32_t           dst_sp_align;
 
     /* number of arguments to pass to the target function */
-    uint32_t args = (uint32_t) (svc_id & UVISOR_SVC_CUSTOM_MSK);
+    uint32_t args = (uint32_t) (svc_id & UVISOR_SVC_FAST_NARGS_MASK);
```
meriac: args needs shifting ?

core/system/inc/svc_gw.h
AlessandroA: Because the full opcode is 16 bits, while the `NARGS` mask is 8 bits (only covers imm8 field of `svc`).

If `NARGS_MASK` is `0x70` (8 bits) I want it to become `0xFF8F` (16 bits).

AlessandroA:
Updated the following:
* Fixed args shifting in `svc_cx.c`.
    * Also added a check on args.
* Changed names in `SVCall_IRQn_Handler` to be more explicit and close to the corresponding symbol.

Missing points to discuss:
* `uint16_t` casting (see my comment)
* Masking with `0xFF` (see my comment)

@meriac @Patater

meriac:
Please split the formatting into a separate patch @AlessandroA

AlessandroA:
Update:
* Fixed "allowwed".
* Removed unneeded cast to `uint32_t`.
* Split in 2 commits (one for formatting).
* Removed all occurrences of `& 0xFF` from masks themselves and only added to macros explicitly used as 8 bits immediates.

Please wait for merge. I'll test it on hardware in the morning.

AlessandroA:
Update:

* Moved `& 0xFF` masking right at the asm inline point
* Reserved 0th entry in custom table as not implemented (will be used for `uvisor_get_version()`)

@meriac

PR #169, opened 2016-02-24
Makefile: Create new platform BUILD_MODE folders

Patater:
Automatically create new platform BUILD_MODE (debug, release) folders in
the uvisor-lib. Without this, the uvisor build will fail for new platforms.

@AlessandroA 
@meriac


AlessandroA:
LGTM

PR #172, opened 2016-02-26
Workaround for bug in memory access macros

AlessandroA:
This is a temporary workaround - #173 has been created to keep track of the original bug. 

If optimizations are turned on, the r0 register, in which the address
for the memory access is stored, might contain garbage. The signature
for the `uvisor_{read, write}{8, 16, 32}` APIs only specified a volatile
data-pointed-to, but not a volatile pointer.

@meriac @Patater

### DIFF
--- a/core/mbed/uvisor-lib/secure_access.h
+++ b/core/mbed/uvisor-lib/secure_access.h
@@ -24,13 +24,13 @@
         switch(sizeof(type)) \
         { \
             case 4: \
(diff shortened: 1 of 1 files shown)


Patater:
LGTM

AlessandroA:
Marked as a temporary workaround. #173 has been created to keep track of the original bug. @meriac

PR #175, opened 2016-02-26
Implement Box ID API

Patater:
Provide the ability to query:

- Numerical box ID of one's own box
- Numerical box ID of an inbound (across box boundary) caller
- String ID for box via numerical box ID

Provide the ability to set a box namespace. This change breaks binary compatibility and source-level backwards compatibility, as it forces a user of uVisor to specify a `UVISOR_BOX_NAMESPACE(NULL)` or `UVISOR_BOX_NAMESPACE("com.client.some.name.here")` before `UVISOR_BOX_CONFIG`.

This has been tested on frdm-k64f-gcc with a modified uvisor-helloworld. The challenge box can correctly identify itself and the box id of the calling secure gateway when verifying its secret. The calling secure gateway is box 0, an anonymous box. Likewise, box 0 is also able to correctly identify itself and reports "-1" as the calling box (since box 0 wasn't called through a secure gateway).

@AlessandroA
@meriac


core/mbed/uvisor-lib/box_config.h
```diff
@@ -48,35 +53,35 @@ UVISOR_EXTERN const uint32_t __uvisor_mode;
 /* this macro selects an overloaded macro (variable number of arguments) */
 #define __UVISOR_BOX_MACRO(_1, _2, _3, _4, NAME, ...) NAME
 
-#define __UVISOR_BOX_CONFIG(box_name, acl_list, acl_list_count, stack_size, context_size) \
```
meriac: @Patater please keep `box_name` as it refers to the "variable name" of  the box. `uvisor_box_name_` on the other hand refers to the box prefix, so __uvisor_box_prefix makes more sense in this context.

Patater: This was changed so that we would avoid confusion (cognitive overhead) between the box name as provided by UVISOR_BOX_NAME and the prefix we stick on the front of variables declared or defined by the macro.

Patater:
- Rename box_name to box_namespace
- Fix Obiwan error
- Add back a newline that was deleted on accident

Patater:
- Use new error defines for returning errors from uvisor_box_namespace

meriac:
LGTM

Patater:
- Make sure copyright header years are updated in files that changed

AlessandroA:
LGTM

Patater:
- Remove an obiwan error
- Correctly print the box_id when the box namespace is too long
- Rename `name` to `box_namespace` in cfgtbl
- Use `int` for box_id type consistently

Patater:
- Don't halt when the box ID provided to `uvisor_box_namespace_from_id` is invalid. Return an error instead.

Patater:
- A box ID of 0 is valid. Make the vmpu_is_box_id_valid function reflect this fact.

meriac:
LGTM

PR #176, opened 2016-02-26
Generalize platform-specific static MPU initialization (ARMv7-M)

AlessandroA:
This PR generalizes the configuration of static MPU regions in platforms that use the ARMv7-M MPU.

Two configurations are available:
1. uVisor shares the SRAM with the host. [default]
2. uVisor is placed in standalone SRAM.

As stated in the docs, symbols like `FLASH_ORIGIN`, `SRAM_ORIGIN`, etc., always refer to
> [...] the boundaries of the physical memories where uVisor is placed (Flash, SRAM).

So if the uVisor is placed in a standalone SRAM we also need to know the boundaries of the host SRAM, so that we can open an MPU region for it.

That's why the `HOST_SRAM_ORIGIN_MIN` and `HOST_SRAM_LENGTH_MAX` have been introduced. The `_MIN` and `_MAX` make sure that with a one-time only configuration a region is opened for the whole SRAM for all configurations (no need to differentiate binaries on this). The region for the SRAM can be larger than the physical memory.

The vMPU module detects automatically the configuration (shared, standalone) and sets up the MPU accordingly.

@meriac @Patater


AlessandroA:
@meriac Updated PR with:
* Added check on `UVISOR_IN_STANDALONE_SRAM`.
* Added sanity check on memory lengths.

PR #178, opened 2016-03-01
Add premilinary support for debug boxes

AlessandroA:
This PR includes the following:
* Version 0 of a driver specification for a debug box. This includes
  only 2 handlers: `get_version(void)` and `halt_error(int reason)`.
* A prototype implementation of debug box back-end functionality:
    * A box can register itself on a first-come-first-served basis as a
      debug box. A struct containing the handlers for a driver
      specification is passed to uVisor during registration.
    * uVisor is able to de-privilege and execute handlers from the debug
      box driver. Currently it cannot return from them (but it can
      reboot after them).

@meriac @Patater


Patater:
Do "Fix layout alignment" first.

Document the motivation behind "Increase Flash usage constraint for uVisor". Don't capitalize "Flash". If the increase in size is required and 100% motivated by "Add preliminary support for debug boxes", then increase the size as a part of that commit. If there is some other motivation, but the increase is still required by "Add preliminary support for debug boxes", increase the size first so that all the requirements are in order before they are required.

AlessandroA:
* "Fix layout alignment" now is the first commit.
* Flash constraints included in "Add preliminary support for debug boxes"

core/debug/inc/debug.h
```diff
@@ -18,8 +18,19 @@
 #define __DEBUG_H__
 
 #include <uvisor.h>
+#include "debug_exports.h"
 #include "halt.h"
 
+/* Debug box handle.
+ * This is use internally by uVisor to keep track of the registered debug box
```
Patater: used

core/debug/inc/debug.h
```diff
@@ -18,8 +18,19 @@
...
+ * and access its driver when needed. */
```
Patater: to access

core/debug/inc/debug.h
```diff
@@ -18,8 +18,19 @@
...
+typedef struct TDebugBox {
+    const TUvisorDebugDriver *driver;
+    uint8_t box_id;
+    int initialized;
+} UVISOR_PACKED TDebugBox;
```
Patater: Why is this packed? Add a comment in the code explaining why.

core/debug/inc/debug.h
```diff
@@ -33,6 +44,10 @@ void debug_fault_mpu(void);
 void debug_fault(THaltError reason, uint32_t lr, uint32_t sp);
 void debug_mpu_config(void);
 void debug_map_addr_to_periph(uint32_t address);
+void debug_register_box(const TUvisorDebugDriver * const driver);
```
Patater: Would `debug_register_driver` be more accurate or preferred?

core/debug/inc/debug_exports.h
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef __DEBUG_EXPORTS_H__
+#define __DEBUG_EXPORTS_H__
+
+#include <stdint.h>
+#include "uvisor_exports.h"
```
Patater: Move `#include <stdint.h>` after uvisor_exports.h

core/debug/inc/debug_exports.h
```diff
@@ -0,0 +1,34 @@
...
+
+/* Debug box driver -- Version 0
+ * A constant instance of this struct must be instantiated by the unprivileged
+ * code to setup a debug box.*/
+typedef struct TUvisorDebugDriver {
+    void *get_version;
+    void *halt_error;
```
Patater: Why not use function pointers?

core/debug/src/debug.c
```diff
@@ -230,3 +230,117 @@ void debug_fault(THaltError reason, uint32_t lr, uint32_t sp)
...
+
+void debug_reboot(void)
+{
+    if (!g_debug_box.initialized || g_active_box != g_debug_box.box_id) {
+        HALT_ERROR(NOT_ALLOWED, "This function can only be called from the context of an initialized debug box.\n\r");
+    }
+
+    /* Reboot.
+     * This can only be called from privileged code. */
+    NVIC_SystemReset();
```
Patater: If `NVIC_SystemReset` can only be called from privileged code, and `debug_reboot` can only be called from the context of a unprivileged, initialized debug box, why are we calling it?

I think this comment is lying. We can call the function from unprivileged code, but an exception occurs and uVisor traps it and reboots.

Please update the comment to be more clear in explaining what is happening.

core/debug/src/debug.c
```diff
@@ -230,3 +230,117 @@ void debug_fault(THaltError reason, uint32_t lr, uint32_t sp)
...
+}
+
+/* FIXME: Currently it is not possible to return to a regular execution flow
+ *        after the execution of the debug box handler. It is possible to
+ *        reboot, though. */
```
Patater: It should be possible to return. Are we making a decision not to implement such a capability right now?

core/debug/src/debug.c
```diff
@@ -230,3 +230,117 @@ void debug_fault(THaltError reason, uint32_t lr, uint32_t sp)
...
+    __set_PSP((uint32_t) dst_sp);
+
+    /* Upon execution return debug_handler will be executed. Upon return from
+     * debug_handler, return_handler will be executed. */
+    return;
+}
+
+void debug_get_version(void)
+{
+    HALT_ERROR(NOT_IMPLEMENTED, "This handler is not implemented yet. Only version 0 is supported.\n\r");
```
Patater: We should implement this before releasing anything.

core/debug/src/debug.c
```diff
@@ -230,3 +230,117 @@ void debug_fault(THaltError reason, uint32_t lr, uint32_t sp)
...
+    if (g_debug_box.initialized) {
+        HALT_ERROR(NOT_ALLOWED, "The debug box has already been initialized.\n\r");
+    }
+
+    /* Check the driver version. */
+    /* FIXME: Currently we cannot de-privilege, execute, and return to a
+     *        user-provided handler, so we are not calling the get_version()
+     *        handler. The version of the driver will be assumed to be 0. */
+
+    /* Chech that the debug driver table and all its entries are in public
```
Patater: Czech

core/debug/src/debug.c
```diff
@@ -230,3 +230,117 @@ void debug_fault(THaltError reason, uint32_t lr, uint32_t sp)
...
+     * Flash. */
```
Patater: Don't capitalize flash.

core/debug/src/debug.c
```diff
@@ -230,3 +230,117 @@ void debug_fault(THaltError reason, uint32_t lr, uint32_t sp)
...
+    if (!vmpu_flash_addr((uint32_t) driver) || !vmpu_flash_addr((uint32_t) driver + sizeof(TUvisorDebugDriver))) {
+        HALT_ERROR(SANITY_CHECK_FAILED, "The debug box driver struct must be stored in public Flash.\n\r");
```
Patater: flash

core/debug/src/debug.c
```diff
@@ -230,3 +230,117 @@ void debug_fault(THaltError reason, uint32_t lr, uint32_t sp)
...
+    }
+    for (i = 0; i < DEBUG_BOX_HANDLERS_NUMBER; i++) {
+        if (!vmpu_flash_addr(*((uint32_t *) driver + i))) {
+            HALT_ERROR(SANITY_CHECK_FAILED, "Each handler in the debug box driver struct must be stored in public Flash.\n\r");
```
Patater: flash

core/debug/inc/debug.h
AlessandroA: It actually doesn't need to be. Fixed.

core/debug/inc/debug.h
```diff
@@ -18,8 +18,19 @@
...
 
+/* Debug box handle.
+ * This is use internally by uVisor to keep track of the registered debug box
+ * and access its driver when needed. */
+typedef struct TDebugBox {
+    const TUvisorDebugDriver *driver;
+    uint8_t box_id;
+    int initialized;
+} UVISOR_PACKED TDebugBox;
+TDebugBox g_debug_box;
```
Patater: This is a one-per-uVisor instance. Do we want to consider this being one-per-box and place it in the box context (box-local memory)? We'd also handle registration on a per-box (instead of per-uVisor) basis.

core/system/src/svc.c
```diff
@@ -36,24 +37,28 @@ void __svc_not_implemented(void)
...
+    benchmark_start,            // 11
+    benchmark_stop,             // 12
+    halt_user_error,            // 13
+    unvic_irq_level_get,        // 14
+    vmpu_box_id_self,           // 15
+    vmpu_box_id_caller,         // 16
     vmpu_box_namespace_from_id, // 17
+    debug_reboot,               // 18
+    /* FIXME: This function will be made automatic when the debug box ACL is
+     *        introduced. The initialization will happend at uVisor boot time. */
```
Patater: happen

core/debug/inc/debug.h
AlessandroA: I would keep `debug_register_box` as it is used to register box number X as a debug box. The driver is a requirement in this process.

core/debug/src/debug.c
AlessandroA: Added comment that explains this better. This anyway *cannot* be called from unprivileged code. Unprivileged code will run `__debug_reboot` instead (without knowing -- It's part of the stack tampering process, see [this line](https://github.com/ARMmbed/uvisor/pull/178/files#diff-8866e0e2dcd8960e585a789e8e817ff1R288)). `__debug_reboot` then calls the proper SVCall.

core/debug/src/debug.c
AlessandroA: It's possible to return but it requires to develop the context switch *out*. I'd take this shortcut for the moment so that we have a smaller test surface. It's not trivial to perform the switch out, especially if we must take into consideration de-privileging from a fault.

core/debug/src/debug.c
AlessandroA: The reason this is not implemented is because of [this comment](https://github.com/ARMmbed/uvisor/pull/178/files#r54702805). This function is a deprivilege-execute-return-reprivilege one, so it is not currently supported.

core/debug/inc/debug.h
```diff
@@ -18,8 +18,19 @@
...
 #include <uvisor.h>
+#include "debug_exports.h"
 #include "halt.h"
 
+/* Debug box handle.
+ * This is used internally by uVisor to keep track of the registered debug box
+ * and to access its driver when needed. */
+typedef struct TDebugBox {
+    const TUvisorDebugDriver *driver;
+    uint8_t box_id;
```
meriac: @AlessandroA please use the box_id to the end or make it int-sized for alignment purposes.

core/debug/inc/debug.h
meriac: @AlessandroA I agree with @Patater here - 'debug_register_driver' is more appropriate as the user registers a non-uVisor aware debug drivers. It's just pure coincidence that the driver runs in a debug box and that box requires additional ACLs to do that call.

AlessandroA: Ok!

core/debug/src/debug.c
```diff
@@ -230,3 +230,123 @@ void debug_fault(THaltError reason, uint32_t lr, uint32_t sp)
...
+
+/* This function must be called as an SVCall handler.
+ * All debug handlers that are required to reboot upon exit should use the
+ * __debug_reboot function as return value, which triggers the SVCall. This note
+ * applies to uVisor internally, as the actual debug box does not need to care
+ * about this. */
+void debug_reboot(void)
+{
+    if (!g_debug_box.initialized || g_active_box != g_debug_box.box_id) {
+        HALT_ERROR(NOT_ALLOWED, "This function can only be called from the context of an initialized debug box.\n\r");
```
meriac: Please add a FIXME for returning an error code in case an unauthorised wrong box calls this function. Potentially add a JIRA backlog-ticket (Q3?) for a reset-ACL to allow individual boxes to reset the system.

core/debug/inc/debug_exports.h
AlessandroA: You're right. They also help with the driver registration (to check signatures). See also the updated [uvisor-helloworld example](https://github.com/ARMmbed/uvisor-helloworld/pull/60/files#diff-59b37b296cf946ab6951aba2ce0bee36R63).

core/debug/inc/debug.h
AlessandroA: Moved `box_id` to the end. I'd still keep it as `uint8_t` as that's what we use for box IDs.

AlessandroA:
* Fixed typos.
* Changed `debug_register_box` to `debug_register_driver`.
* Moved `box_id` to end of struct in `TDebugBox`.
* Turned `void *` into function pointers in the driver specifications.

core/debug/src/debug.c
meriac: @Patater - agreed with @AlessandroA - debug_get_version is a placeholder now for technical reasons. It might be helpful though to explain the reasoning in a FIXME-comment

AlessandroA:
* Added the FIXME for the halt on an unauthorized reboot request.

core/debug/inc/debug.h
Patater: Externally, we use int for box_id (since the Box ID API introduction). Internally, we currently use uint8_t.

AlessandroA:
* Added TODO that explains why the `debug_get_version` handler cannot be implemented yet.

meriac:
@AlessandroA I disagree with the notion that uVisor controls SP - as an attacker I can pretend any stack pointer at the time I am switched out.

meriac:
Merging, @AlessandroA - please add new github issue about verifying that all unprivileged access in uVisor uses helper functions.

PR #179, opened 2016-03-02
Keep object files for each configuration/build mode

AlessandroA:
Object files are now saved as:
```make
$(ORIGINAL_BASENAME).c.$(CONFIGURATION_LOWER).$(BUILD_MODE).o
```
in the same folder of the original file.

@Patater @meriac


Patater:
LGTM

PR #180, opened 2016-03-02
Remove duplication of g_active_box

AlessandroA:
A duplicated state was kept in functions from `svc_cx.*`, and it is now
removed.

The state that is now kept is `g_active_box`, which has been moved to
`svc_cx.*` as that is the place where context states are kept.

@meriac @Patater


Patater:
LGTM

PR #181, opened 2016-03-02
Introduce consistent use of vmpu_is_box_id_valid

AlessandroA:
This function is now consistently used to check if a box ID is valid. If the boxes have not been enumerated yet the check is done agains the maximum number of boxes allowed.

@meriac @Patater

### DIFF
--- a/core/system/inc/svc_gw.h
+++ b/core/system/inc/svc_gw.h
@@ -55,12 +55,13 @@ static inline uint32_t svc_gw_get_dst_fn(TSecGw *svc_pc)
 
 static inline uint8_t svc_gw_get_dst_id(TSecGw *svc_pc)
 {
-    uint32_t box_id = svc_pc->cfg_ptr - __uvisor_config.cfgtbl_ptr_start;
+    uint8_t box_id;
 
+    box_id = (uint8_t) ((svc_pc->cfg_ptr - __uvisor_config.cfgtbl_ptr_start) & 0xFF);
     if(box_id <= 0 || box_id >= g_vmpu_box_count)
         HALT_ERROR(SANITY_CHECK_FAILED, "box_id out of range (%i)", box_id);
 
-    return (uint8_t) (box_id & 0xFF);
+    return box_id;
 }
 
 #endif/*__SVC_GW_H__*/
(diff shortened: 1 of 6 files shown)


Patater:
Squish "Make vmpu_is_box_id_valid extern" with the commit that first adds code that requires `vmpu_is_box_id_valid` to be extern.

Make style changes in a separate commit from `Make vmpu_is_box_id_valid extern` (where you move the brace of the if statement).

Do "Move cast to uint8_t at the variable declaration" before any meaningful changes: style changes first, then refactoring changes, then semantic changes.

core/system/inc/svc_gw.h
```diff
@@ -55,12 +55,14 @@ static inline uint32_t svc_gw_get_dst_fn(TSecGw *svc_pc)
 
 static inline uint8_t svc_gw_get_dst_id(TSecGw *svc_pc)
 {
-    uint32_t box_id = svc_pc->cfg_ptr - __uvisor_config.cfgtbl_ptr_start;
+    uint8_t box_id;
 
-    if(box_id <= 0 || box_id >= g_vmpu_box_count)
```
Patater: This condition seems different from the `vmpu_is_box_id_valid` condition. Consider fixing it in a separate commit before refactoring to use `vmpu_is_box_id_valid`. This makes it obvious that the condition is changing.

Specifically, this wouldn't allow a box_id of 0, but `vmpu_is_box_id_valid` does allow a box_id of 0. This might be a bug deserving a fix in its own commit, or we might be introducing a bug by starting to use `vmpu_is_box_id_valid` here.

AlessandroA: Thank you for spotting this. Currently we do *not* allow secure gateways to box 0, so the condition must still hold (I'll change the check accordingly for this special case).

@meriac In general, are we sure we don not want to allow secure gateways to box 0? I think this restriction goes against the principle of "box 0 is just a box like all other boxes". Interrupts can be context switched to box 0 normally, for example. If the restriction is not needed for security purposes, I'd rather remove it.

AlessandroA: I've introduced the additional check on `!box_id` to keep consistency with the previous functionality. A comment explains that. I'd still like to know if we want to allow sgw to box 0 -- in that case I'd introduce a new commit with the change.

AlessandroA:
* `vmpu_is_box_id_valid` made `static inline`, instead of `extern`.
    * @Patater I'd leave this in a separate commit as it's done for performance purposes (added to the commit message)
* Moved layout changes first; the brace fix is in a separate commit.

There's still one open question about whether we want to support explicit secure gateways to box 0 or not.

@meriac @Patater

Patater:
I am fine with the commit organization.

PR #182, opened 2016-03-03
Provide a function to check if a buffer is within box context

Patater:
This is useful when a box needs to verify that certain inputs to it are
within or not within its own context storage.

@AlessandroA
@meriac

### DIFF
--- a/core/mbed/uvisor-lib/box_config.h
+++ b/core/mbed/uvisor-lib/box_config.h
@@ -18,8 +18,55 @@
 #define __UVISOR_LIB_BOX_CONFIG_H__
 
 #include <stddef.h>
+#include <stdbool.h>
 
 UVISOR_EXTERN const uint32_t __uvisor_mode;
+
+/* In an effort to hide the box-local variables that
+ * uvisor_buffer_overlaps_box_context needs from colliding with other boxes,
+ * and to avoid defining uvisor_buffer_overlaps_box_context inside a macro that
+ * would make uvisor_buffer_overlaps_box_context itself box-local, we privately
+ * predeclare those box-local variables as private. The exact mechanism of
+ * privacy differs between C and C++. */
+#ifdef __cplusplus
+/* In C++, we can't predeclare variables that will later be defined as static.
+ * So, we instead define the variables as extern, but in a private anonymous
+ * namespace. */
+namespace
+{
+    extern const uintptr_t __uvisor_ctx_addr;
+    extern const size_t __uvisor_ctx_size;
+}
+#else
+/* In C, we can predeclare variables that will later be defined as static. */
+const size_t __uvisor_ctx_size;
+const uintptr_t __uvisor_ctx_addr;
+#endif
+
+#ifdef __cplusplus
+/* In C++, we define the private variables that
+ * uvisor_buffer_overlaps_box_context needs inside a private anonymous
+ * namespace. */
+#define __UVISOR_DEFINE_CTX_SIZE(box_name) \
+    namespace \
+    { \
+        const size_t __uvisor_ctx_size = sizeof(box_name ## _reserved); \
+    }
+
+#define __UVISOR_DEFINE_CTX_ADDR \
+    namespace \
+    { \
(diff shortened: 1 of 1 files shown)


AlessandroA:
Is this an API that we expect to be used explicitly by users in a secure box? If so I don't like hiding the API inside the macro for the box configuration.

I'd rather put the function `uvisor_buffer_overlaps_box_context` in uvisor-lib as any other public API. When that function is called, `uvisor_ctx` contains the currently active box context pointer so it will work seamlessly for all boxes.

In the future, if this function is expanded to also check other memories (stack, heap), it may include an SVCall that asks uVisor to perform the remaining checks.

Patater:
This can't be a plain inline because uvisor_ctx is not declared in this header file. This header file only provides macros that declare uvisor_ctx, so we likewise can only provide macros that define inline functions that make use of uvisor_ctx.

Patater:
@AlessandroA
Is the current implementation not in uvisor-lib? I'm not sure where you are proposing to move it to, and if it could still be a static inline function if we moved it there.

AlessandroA:
Just declare is as `extern void *uvisor_ctx;`.

See [here](https://github.com/ARMmbed/uvisor/blob/master/core/mbed/source/disabled.cpp#L29)

AlessandroA:
But I see your point now, you wouldn't have access to the size, which you need. it's not ideal but I don't have a solution to that on top of my head.

AlessandroA:
@meriac Any comment? If not I would merge. I had comments about the API visibility but these can be solved in a second step. This PR currently blocks ARMmbed/uvisor-helloworld#61

meriac:
@Patater I don't see the value of stuffing `uvisor_buffer_overlaps_box_context` into `#define __UVISOR_DECLARE_BUFFER_OVERLAPS_BOX_CONTEXT`. Please remove - the `static inline` is probably good enough.

Patater:
The value of putting `uvisor_buffer_overlaps_box_context` into `__UVISOR_DECLARE_BUFFER_OVERLAPS_BOX_CONTEXT` is to avoid repeating the function body twice (once in `__UVISOR_BOX_CONFIG_CONTEXT` and once in `__UVISOR_BOX_CONFIG_NOACL`). There are other ways to solve the problem if this way feels bad.

meriac:
@Patater why would you declare uvisor_buffer_overlaps_box_context twice (__UVISOR_BOX_CONFIG_CONTEXT & __UVISOR_BOX_CONFIG_NOACL) - It's just a standalone function declaration. I might not see the problem you're addressing.

Patater:
The definition is done inside those macros because that's the only place where the declaration of uvisor_ctx is valid. Outside of those macros, uvisor_ctx is not defined, and uvisor_ctx's type is unknown.

Patater:
- Rework the implementation so that uvisor_buffer_overlaps_box_context is declared in box-config.h outside of box-local macro context
 - The tricks we have to do, which aren't portable between C and C++, to predeclare the box-local variables that uvisor_buffer_overlaps_box_context needs, make this approach, while effective in getting the definition of uvisor_buffer_overlaps_box_context outside of a macro, not worth it.

Patater:
For context, here is the error message we are trying to solve for the incorrect, naive implementation (https://github.com/Patater/uvisor/commit/6dcb15d4f545f77cce0131fc8ed3d9d1c1aaf5f2).

```
In file included from uvisor/release/uvisor-lib/uvisor-lib.h:38:0,
                 from uvisor/release/source/disabled.cpp:17:
uvisor/release/uvisor-lib/box_config.h: In function 'bool uvisor_buffer_overlaps_box_context(const void*, size_t)':
uvisor/release/uvisor-lib/box_config.h:116:48: error: 'uvisor_ctx' was not declared in this scope
     uintptr_t __uvisor_ctx_addr = (uintptr_t) &uvisor_ctx;
                                                ^
uvisor/release/uvisor-lib/box_config.h:121:33: error: '__uvisor_ctx_size' was not declared in this scope
             __uvisor_ctx_addr + __uvisor_ctx_size > buffer_addr);
```

Patater:
Drop this PR because these implementations are too annoying to read and not actually want we want functionally, anyway.

PR #183, opened 2016-03-03
Return the proper number of bytes copied when reading box namespaces

Patater:
@meriac 
@AlessandroA


Patater:
- Explained why the halt
- Used a + 1 instead of a ++ because _style_

core/system/src/mpu/vmpu.c
```diff
@@ -449,11 +449,15 @@ static int copy_box_namespace(const char *src, char *dst)
...
+            return bytes_copied + 1 /* Plus 1 for the terminating-null */;
         }
     }
 
-    return bytes_copied;
+    /* We did not find a terminating null in the src. The src has been verified
+     * in vmpu_box_namespace_from_id as being in the box config table. It is a
+     * programmer error if the namespace in the box config table is not
+     * null-terminated, so we halt. */
+    HALT_ERROR(SANITY_CHECK_FAILED, "box namespace missing terminating-null");
```
AlessandroA: I think the comment got lost with a rebase. Could you capitalize and punctuate the halt error message?

Patater:
- Capitalize the halt message. (I didn't add a period because the message is just a sentence fragment.)

AlessandroA:
Thanks for the edits! LGTM.

Patater:
XXX Don't merge yet.

Patater:
- Added commit to mark halt functions as `noreturn`, to get rid of a compile error

Patater:
OK to merge if LGTY

AlessandroA:
Could you use `UVISOR_NORETURN` for that? (also, please keep the function signature all in one line)

Patater:
What is our line limit? And if one function declaration goes over that limit, would you rather the functions be consistently on two line or that only the ones that are too long are on two lines?

Patater:
It appears now that many halt functions can return...

This might break some code.

Patater:
GitHub is showing the commits weirdly again. Look at the real branch with git to see what is going on, if confused.

- Dropped the UVISOR_NORETURN commit, because halt can maybe return (in the future)
- Rebased on origin/master

Patater:
I am concerned if we push the merge button that GitHub will pull in bad things. Discarding this PR in favor of a new one.

PR #184, opened 2016-03-03
Return the proper number of bytes copied when reading box namespaces

Patater:
@AlessandroA @meriac


AlessandroA:
LGTM

AlessandroA:
Is this tested? If so I'll merge.

Patater:
Yes, tested with the current version of example-uvisor-box-id which expects the terminating null to be included in the number of bytes.

PR #190, opened 2016-03-14
Fix check on flash boundary

AlessandroA:
We need to distinguish checks on the physical and the public flash,
respectively.

@meriac @Patater


AlessandroA:
* Added check on `driver == NULL` and `*(driver + i) == NULL`

AlessandroA:
* Added commit that checks both start and end address of `box_namespace`.

PR #193, opened 2016-03-15
Restore priority bits in NVIC register

AlessandroA:
The register was modified to check the number of implemented priority
bits but its original value was not restored.

@meriac
@Patater

### DIFF
--- a/core/system/src/unvic.c
+++ b/core/system/src/unvic.c
@@ -443,10 +443,18 @@ void __unvic_svc_cx_out(uint32_t *svc_sp, uint32_t *msp)
 
 void unvic_init(void)
 {
-    /* Detect the number of implemented priority bits. */
+    uint8_t prio_bits;
+    uint8_t volatile *prio;
+
+    /* Detect the number of implemented priority bits.
+     * The architecture specifies that unused/not implemented bits in the
+     * NVIC IP registers read back as 0. */
     __disable_irq();
-    *((volatile uint8_t *) &(NVIC->IP[0])) = 0xFFU;
-    g_nvic_prio_bits = (uint8_t) __builtin_popcount(*((volatile uint8_t *) &(NVIC->IP[0])));
+    prio = (uint8_t volatile *) &(NVIC->IP[0]);
+    prio_bits = *prio;
+    *prio = 0xFFU;
+    g_nvic_prio_bits = (uint8_t) __builtin_popcount(*prio);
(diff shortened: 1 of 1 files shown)


Patater:
What bad things happen if the original value isn't restored?

AlessandroA:
The IRQn slot that corresponds to that register would find its default priority messed up. This is not a real issue as we are/will be setting all IRQs to a default known state, but I think it's not good practice to modify a system register to detect a hardware feature *and then leave it modified*.

Patater:
Since we are setting all IRQs to a default known state, then this isn't necessary to do here. The code was probably written with that fact in mind.

However, in the uVisor disabled case, do we still reset all IRQs to a default known state?

AlessandroA:
No we don't (disabled case). In that instance the code really should go there.

When uVisor is enabled, this is not critical — in fact, it is not a problem at all. It's just a matter of style/good practice. @meriac ? If you agree with @Patater I'm happy to close this one (no merge).

core/system/src/unvic.c
```diff
@@ -443,10 +443,16 @@ void __unvic_svc_cx_out(uint32_t *svc_sp, uint32_t *msp)
...
+    uint8_t prio_bits;
+    uint8_t volatile *prio;
+
     /* Detect the number of implemented priority bits. */
     __disable_irq();
-    *((volatile uint8_t *) &(NVIC->IP[0])) = 0xFFU;
-    g_nvic_prio_bits = (uint8_t) __builtin_popcount(*((volatile uint8_t *) &(NVIC->IP[0])));
+    prio = (uint8_t volatile *) &(NVIC->IP[0]);
+    prio_bits = *prio;
+    *prio = 0xFFU;
```
meriac: @AlessandroA i would add a comment here on why we're doing the next two lines ("the architecture specifies that unused bits read back as zeroes" or similar)

AlessandroA: Added.

Patater:
If this is not performance critical code, I'm fine with leaving the restoration of the old value here, even though we initialize it soon after. Assumptions can and do change. Unwritten assumptions make things even worse. We'll have less brittle code if we do the restoration here.

Imagine if you copy pasted this code to some uVisor disabled initialization code... You'd have to realize what assumptions this code was written under and think if those assumptions were still valid, and then modify the code to meet the requirements of the code's new home.

PR #195, opened 2016-03-17
Implement g_nvic_prio_bits for UVISOR_DISABLED

AlessandroA:
When __uvisor_mode is UVISOR_DISABLED the NVIC APIs are used instead of
the vIRQ ones. Now that those APIs are pre-compiled into a library, the
NVIC APIs use the uVisor built-in definitions from cmsis/core_generic.h.

In that file __NVIC_PRIO_BITS is re-defined as a global variable which
is evaluated at start-up. This behaviour has now been mirrored to the
case of disabled uVisor.

@meriac @Patater 

**Note**: This was previously in a previous PR.

### DIFF
--- a/api/src/disabled.c
+++ b/api/src/disabled.c
@@ -21,6 +21,9 @@
 #include <stdint.h>
 #include <string.h>
 
+/* Number of implemented priority bits */
+uint8_t g_nvic_prio_bits;
+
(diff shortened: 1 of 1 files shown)


AlessandroA:
@meriac @Patater Commit message updated. The added part (Warning: ...) will be added to a GH issue.

PR #197, opened 2016-03-18
Add libraries for all configurations in api/lib

AlessandroA:
@meriac @Patater


Patater:
I ddin't check the binaries, but the versions look good to me.

AlessandroA:
:smile:

PR #198, opened 2016-03-24
Fix out-by-one error in vmpu_region_bits

nchong:
This function should compute ceil(log2(size)); the number of bits needed to
address a region of 'size' bytes. If size < 32 then this should be 5 bits
(since the minimum supported region size in v7M is 32 bytes). The function was
incorrect for this case (returning 6 bits) because it rounded-up *after* the
minimum size check.

The fix has been checked exhaustively by CBMC.


ciarmcom:

PR #199, opened 2016-04-01
Fix debug disabled

AlessandroA:
This ensures that applications can run in debug mode with uVisor
disabled. Semihosting is still a requirement for applications that run
with uVisor enabled.

@meriac @Patater

### DIFF
--- a/core/system/src/main.c
+++ b/core/system/src/main.c
@@ -58,6 +58,14 @@ UVISOR_NOINLINE void uvisor_init_post(void)
 
 void main_entry(void)
 {
+    /* Return immediately if the magic is invalid or uVisor is disabled.
+     * This ensures that no uVisor feature that could halt the system is
+     * active in disabled mode (for example, printing debug messages to the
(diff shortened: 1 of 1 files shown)


Patater:
LGTM

AlessandroA:
@meriac @Patater Updated after offline conversation. Now `main_entry` returns immediately if the magic is wrong or uVisor is disabled. The magic is still checked in the vMPU sanity checks.

core/system/src/main.c
```diff
@@ -58,6 +58,14 @@ UVISOR_NOINLINE void uvisor_init_post(void)
 
 void main_entry(void)
 {
+    /* Return immediately if the magic is invalid or uVisor is disabled.
+     * This ensure that no uVisor feature that could halt the system is
```
Patater: s/ensure/ensures

Patater:
LGTM

AlessandroA:
Thanks! Fixed.

PR #200, opened 2016-04-04
Simplify uvisor lib

AlessandroA:
This PR ensures that internally uVisor does not rely on any yotta-specific symbol. Instead, it is the job of the host-OS glue layer (uvisor-lib) to define the relevant symbols based on OS-specific information.

Currently the only 2 symbols that are used for this are:
* `UVISOR_PRESENT`.
* `ARCH_MPU_${ARCH_MPU}`.

**Example**:

In mbed OS the symbol `UVISOR_PRESENT` is mapped to `YOTTA_CFG_UVISOR_PRESENT`. The symbol `ARCH_MPU_*` depends on the target, which can be inferred using `TARGET_LIKE_*` from the config symbols.

See the [related PR](https://github.com/ARMmbed/uvisor-lib/pull/43) to check how this is implemented in uvisor-lib.

@meriac @Patater @niklas-arm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,6 +152,7 @@ CFLAGS_PRE:=\
         $(OPT) \
         $(DEBUG) \
         $(WARNING) \
+        -DUVISOR_PRESENT \
         -DARCH_MPU_$(ARCH_MPU) \
         -D$(CONFIGURATION) \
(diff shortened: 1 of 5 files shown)


Patater:
LGTM

PR #204, opened 2016-04-13
[wip] Add page allocator to uVisor core

niklas-arm:
This adds a page allocator to uVisor core with a max. of 64 pages.
Page size is defaulted to 16kB, this may need to be adapted, depending on the application type (how do we do that?).

Areas for review:
- correct validation of inputs,
- a change to the config wrapper!
- this is missing any actual protection code for securing pages against access from boxes!


core/system/src/page_allocator.c
```diff
@@ -0,0 +1,181 @@
...
+ */
+#include <uvisor.h>
+#include "page_allocator.h"
+
+
+/* Must be UVISOR_PAGE_SIZE for MPU alignment in ARMv7-M */
+#define UVISOR_PAGE_ALIGNMENT UVISOR_PAGE_SIZE
+
+/* We can only have a total of 64 pages for now */
+#define UVISOR_PAGE_TABLE_COUNT ((size_t)64)
```
meriac: @niklas-arm add a #ifndef UVISOR_PAGE_TABLE_COUNT around that to make it overrideable

niklas-arm: Hm, currently it's not completely implementation independent, in `page_free` there is this (ugly) bit mask:
```
 /* contains the bit mask of all the pages we need to free */
uint64_t free_mask = 0;
```
The idea was that you only compute the pages to be freed once, and store it in a bitmask.
I can rewrite it to compute the page index on the fly again from the page origin, that would make this implementation independent.

PR #208, opened 2016-04-29
Updates unsupported target code.

niklas-arm:
Code will be copied into uvisor-lib, see https://github.com/ARMmbed/uvisor-mbed-lib/pull/6.

cc @Patater @meriac @AlessandroA


AlessandroA:
+1

PR #210, opened 2016-05-04
Add support for IAR

c1728p9:
Add IAR compiler intrinsics, implement vmpu_bits for IAR and update
the macro ADDRESS_WRITE to be compatible with IAR.

Issue #173, opened 2016-02-26
Memory access macros do not work when optimization is turned on


AlessandroA:
**Bug description**
On the NXP FRDM-K64F it has been observed that turning on compiler
optimizations interferes with the content of the `r0` register in memory access
macros.

Although `r0` is in the register list of the inline asm used for the
`uvisor_write32()` API, it contains garbage at the moment of use. The bug is
observed in the form of a hard fault due to an invalid address (`r0`) in a `str`
instruction.

**To reproduce**
Create a yotta application that uses the `frdm-k64f-gcc` target. Use the
following source code:

``` C
#include "mbed-drivers/mbed.h"

PwmOut led(D5);

void app_start(int, char **) {
    led = 0.5f;
}
```

and compile in release mode with `yotta build -r`. Run the program and observe the hard fault.
The fault comes from the `pwmout_write` function, in which a series of optimizations leads
to the following `str` instruction:

``` asm
str r1, [r0]
dsb
nop
nop
nop
```

where `r0 = 0x0`. The instructions above implement the `uvisor_write32()` API.

AlessandroA:
Notes:
* uVisor does not need to be enabled to observe this bug.
* The bug is currently only observed on the K64F as that platform's low level APIs boil down to calls to uVisor memory access macros.

ciarmcom:
ARM Internal Ref: IOTSFW-2133

mjs-arm:
Here is  a (almost) minimal reproducer of the issue.  This code was reduced directly from the failing code above by folding away the various layers of macro and inline function:

typedef long unsigned int uint32_t ;
void foo(float value)
{
  register uint32_t r0 __asm__("r0") = 0x40038000u;
  register uint32_t r1 __asm__("r1") = value;
    __asm__ volatile( "str"" %[r1], [%[r0]]\n"
                      :
                      : [r0] "r" (r0),
                        [r1] "r" (r1));
}

The issue can be provoked in gcc-4.9 using the above code, the load of the address into r0 is corrupted by the injection of library call to truncate the float to an integer during the setup of r1.

Relevant parts of the gcc manual are:
https://gcc.gnu.org/onlinedocs/gcc-5.3.0/gcc/Extended-Asm.html#Extended-Asm
https://gcc.gnu.org/onlinedocs/gcc-5.3.0/gcc/Local-Reg-Vars.html#Local-Reg-Vars

The local register var extension in gcc is 'fragile', note the following guidance from the documentation link above:

[...]

PR #211, opened 2016-05-05
Fix compiler attributes and functions for IAR and unsupported targets

AlessandroA:
This PR is a modification of [this previously filed PR](https://github.com/ARMmbed/uvisor/pull/210) by @c1728p9 .

It introduces the following changes:
* `ADDRESS_WRITE` is now a static inline function.
* Compiler attributes have been split into core-only and shared.
    * `UVISOR_FORCEINLINE` is now conditionally defined based on the toolchain.
* `vmpu_bits` is only defined with `UVISOR_PRESENT=1`.

@meriac @Patater @niklas-arm @c1728p9 

@c1728p9 Please it would help if you could test this on IAR as we don't have access to it yet.

### DIFF
--- a/api/inc/secure_access.h
+++ b/api/inc/secure_access.h
@@ -19,42 +19,8 @@
 
 #include "api/inc/uvisor_exports.h"
 #include "api/inc/vmpu_exports.h"
+#include <stddef.h>
 #include <stdint.h>
-
-/* the switch statement will be optimised away since the compiler already knows
- * the sizeof(type) */
-#define ADDRESS_WRITE(type, addr, val) \
-    { \
-        switch(sizeof(type)) \
-        { \
-            case 4: \
-                uvisor_write32((uint32_t volatile * volatile) (addr), (uint32_t) (val)); \
-                break; \
-            case 2: \
-                uvisor_write16((uint16_t volatile * volatile) (addr), (uint16_t) (val)); \
-                break; \
-            case 1: \
-                uvisor_write8((uint8_t volatile * volatile) (addr), (uint8_t ) (val)); \
-                break; \
-            default: \
-                uvisor_error(USER_NOT_ALLOWED); \
-                break; \
-        } \
-    }
-
-/* the conditional statement will be optimised away since the compiler already
- * knows the sizeof(type) */
-#define ADDRESS_READ(type, addr) \
-    (sizeof(type) == 4 ? uvisor_read32((uint32_t volatile * volatile) (addr)) : \
-     sizeof(type) == 2 ? uvisor_read16((uint16_t volatile * volatile) (addr)) : \
-     sizeof(type) == 1 ? uvisor_read8((uint8_t volatile * volatile) (addr)) : 0)
-
-#define UNION_READ(type, addr, fieldU, fieldB) \
-    ({ \
-        type res; \
-        res.fieldU = ADDRESS_READ(type, addr); \
-        res.fieldB; \
-    })
 
 static UVISOR_FORCEINLINE void uvisor_write32(uint32_t volatile * volatile addr, uint32_t val)
 {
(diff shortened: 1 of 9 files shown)


api/inc/secure_access.h
```diff
@@ -21,69 +21,70 @@
...
+        case 1:
+            uvisor_write8((volatile uint8_t * volatile) addr, (uint8_t) val);
+            break;
+        default:
+            uvisor_error(USER_NOT_ALLOWED);
+            break;
+    }
+}
+
+#define ADDRESS_WRITE(type, addr, val) __address_write((int) sizeof(type), (volatile uint32_t *) addr, (uint32_t) val)
```
Patater: If `sizeof_type` is `size_t`, you can get rid of the `int` cast.

Patater:
LGTM. Nothing I commented about would be a blocker.

AlessandroA:
* Removed cast to `int`.
* Added volatile keyword where missing.

Patater:
LGTM

api/inc/secure_access.h
```diff
@@ -86,4 +52,40 @@ static UVISOR_FORCEINLINE uint8_t uvisor_read8(uint8_t volatile * volatile addr)
     return UVISOR_ASM_MEMORY_ACCESS(ldrb, uint8_t, addr);
 }
 
+/* The switch statement will be optimised away since the compiler already knows
+ * the sizeof_type. */
+static UVISOR_FORCEINLINE void __address_write(size_t sizeof_type, volatile uint32_t *addr, uint32_t val)
```
meriac: @AlessandroA : are you confident that a void return type works with IAR ?

api/inc/secure_access.h
```diff
@@ -86,4 +52,40 @@ static UVISOR_FORCEINLINE uint8_t uvisor_read8(uint8_t volatile * volatile addr)
...
+            break;
+    }
+}
+
+#define ADDRESS_WRITE(type, addr, val) __address_write(sizeof(type), (volatile uint32_t *) addr, (uint32_t) val)
+
+/* the conditional statement will be optimised away since the compiler already
+ * knows the sizeof(type) */
+#define ADDRESS_READ(type, addr) \
+    (sizeof(type) == 4 ? uvisor_read32((uint32_t volatile * volatile) (addr)) : \
```
meriac: @AlessandroA : turn that into an inline with switch statement, too for consistency ?

api/inc/unsupported.h
```diff
@@ -133,63 +134,64 @@ static inline UVISOR_FORCEINLINE void uvisor_write(uint32_t addr, uint32_t val,
...
+     sizeof(type) == 2 ? uvisor_read16((volatile uint16_t * volatile) (addr)) : \
+     sizeof(type) == 1 ? uvisor_read8((volatile uint8_t * volatile) (addr)) : 0)
+
+/* The switch statement will be optimised away since the compiler already knows
+ * the sizeof_type. */
+static UVISOR_FORCEINLINE void __address_write(size_t sizeof_type, volatile uint32_t *addr, uint32_t val)
+{
+    switch(sizeof_type) {
+        case 4:
+            uvisor_write32((volatile uint32_t * volatile) addr, (uint32_t) val);
```
c1728p9: Why are both the pointer and the value being pointed to volatile?  Only the value being pointed to is in volatile memory.  This also introduces the following warning on IAR:
[Warning] unsupported.h@180: [Pe191]: type qualifier is meaningless on cast type

api/inc/uvisor_exports.h
```diff
@@ -34,15 +34,15 @@
...
+#else
+#define UVISOR_FORCEINLINE inline __attribute__((always_inline))
+#endif
 #define UVISOR_PACKED      __attribute__((packed))
 #define UVISOR_WEAK        __attribute__((weak))
-#define UVISOR_ALIAS(f)    __attribute__((weak, alias (#f)))
-#define UVISOR_LINKTO(f)   __attribute__((alias (#f)))
 #define UVISOR_NORETURN    __attribute__((noreturn))
-#define UVISOR_NAKED       __attribute__((naked))
 #define UVISOR_RAMFUNC     __attribute__ ((section (".ramfunc"), noinline))
```
c1728p9: These attributes are invalid for IAR.  You'll need to define them inside __ICCARM__ as well

c1728p9:
I gave the latest uvisor-mbed-lib a try, and I'm still getting errors.  I think this patch missed a bit from the one it was based on.

api/inc/unsupported.h
AlessandroA: This is due to an issue, reported on GCC. Without the `volatile` compiler optimizations for the read/write functions were preventing our inline assembly code to be compiled as desired. It was fixed in #172 and has #173 has been opened to keep track of it.

But now that I think about it, the change here was only introduced for consistency with the supported case — it's not really needed and it shouldn't be there. I will remove this in a new PR.

PR #215, opened 2016-05-12
WIP: Add OS private BSS

Patater:

This goes along with an mbed-os change to the linker script. Here is what the uvisor.bss section of the K64F linker script could look like.

``` diff
    /* Ensure that the uVisor .bss is the first section in the public SRAM. */                                                                                                                                                                                                                                                
    .uvisor.bss (NOLOAD):                                                                                                                                                                                                                                                                                                     
    {                                                                                                                                                                                                                                                                                                                         
        . = ALIGN(32);                                                                                                                                                                                                                                                                                                        
        __uvisor_bss_start = .;                                                                                                                                                                                                                                                                                               

        /* uVisor main .bss (protected) */                                                                                                                                                                                                                                                                                    
        . = ALIGN(32);                                                                                                                                                                                                                                                                                                        
        __uvisor_bss_main_start = .;                                                                                                                                                                                                                                                                                          
        KEEP(*(.keep.uvisor.bss.main))                                                                                                                                                                                                                                                                                        
        __uvisor_bss_main_end = .;                                                                                                                                                                                                                                                                                            

+        /* OS private BSS (protected) */                                                                                                                                                                                                                                                                                      
+       __uvisor_os_priv_start = .;                                                                                                                                                                                                                                                                                           
+        *(.bss.os_priv)                                                                                                                                                                                                                                                                                                       
+        . = ALIGN(32);                                                                                                                                                                                                                                                                                                        
+        __uvisor_os_priv_end = .;                                                                                                                                                                                                                                                                                             
+                                                                                                                                                                                                                                                                                                                              
        /* Secure boxes .bss (protected) */                                                                                                                                                                                                                                                                                   
        . = ALIGN(32);                                                                                                                                                                                                                                                                                                        
        __uvisor_bss_boxes_start = .;                                                                                                                                                                                                                                                                                         
        KEEP(*(.keep.uvisor.bss.boxes))                                                                                                                                                                                                                                                                                       
        . = ALIGN(32);                                                                                                                                                                                                                                                                                                        
        __uvisor_bss_boxes_end = .;                                                                                                                                                                                                                                                                                           

        . = ALIGN(32);                                                                                                                                                                                                                                                                                                        
        __uvisor_bss_end = .;                                                                                                                                                                                                                                                                                                 
    } > RAM 
```

### DIFF
--- a/api/src/uvisor-input.S
+++ b/api/src/uvisor-input.S
@@ -52,6 +52,10 @@ uvisor_config:
     .long __uvisor_bss_main_start
     .long __uvisor_bss_main_end
 
+    /* Start and end address of OS private bss */
+    .long __uvisor_os_priv_start
+    .long __uvisor_os_priv_end
+
     /* Start and end address of uvisor secure boxes bss */
     .long __uvisor_bss_boxes_start
     .long __uvisor_bss_boxes_end
--- a/api/src/uvisor-input.S
+++ b/api/src/uvisor-input.S
@@ -38,42 +38,42 @@ uvisor_config:
     /* MBED_VERSION */
     .long 0
 
-    /* uvisor mode
+    /* uVisor mode
        0: disabled
        1: permissive
        2: enabled */
     .long __uvisor_mode
 
-    /* start and end address of protected bss */
+    /* Start and end address of protected bss */
     .long __uvisor_bss_start
     .long __uvisor_bss_end
 
-    /* start and end address of uvisor main bss */
+    /* Start and end address of uvisor main bss */
     .long __uvisor_bss_main_start
     .long __uvisor_bss_main_end
 
-    /* start and end address of uvisor secure boxes bss */
+    /* Start and end address of uvisor secure boxes bss */
     .long __uvisor_bss_boxes_start
     .long __uvisor_bss_boxes_end
 
-    /* start and end address of uvisor code and data */
+    /* Start and end address of uvisor code and data */
     .long __uvisor_main_start
     .long __uvisor_main_end
 
-    /* start and end address of protected flash region
+    /* Start and end address of protected flash region
     /* note: remaining flash size is available for configuration storage */
     .long __uvisor_secure_start
     .long __uvisor_secure_end
 
-     /* start and end address of boxes configuration tables */
+     /* Start and end address of boxes configuration tables */
     .long __uvisor_cfgtbl_start
     .long __uvisor_cfgtbl_end
 
-    /* start and end address of list of pointers to boxes configuration tables */
+    /* Start and end address of list of pointers to boxes configuration tables */
     .long __uvisor_cfgtbl_ptr_start
     .long __uvisor_cfgtbl_ptr_end
 
-    /* pointer to __uvisor_box_context */
+    /* Pointer to __uvisor_box_context */
     .long __uvisor_box_context
 
     /* Physical memories' boundaries */
@@ -83,7 +83,7 @@ uvisor_config:
     .long __uvisor_sram_end;
 
 __uvisor_mode:
-    /* uvisor default mode - user can override weak reference */
+    /* uVisor default mode - user can override weak reference */
     .long 0
 
 .section .bss
--- a/core/system/src/mpu/vmpu.c
+++ b/core/system/src/mpu/vmpu.c
@@ -95,6 +95,11 @@ static int vmpu_sanity_checks(void)
     assert( (uint32_t) __uvisor_config.bss_main_end == (SRAM_OFFSET_START +
                                                         UVISOR_SRAM_LENGTH) );
 
+    /* Verify OS private BSS */
+    assert(__uvisor_config.os_priv_end >= __uvisor_config.os_priv_start);
+    assert(__uvisor_config.os_priv_start == __uvisor_config.bss_main_end);
+    assert(__uvisor_config.os_priv_end > __uvisor_config.bss_main_start);
+
     /* verify that secure flash area is accessible and after public code */
     assert(!vmpu_public_flash_addr((uint32_t) __uvisor_config.secure_start));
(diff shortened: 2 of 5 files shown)


core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -703,7 +703,7 @@ void vmpu_arch_init_hw(void)
     vmpu_acl_static_region(
         2,
         (void *) SRAM_ORIGIN,
-        ((uint32_t) __uvisor_config.bss_end) - SRAM_ORIGIN,
+        ((uint32_t) __uvisor_config.os_priv_end) - SRAM_ORIGIN,
```
AlessandroA: Could you add a comment here that says that we are assuming that priv memory and uVisor bss are contiguous?

core/system/src/mpu/vmpu_freescale_k64_mem.c
```diff
@@ -254,7 +254,7 @@ void vmpu_mem_init(void)
     int res;
 
     /* uvisor SRAM only accessible to uvisor */
-    res = vmpu_mem_add_int(0, (void*)SRAM_ORIGIN, ((uint32_t)__uvisor_config.bss_main_end)-SRAM_ORIGIN,
+    res = vmpu_mem_add_int(0, (void*)SRAM_ORIGIN, ((uint32_t)__uvisor_config.os_priv_end)-SRAM_ORIGIN,
```
AlessandroA: Same comment as above (add assumptions).

core/system/src/mpu/vmpu.c
```diff
@@ -95,6 +95,10 @@ static int vmpu_sanity_checks(void)
     assert( (uint32_t) __uvisor_config.bss_main_end == (SRAM_OFFSET_START +
                                                         UVISOR_SRAM_LENGTH) );
 
+    /* Verify OS private BSS */
+    assert(__uvisor_config.os_priv_start == __uvisor_config.bss_main_end);
+    assert(__uvisor_config.os_priv_end > __uvisor_config.bss_main_start);
```
AlessandroA: This should be `__uvisor_config.os_priv_end >= __uvisor_config.os_priv_start` right?

In thiw way the 2 assertions are independent and orthogonal:

* The first one checks the assumption on contiguous sections (uVisor bss followed by os priv mem).
* The second one checks that the os priv mem is a sane section (end >= start).

api/src/uvisor-input.S
```diff
@@ -52,6 +52,10 @@ uvisor_config:
     .long __uvisor_bss_main_start
     .long __uvisor_bss_main_end
 
+    /* start and end address of OS private bss */
```
AlessandroA: s/start/Start

AlessandroA:
Could you add a description to the commit? This is technically introducing a new important feature, because it allows an operating system to hold private data when running in privileged mode.

This will all be captured by a documentation extension that explains the privileged features exported by uVisor.

core/system/src/mpu/vmpu_armv7m.c
Patater: We aren't assuming. We have a sanity check to make sure.

Added a comment.

core/system/src/mpu/vmpu.c
Patater: My second assertion was checking that the OS private BSS ends after the start of uVisor main BSS, which should be strictly greater than, as we always reserve a non-zero UVISOR_SRAM_SIZE amount of data in the main BSS.

Your proposed second assertion is another good one to add, though.

Patater:
Updated to address Alessandro comments

core/system/src/mpu/vmpu.c
```diff
@@ -95,6 +95,11 @@ static int vmpu_sanity_checks(void)
     assert( (uint32_t) __uvisor_config.bss_main_end == (SRAM_OFFSET_START +
                                                         UVISOR_SRAM_LENGTH) );
 
+    /* Verify OS private BSS */
+    assert(__uvisor_config.os_priv_end >= __uvisor_config.os_priv_start);
+    assert(__uvisor_config.os_priv_start == __uvisor_config.bss_main_end);
+    assert(__uvisor_config.os_priv_end > __uvisor_config.bss_main_start);
```
AlessandroA: I'm still not sure this is needed. This section of the sanity checks only check the os private memory section right? In that case I would expect that they check that (i) it is a legit section and (ii) it comes after the end of bss_main.

We give for granted that bss_main is legit as well (it is checked by other checks anyway), so if os_priv_start is == bss_main_end and os_priv_end is >= os_priv_start, then I can trust os_priv_end is > bss_main_start (because I trust bss_main_start < bss_main_end, from other checks)

Patater: Yes, I agree. I'll remove the check.

api/src/uvisor-input.S
```diff
@@ -38,42 +38,46 @@ uvisor_config:
...
-    /* start and end address of uvisor main bss */
+    /* Start and end address of uvisor main bss */
     .long __uvisor_bss_main_start
     .long __uvisor_bss_main_end
 
-    /* start and end address of uvisor secure boxes bss */
+    /* Start and end address of OS private bss */
+    .long __uvisor_os_priv_start
+    .long __uvisor_os_priv_end
+
```
meriac: @Patater : do we still need these symbols exported to uVisor?

core/system/src/mpu/vmpu_armv7m.c
meriac: @Patater - as discussed bss_end needs to stay bss_end (os_priv_end not correct).

api/src/uvisor-input.S
Patater: No. I'm reworking the implementation not to need them anymore.

mazimkhan:
retest

Patater:
Closing PR. We'll take a different approach to solving this problem.

PR #216, opened 2016-05-16
Fix IAR again

AlessandroA:
Fixes:

* Attributes not re-defined for IAR.
* Wrong `volatile` cast previously introduced (see commit description).

@meriac @Patater @niklas-arm  @c1728p9


AlessandroA:
@c1728p9 Please get back to me if the build is still failing!

c1728p9:
Hi @AlessandroA, thanks for the quick update.  Could you push this to uvisor-mbed-lib as well?

PR #218, opened 2016-05-22
Context switches unification

AlessandroA:
This PR introduces a more consistent/unified model for **state-keeping** and **context switching** in uVisor.

The main new files to review are `context.{c,h}`. They provide APIs to switch context and keep state internally. Only the current stack pointers and the currently active box are exposed to the rest of the code.

Now context switches are distinguished between _function-bound_ (secure gateways, IRQs) and _unbound_ (thread context switching). The main difference is that _unbound_ context switches do not tamper with stacks.

**Notes**:

* Currently untested — Just up for general review.
* There is a commit that deletes the secure gateway after the refactoring. This is to allow to re-introduce it in case it was needed in the future.
* The debug box de-privileging hasn't been ported yet.
* @niklas-arm These new APIs should be put together to "build" the thread switching function. I think in the end it will look very similar to what you have now with the old format.

@meriac @niklas-arm @Patater


AlessandroA:
The new files have extensive comments and doxygen documentation. The reason is that we regard this code as highly critical and comments will help clarifying expectations and assumptions for each function (also useful for verification — @nchong)

core/debug/src/debug.c
```diff
@@ -58,6 +60,7 @@ void debug_init(void)
     SCnSCB->ACTLR |= 0x2;
 }
 
+#if 0
```
Patater: Please delete or leave this code in, depending on what you want to do.

core/system/inc/context.h
```diff
@@ -0,0 +1,179 @@
...
+uint8_t g_active_box;
+
+/** Array of current state for all boxes
+ *
+ * @warning The stack pointer field is not an accurate snapshot of the boxes'
+ * stack pointer at every time. It is only updated when a context switch occurs.
+ */
+TContextCurrentState g_context_current_states[UVISOR_MAX_BOXES];
+
+/** Returns a pointer to the previous box context state.
```
Patater: Use imperative mood.

core/system/src/context.c
```diff
@@ -0,0 +1,266 @@
...
+ * @warning This function trusts all the arguments that are passed to it. Input
+ * verification should be performed by the caller.
+ *
+ * @param context_type[in]  Type of context switch to perform
+ * @param src_id[in]        ID of the box we are switching context from
+ * @param src_sp[in]        Stack pointer of the box we are switching from */
+static void context_state_push(TContextSwitchType context_type, uint8_t src_id, uint32_t src_sp)
+{
+    /* Check that the state stack does not overflow. */
+    if (g_context_p == UVISOR_CONTEXT_MAX_DEPTH) {
```
Patater: What about underflow? `g_context_p` is currently an `int`.

What about overflow beyond `UVISOR_CONTEXT_MAX_DEPTH`?

core/system/src/context.c
```diff
@@ -0,0 +1,266 @@
...
+ * @internal
+ *
+ * @warning This function trusts all the arguments that are passed to it. Input
+ * verification should be performed by the caller.
+ *
+ * @returns the pointer to the previous box context state. */
+static TContextPreviousState * context_state_pop(void)
+{
+    /* Check that the state stack does not underflow. */
+    if (!g_context_p) {
```
Patater: What about underflow below `0`? `g_context_p` is currently an `int`.

core/system/src/context.c
```diff
@@ -0,0 +1,266 @@
...
+/* Forges a new exception stack frame and copy arguments from an old one. */
+uint32_t context_forge_exc_sf(uint32_t src_sp, uint8_t dst_id, uint32_t dst_fn, uint32_t dst_lr, uint32_t xpsr, int nargs)
+{
+    uint32_t dst_sp;
+    uint32_t exc_sf_alignment;
+
+    /* Destination box: Gather information from the current state. */
+    /* Note: Here we allow source and destination box IDs to be the same, as
+     * this function is used also for IRQs. */
+    if (dst_id != dst_id) {
```
Patater: `dst_id` is always `dst_id`. This branch will never happen.

core/system/src/context.c
```diff
@@ -0,0 +1,266 @@
...
+    ((uint32_t *) dst_sp)[5] = dst_lr;
+
+    /* Return value. This is always OR-ed with 1 (Thumb bit). */
+    ((uint32_t *) dst_sp)[6] = dst_fn | 1;
+
+    /* Stacked xPSR register. This always includes information on the stack
+     * frame alignment, regardless of the input from the caller. */
+    ((uint32_t *) dst_sp)[7] = xpsr | (exc_sf_alignment << 9);
+
+    /* Return pointer to the newly created stack frame. */
```
Patater: This says "return pointer", and I see lots of casting going on all over the place, so maybe it would be more natural (less casting involved) if we store and consume stack pointers as pointers instead of `uint32_t`.

core/system/src/context.c
```diff
@@ -0,0 +1,266 @@
...
+    exc_sf_alignment = (((uint32_t *) dst_sp)[7] & 0x4) ? 1 : 0;
+    g_context_current_states[dst_id].sp = dst_sp + CONTEXT_SWITCH_EXC_SF_SIZE + exc_sf_alignment;
+}
+
+/** Validates an exception stack frame, checking that the currently active box
+ *  has access to it.
+ *
+ * @internal
+ *
+ * The exception stack frame is validated through direct access: If the stack
```
Patater: Replace `:` with `.`.

core/system/src/context.c
```diff
@@ -0,0 +1,266 @@
...
+     * destination box.
+     * This is only needed if the context switch is tied to a function. Unbound
+     * context switches require the host OS to set the correct stack pointer
+     * before handling execution to the unprivileged code, and for the same
+     * reason do not require state-keeping.  */
+    /* This function halts if it finds an error. */
+    if (context_type == CONTEXT_SWITCH_FUNCTION_GATEWAY ||
+        context_type == CONTEXT_SWITCH_FUNCTION_ISR     ||
+        context_type == CONTEXT_SWITCH_FUNCTION_DEBUG) {
+        context_state_push(context_type, src_id, src_sp);
```
Patater: Just an aside on our style: I would find these sorts of functions more readable with the `{` on a new line below the `if` predicates.

core/system/src/mpu/vmpu.c
```diff
@@ -423,21 +423,20 @@ int vmpu_box_id_self(void)
 
 int vmpu_box_id_caller(void)
```
Patater: Just a note: I expect this function to be deleted when we remove secure gateways.

core/system/src/secure_gateway.c
```diff
@@ -0,0 +1,250 @@
...
+ * register doesn't get polluted and to provide context privacy during a context
+ * switch. The actual function is ::secure_gateway_context_switch_out. */
+void UVISOR_NAKED secure_gateway_out(uint32_t svc_sp)
+{
+    /* According to the ARM ABI, r0 will have the following value when this
+     * function is called:
+     *   r0 = svc_sp */
+    asm volatile(
+        "push {lr}\n"                                   /* Preserve the lr register. */
+        "bl   secure_gateway_context_switch_out\n"      /* secure_gateway_context_switch_out(svc_sp) */
```
Patater: Does `secure_gateway_context_switch_out` make any assumptions about the stack frame? Is it aware of the extra `lr` on the stack? Does it require the extra `lr` on the stack in order to make its assumptions about the stack frame correct?

core/system/src/secure_gateway.c
```diff
@@ -0,0 +1,250 @@
...
+    dst_sp = context_validate_exc_sf(svc_sp);
+    context_discard_exc_sf(g_active_box, dst_sp);
+
+    /* Perform the context switch back to the previous state.
+     * Upon return, the previous state structure will be populated. */
+    previous_state = context_switch_out(CONTEXT_SWITCH_FUNCTION_GATEWAY);
+
+    /* Copy the return value from the destination stack frame.
+     * Note: We trust the source box stack pointer as stored in the previous
+     *       state.*/
```
Patater: I hope the stack pointer as stored in the previous state includes our saved `lr` and isn't just the `svc_sp`.

core/system/src/unvic.c
```diff
@@ -280,163 +282,169 @@ int unvic_irq_level_get(void)
...
+        "mov  r4,  #0\n"                                /*     Clear r4  */
+        "mov  r5,  #0\n"                                /*     Clear r5  */
+        "mov  r6,  #0\n"                                /*     Clear r6  */
+        "mov  r7,  #0\n"                                /*     Clear r7  */
+        "mov  r8,  #0\n"                                /*     Clear r8  */
+        "mov  r9,  #0\n"                                /*     Clear r9  */
+        "mov  r10, #0\n"                                /*     Clear r10 */
+        "mov  r11, #0\n"                                /*     Clear r11 */
+        "unvic_gateway_no_regs_clearing:\n"             /* } else { ; }  */
+        "pop  {lr}\n"                                   /* Restore the lr register. */
```
Patater: All this other `lr` saving stuff is also tricky and needs a careful look over.

core/debug/src/debug.c
AlessandroA: Removed in a new commit.

core/system/src/context.c
AlessandroA: Turned into `>='.

core/system/src/context.c
AlessandroA: Turned into a `uint32_t`.

core/system/src/context.c
AlessandroA: Thanks! My typo. Now fixed.

core/system/src/context.c
AlessandroA: Changed the comment to make clearer. @meriac and I discussed this some time ago and concluded that for security purposes is better to have pointer stored and treated as `uint32_t`, which prevents us from using the as a pointers (de-reference it) by mistake.

In general, we are not supposed to cast-and-deference these pointers directly, but instead should use the unprivileged-access functions `vmpu_unpriv_{read,write}` which result in `*t` instructions. In this case this is not possible as we are accessing elements of a box different from the currently active one.

core/system/src/context.c
AlessandroA: I agree, the `if` content is aligned with the `if` conditions and makes it confusing, but for the moment I'd stick with this code style since it's the one we decided to adopt. If we introduce exceptions it will be difficult for (future) tools to check our style.

core/debug/src/debug.c
```diff
@@ -60,83 +60,6 @@ void debug_init(void)
     SCnSCB->ACTLR |= 0x2;
 }
 
-static void debug_cx_state(int _indent)
```
Patater: I'd recommend we fix these to work with the new APIs instead of deleting. The refactoring would go into the refactoring commit, just like the Box ID changes did.

core/system/src/context.c
Patater: Yup, I'm not recommending any changes to the code right now. I was just pointing out an instance of when our style is suboptimal on clarity. (In this case, it's hard to tell where the predicates end and the body starts.)

core/system/src/unvic.c
AlessandroA: Here the approach is that the wrapper calls a dedicated handler (the actual switch-in/out functions) and then returns. If you are call-and-return-ing you always need to push the `lr` and then pop it again as the `pc`:

```
push {r4 - r11, lr}
... /* do some stuff */
pop {r4 - r11, pc}
```

We changed it slightly here (two consecutive `push` ops) because we will `pop` them separately: The first `pop` retrieves the `lr`, which is needed for exception return (in this case it's a `EXC_RETURN` `lr`). The second `pop` will be done at context-switch-out and will retrieve the callee-saved registers from the old stack.

The actual context switch functions have no expectations on their own stack frames, that's why we de-coupled the wrapper and the actual handler. In this way, the switch functions just follow the ABI, the wrappers do the dark magic.

core/system/src/secure_gateway.c
AlessandroA: See [this comment](https://github.com/ARMmbed/uvisor/pull/218#discussion_r64188546).

core/system/src/secure_gateway.c
AlessandroA: See [this comment](https://github.com/ARMmbed/uvisor/pull/218#discussion_r64188546).

AlessandroA:
* Addressed comments by @Patater. Thanks for reviewing!

core/debug/src/debug.c
AlessandroA: The reason I didn't do it is that this requites slightly more change than the rest of the code, because it uses some internal states that are not exposed. The function is very old anyway and it was not showing all meaningful information (I was planning to change it anyway).

If it's ok I'd still remove it and add a ticket to re-implement context-switch debugging. Currently debug for context switches was disabled anyway.

Patater: Since it is too much work to do now, tracking in a ticket sounds good to me.

I'd delete the code before the refactor commit, to keep everything building without any awkward commenting out in any commits.

AlessandroA: Good point! Done.

AlessandroA:
* Tested on `ARMmbed/uvisor-helloworld` (with yotta).

AlessandroA:
Closing in favor of #220, which is targeted at the **dev** branch.

PR #219, opened 2016-05-23
Disable all IRQs: Added virtualised APIs

AlessandroA:
The new APIs allow a user to disable all IRQs for the currently active
box. Interrupts belonging to other boxes are never affected by this
call.

Box 0 interrupts are treated as other boxes' IRQs.

@meriac @Patater @niklas-arm


AlessandroA:
**Notes**:

* Currently under test, not all cases tested.
* Please review first the API names and programmer's model.

@meriac @Patater @niklas-arm

core/system/inc/unvic.h
```diff
@@ -53,4 +54,18 @@ extern int      unvic_default(uint32_t isr_id);
 
 extern void unvic_init(void);
 
+/** Disables all interrupts for the currently active box.
```
Patater: Use imperative mood.

core/system/src/unvic.c
```diff
@@ -179,6 +179,57 @@ void unvic_irq_disable(uint32_t irqn)
...
+    for (i = 0; i < NVIC_VECTORS; i++) {
+        if (g_unvic_vector[i].id == g_active_box) {
+            /* Re-enable the previously disabled IRQ. */
+            if (g_unvic_vector[i].was_enabled) {
+                NVIC_EnableIRQ(i);
+
+                /* Reset the state. This is only needed in case someone calls
+                 * this function without having previously called the
+                 * disable-all one. */
+                g_unvic_vector[i].was_enabled = false;
```
Patater: Would anything bad happen if the most recently enabled interrupt triggered before we had a chance to set `was_enabled`? Imagine if the interrupt handler called `vIRQ_DisableAll`.

core/system/src/unvic.c
```diff
@@ -179,6 +179,57 @@ void unvic_irq_disable(uint32_t irqn)
...
+
+    /* Iterate over all the IRQs owned by the currently active box. */
+    for (i = 0; i < NVIC_VECTORS; i++) {
+        if (g_unvic_vector[i].id == g_active_box) {
+            /* Remember the state for this IRQ. The state is the NVIC one, so we
+             * are sure we don't enable spurious interrupts. */
+            g_unvic_vector[i].was_enabled = (bool) UNVIC_IS_IRQ_ENABLED(i);
+
+            /* Disable the IRQ. */
+            NVIC_DisableIRQ(i);
```
Patater: Would anything bad happen if the interrupt we set to `was_enabled` triggered before we had a chance to call `NVIC_DisableIRQ`?

core/system/src/unvic.c
AlessandroA: I don't think so.

If for example you call `vIRQ_DisableAll` while in an IRQ, that very IRQ will also be disabled (which means it will not fire again), but the current ISR still runs fine. The uVisor does not check that a context switch out for an IRQ comes from an actually enabled IRQ (the ISR could change the NVIC state in other ways, like with `vIRQ_DisableIRQ`.

The `was_enabled` flag is never used to change the NVIC state in other functions, just here in the `*all` ones.

That's a test case I will add anyway!

core/system/src/unvic.c
AlessandroA: This corner case might occur in the following cases:

* When the IRQ triggers while we are running `vIRQ_DisableAll()`.
* When we are running in an ISR for an IRQ and, even before calling `vIRQ_DisableAll()`, another IRQ with lower priority is triggered.

In those case the IRQ becomes pending. With the `vIRQ_DisableAll()` function we would disable that IRQ as well. From the ARMv7-M ARM:

> When an interrupt is disabled, interrupt assertion causes the interrupt to become pending, but the interrupt cannot become active. If an interrupt is active when it is disabled, it remains in the active state until this is cleared by a reset or an exception return. Clearing the enable bit prevents any new activation of the associated interrupt.

So in both cases, disabling a pending IRQ will leave it pending. It will probably be served as soon as someone calls `vIRQ_EnableAll()`.

AlessandroA:
WRT [this comment](https://github.com/ARMmbed/uvisor/pull/219#discussion_r64191570) it is probably worth noting that the new APIs do not preclude a user from enabling a single IRQ between a `vIRQ_DisableAll()` and the consecutive `vIRQ_EnableAll()` call.

For example:

```C
...
vIRQ_DisableAll();

/* Do some stuff. */

vIRQ_EnableIRQ(IRQ_42); /* This will be enabled. */

/* Do some other stuff. */

vIRQ_EnableAll();
...
```

The example above will not fail, but the user will have mistakenly enabled an individual IRQ. **If we agree that this is still OK, I will add a comment on this to the API doc**.

The classic `__disable_irq()` intrinsic will not allow this, as it does not change the NVIC state, it literally disables the whole NVIC module, like in this other example:

```C
...
__disable_irq();

/* Do some stuff. */

vIRQ_EnableIRQ(IRQ_42); /* This will not be enabled! */

/* Do some other stuff. */

__enable_irq();
...
```

@meriac @patater @niklas-arm

AlessandroA:
Closing in favor of #221, which targets the **dev** branch.

Issue #91, opened 2015-10-28
Add register clearing between context switches


meriac:
- for security reasons CPU core registers need to be cleared whenever execution shifts to a new box to avoid leakage of information
- a box can opt-out from register clearing for performance reasons in case privacy is not required

rainierwolfcastle:
ARM Internal Ref: IOTSFW-1120

PR #220, opened 2016-05-23
Context switches unification

AlessandroA:
**This PR replaces #218. Please refer to it for previous comments.**

This PR introduces a more consistent/unified model for **state-keeping** and **context switching** in uVisor.

The main new files to review are `context.{c,h}`. They provide APIs to switch context and keep state internally. Only the current stack pointers and the currently active box are exposed to the rest of the code.

Now context switches are distinguished between _function-bound_ (secure gateways, IRQs) and _unbound_ (thread context switching). The main difference is that _unbound_ context switches do not tamper with stacks.

**Notes**:

* ~~Currently untested — Just up for general review.~~
* The secure gateway has also been ported, but it will be removed by a later commit.
* The debug box de-privileging hasn't been ported yet.
* @niklas-arm These new APIs should be put together to "build" the thread switching function. I think in the end it will look very similar to what you have now with the old format.

@meriac @niklas-arm @Patater


AlessandroA:
* Tested on `ARMmbed/uvisor-helloworld` (with yotta).
* All tests passing (`uvisor-lib` tests).

AlessandroA:
Note: From the PR it looks like https://github.com/ARMmbed/uvisor/pull/220/commits/b9770b2faad096d3930319f0934c93602da699de comes after https://github.com/ARMmbed/uvisor/pull/220/commits/37c1c0ceb0557350c88f7c043e471a2b23847de4 but the real history shows the opposite, following @Patater's [comment](https://github.com/ARMmbed/uvisor/pull/218#discussion_r64189785).

AlessandroA:
* Update: The context switch-out function now throws an error if it's called to perform an unbound context switch (like thread context switching on RTOS). This is because in thread switching you only switch from A to B, there is no going back. If from B you want to go back to A it's just another context switch.

core/system/inc/context.h
```diff
@@ -0,0 +1,179 @@
...
+
+/** Snapshot of a previous context state
+ *
+ * This struct contains the information needed to restore a previously stalled
+ * context. It is needed when returning from a nested context switch.
+ *
+ * @warning We assume that the context type (a value from the
+ * ::TContextSwitchType enum) fits into 8 bits. */
+typedef struct {
+    uint8_t __pad[2]; /**< Padding to get 32 bit alignment */
```
meriac: @AlessandroA  why is the padding at the start, not after the other two 8bit variables ?

core/system/inc/secure_gateway.h
```diff
@@ -0,0 +1,85 @@
+/*
```
meriac: @AlessandroA do we still need the Secure gateway definition? Can't we use that into a dead branch till further notice ?

core/system/src/context.c
```diff
@@ -0,0 +1,283 @@
...
+    uint32_t exc_sf_alignment;
+
+    /* Source box: Gather information from the current state. */
+    src_id = g_active_box;
+
+    /* Destination box: Gather information from the current state. */
+    /* Note: Here we allow source and destination box IDs to be the same, as
+     * this function is used also for IRQs. */
+    if (src_id != dst_id) {
+        dst_sp = g_context_current_states[dst_id].sp;
```
meriac: @AlessandroA Consider to swap the if-branches for better readability (and turn the != to ==)

core/system/src/unvic.c
```diff
@@ -280,163 +282,169 @@ int unvic_irq_level_get(void)
...
-        "mov  r8,  #0\n"
-        "mov  r9,  #0\n"
-        "mov  r10, #0\n"
-        "mov  r11, #0\n"
-        "no_regs_clearing:\n"
-        "pop  {lr}\n"               /* restore lr */
-        "orr  lr, #0x1C\n"          /* return with PSP, 8 words */
-        "bx   lr\n"
-        /* the unprivileged interrupt handler will be executed after this */
+        "push {r4 - r11}\n"                             /* Store the callee-saved registers on the MSP (privileged). */
```
meriac: @AlessandroA for privacy only the required registers should be copied, and the remaining registers saved/zeroed if privacy is enabled.

api/inc/context_exports.h
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef __UVISOR_CONTEX_EXPORTS_H__
+#define __UVISOR_CONTEX_EXPORTS_H__
+
+/** Maximum number of nested context switches. The same state stack is kept for
+ * all kinds of context switches, which includes secure gateways, interrupts,
+ * and task switching. */
+#define UVISOR_CONTEXT_MAX_DEPTH 16
```
Patater: Why did this need to move into an `_exports.h` header?

meriac:
@AlessandroA @Patater LGTM - but we need to remove the secure gateway code in a subsequent request in favour of the upcoming RPC.

Removing it *before* we have the RPC is appropriate as the CGW creates the wrong impression that it functions within a threaded environment.

core/system/inc/context.h
```diff
@@ -0,0 +1,179 @@
...
+    uint8_t type;     /**< Context switch type */
+    uint8_t src_id;   /**< ID of the box the context belongs to */
+    uint32_t src_sp;  /**< Stack pointer to restore for the context */
+} TContextPreviousState;
```
Patater: Consider making this `UVISOR_PACKED`, if padding and alignment are important.

Please document in code comments why the exact padding and alignment are important.

core/system/src/unvic.c
```diff
@@ -280,163 +282,169 @@ int unvic_irq_level_get(void)
...
-    /* de-privilege executionn */
-    __set_PSP((uint32_t) dst_sp);
+    /* De-privilege execution. */
     __set_CONTROL(__get_CONTROL() | 3);
 
-    /* DPRINTF("IRQ %d served with vector 0x%08X\n\r", irqn, */
-    /*                                                 g_unvic_vector[irqn]); */
-    /* FIXME add support for privacy (triggers register clearing) */
+    /* Return whether the destination box requires privacy or not. */
+    /* TODO: Context privacy is currently unsupported. */
```
Patater: Should this remain as a FIXME instead of being turned into a TODO? I'm not sure if we differentiate between FIXME and TODO for the sake of tracking issues in code.

Patater:
> Style: Make header file inclusion consistent
Please comment what exactly is being done in the commit message. Be explicit about the style you are enforcing in the commit message.

core/system/inc/context.h
AlessandroA: Good point! I changed it. I though it didn't make much difference, but actually a single `ldrb` instruction to the struct location fetches the data directly (no need of a second load). That's also why I have now put `src_id` first.

core/system/inc/secure_gateway.h
AlessandroA: 👍 As soon as this PR is merged I'll create a new one that removes the secure gateway. This is to create a single point of removal that we can point to in the future if we want to.

core/system/src/context.c
AlessandroA: Done.

api/inc/context_exports.h
AlessandroA: Not strictly needed by user code with our current use-cases, but we use it in tests.

I can imagine though that it could be use to ensure that a recursive function doesn't go too deep before hitting a `HALT_ERROR`.

core/system/inc/context.h
AlessandroA: Thanks! I added the `UVISOR_PACKED` attribute, which is useful to fulfill the byte-load access mentioned [here](https://github.com/ARMmbed/uvisor/pull/220#discussion_r64707877). I also added a comment for clarity.

core/system/src/unvic.c
AlessandroA: I opted for `TODO` as this is a feature that we don't document. People do not expect it from the docs, hence it not being there is not technically a `FIXME`.

AlessandroA:
* Addresses @meriac's comments.
* Addressed new @Patater's comments.

api/inc/context_exports.h
Patater: Ah, ok. I can see it being useful for tests.

core/system/src/unvic.c
Patater: Probably that this feature is missing should be documented, as it is security critical. I'd guess we already have an issue for it.

Patater:
LTGM

Thanks new @AlessandroA

core/system/src/unvic.c
AlessandroA: See https://github.com/ARMmbed/uvisor/issues/91.

AlessandroA:
* Rebased on `dev`.

PR #221, opened 2016-05-23
Disable all IRQs: Added virtualised APIs

AlessandroA:
**This PR replaces #219. Please refer to it for previous comments.**

The new APIs allow a user to disable all IRQs for the currently active
box. Interrupts belonging to other boxes are never affected by this
call.

Box 0 interrupts are treated as other boxes' IRQs.

@meriac @Patater @niklas-arm


AlessandroA:
**Notes**:

* Currently under test, not all cases tested.
* Please review first the API names and programmer's model.

@meriac @Patater @niklas-arm

AlessandroA:
WRT [this comment](https://github.com/ARMmbed/uvisor/pull/219#discussion_r64191570) it is probably worth noting that the new APIs do not preclude a user from enabling a single IRQ between a `vIRQ_DisableAll()` and the consecutive `vIRQ_EnableAll()` call.

For example:

```C
...
vIRQ_DisableAll();

/* Do some stuff. */

vIRQ_EnableIRQ(IRQ_42); /* This will be enabled. */

/* Do some other stuff. */

vIRQ_EnableAll();
...
```

The example above will not fail, but the user will have mistakenly enabled an individual IRQ. **If we agree that this is still OK, I will add a comment on this to the API doc**.

The classic `__disable_irq()` intrinsic will not allow this, as it does not change the NVIC state, it literally disables the whole NVIC module, like in this other example:

```C
...
__disable_irq();

/* Do some stuff. */

vIRQ_EnableIRQ(IRQ_42); /* This will not be enabled! */

/* Do some other stuff. */

__enable_irq();
...
```

@meriac @patater @niklas-arm

Patater:
This new API isn't going to be backwards compatible if one can re-enable individual interrupts. That makes it a deal breaker.

AlessandroA:
@meriac suggested a change that I think makes it backwards-compatible and still provides the expected feature. If you enable a single IRQ while in a disable-all block, the `was_enabled` flag gets set. When the re-enable-all function gets executed we re-enable all the original IRQ *plus* those that were added in the meantime.

An additional feature that Milosch suggested, which I'm adding now, is to add a counter that increments at every disable-all and decrements at every re-enable-all. In this way:

* Disable/Re-enable-all functions must always go in pair.
* A pair in a nested function does not invalidate the callee pair.

Example:

```C
vIRQ_DisableAll();           /* counter++ */

some_function_that_disables_and_re_enables_irqs(); /* counter++, then counter-- */

vIRQ_EnableAll();            /* counter--; now counter is back to 0 */
```

Patater:
Adding the counter and forcing disables and re-enables to happen in pairs will also break backwards compatibility.

Turning `was_enabled` into a counter might be interesting (at least for debugging purposes), though, when we have funny interrupts going on during the interrupt enabling or disabling SVCs. This would help make sure we are balanced in our use of `was_enabled`, not accidentally enabling or disabling something.

AlessandroA:
We can't completely avoid breaking backwards-compatibility because this is anyway a new API. It is a virtualized disable-all function, which by definition generates different behaviors from what users expect from `__disable_irq()`. I think it's just important that the programmer's flow isn't affected that much.

In the `__disable_irq` case enabling a single IRQ *does* enable it immediately, but prevents it from being served (the NVIC module is turned off). Milosch's proposal would look similar, but we have to fake it (the `NVIC_EnableIRQ` for a new IRQ is only called at the time of `vIRQ_EnableAll()`, because otherwise it would be served).

Patater:
We can do as good of a job as we can in making the behavior match, especially if the default implementation of the virtual functions falls back to `__disable_irq` and the like. Adding 

I recommend we follow Milosch's first suggestion, but not force the enables and disables to come in pairs unless we also force that in the default implementation of the virtual functions.

AlessandroA:
Well we do not *enforce* it but if you never re-enable IRQs after you disable them you are out of luck (but that's expected, I guess?).

BTW, I pushed a new version with these suggestions. I'm writing tests for it.

Patater:
I was thinking of code where multiple disables had been called, all undone with a single enable call. We'd break such code if we enforced that each disable needed a corresponding enable (reference counting).

AlessandroA:
**Don't merge yet**. Remaining:

* Update API comments with info on counters and nested disable-all calls.
* (waiting for tests outcome).

AlessandroA:
@meriac what do you think about Jaeden [comment](https://github.com/ARMmbed/uvisor/pull/221#issuecomment-221003940)?

> I was thinking of code where multiple disables had been called, all undone with a single enable call. We'd break such code if we enforced that each disable needed a corresponding enable (reference counting).

meriac:
@Patater @AlessandroA I don't believe its a common pattern to have unpaired Enable/Disable IRQ calls - i've never seen it. I had another thought though - we would need to suppress thread switching within a process group to make Enable/Disable IRQ work as expected.

The fact that threading still goes on after Enable/Disable IRQ defeats a common use: half-baked thread synchronization.

AlessandroA:
* Added better doxygen comments to explain how the API works.

@meriac Are these APIs missing any feature? If we assume ftm that RTOS uses SysTick only, scheduling is not affected by them.

PR #225, opened 2016-05-26
vIRQ_{Dis,En}ableAll: Fix bug of never-enabled IRQs

AlessandroA:
When nesting multiple pairs of disable-all/enable-all function calls,
the last call to `vIRQ_EnableAll()` was not re-enabling all interrupts.

This was due to the fact that if multiple `vIRQ_DisableAll()` were called,
the wrong state `was_enabled = false` was set, even if the IRQ was
previously enabled.

Upon restoring the state, those IRQs were not re-enabled.

IRQs that were enabled during the critical section were correctly
enabled, because their state was unaffected.

This also improves performance because in the previous implementation
IRQs were always iterated over even if the counter was more than 0,
which is useless.

Fixes:

* bcf5fbe "Disable all IRQs: Added virtualised APIs".

This bug was captured by the new unit tests.

@meriac @Patater @niklas-arm

### DIFF
--- a/core/system/src/unvic.c
+++ b/core/system/src/unvic.c
@@ -209,26 +209,31 @@ void unvic_irq_disable_all(void)
 {
     int irqn;
 
-    /* Iterate over all the IRQs owned by the currently active box and disable
-     * them if they were active before the function call. */
-    for (irqn = 0; irqn < NVIC_VECTORS; irqn++) {
-        if (g_unvic_vector[irqn].id == g_active_box && UNVIC_IS_IRQ_ENABLED(irqn)) {
-            /* Note: We only disable the IRQs if the counter is 0. */
-            if (!g_irq_disable_all_counter[g_active_box]) {
+    /* Only disable all IRQs if this is the first time that this function is
+     * called. */
+    if (g_irq_disable_all_counter[g_active_box] == 0) {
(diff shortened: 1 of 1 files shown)


Patater:
Before we move this feature into master, we should should apply this commit as a fixup to the original in the dev branch.

Patater:
LGTM

core/system/src/unvic.c
```diff
@@ -209,27 +209,29 @@ void unvic_irq_disable_all(void)
...
-    for (irqn = 0; irqn < NVIC_VECTORS; irqn++) {
-        if (g_unvic_vector[irqn].id == g_active_box && UNVIC_IS_IRQ_ENABLED(irqn)) {
-            /* Note: We only disable the IRQs if the counter is 0. */
-            if (!g_irq_disable_all_counter[g_active_box]) {
+    /* Only iterate over all IRQs if this is the first time that this function
+     * is called. The counter is always increased. */
+    if (g_irq_disable_all_counter[g_active_box] > 0) {
+        g_irq_disable_all_counter[g_active_box]++;
+    } else {
+        g_irq_disable_all_counter[g_active_box]++;
```
meriac: @AlessandroA fixme.

AlessandroA: Done.

PR #226, opened 2016-05-27
Remove the secure gateway feature

AlessandroA:
With an RTOS secure gateways are not possible any more and will be
replaced by a different mechanism.

This commit can be re-used in the future to re-introduce the secure
gateway feature.

@meriac @Patater @niklas-arm


api/inc/register_gateway.h
```diff
@@ -20,6 +20,8 @@
 #include "api/inc/uvisor_exports.h"
 #include <stdint.h>
 
+#define UVISOR_REGISTER_GATEWAY_MAGIC 0xCDCDCDCD
```
meriac: @AlessandroA the magic should not contain identical bytes in sequence. Ideally both halves are the same illegal opcode to ensure that normal thumb code will never have this sequence.

AlessandroA:
@meriac Changed the register gateway magic into `0xF7F3A89E`. The opcode `0xF7FxAxxx` is guaranteed to trigger an undefined instruction on ARMv7-M. The `x` elements are ignored and hence have been set randomly.

meriac:
@AlessandroA : Can you please gate that definition by an ifdef for cortex_m || thumb2 - and throw an error if that code is compiled for another platform. Have you verifed that you use the correct endianness by compiling the magic and attempting to disassemble the magic ?

AlessandroA:
* Added the comments.
* Added the check on Thumb-2.

This is how it looks like:

```asm
000070a8 <app_start>:
    ...
    70b2:	df83      svc 131
    70b4:	e009      b.n 70ca <skip_args18>
    70b6:	f7f3a89e  ;<UNDEFINED> instruction: 0xf7f3a89e
    70ba:	12345678  eorsne r5, r4, #120, 12	                     ; other metadata
    70be:	2000027c  andcs r0, r0, ip, ror r2                       ; other metadata
    ...
```

@meriac

PR #227, opened 2016-05-27
1. Add tier-1 page allocator

niklas-arm:
This adds the page allocator algorithm, initialization code and SVC calls.

This is part of larger changes, but I'd prefer to review these parts separately when possible.

@Patater @AlessandroA @meriac


niklas-arm:
Note, that the page heap has to be added to the target linkerscript for the uvisor-header to link correctly.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,239 @@
...
+#   include "mpu/vmpu.h"
+#   include "halt.h"
+#   include "context.h"
+#else
+#   include "uvisor-lib/uvisor-lib.h"
+#   define DPRINTF(...) {}
+#   define g_active_box 0
+#   define vmpu_is_box_id_valid(...) 0
+#   define HALT_ERROR(id, ...) DPRINTF(__VA_ARGS__)
+static uint32_t vmpu_unpriv_uint32_read(uint32_t addr)
```
meriac: @niklas-arm please add comment on replacing these two functions with the real unpriv functions headerfile from uVisor.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,239 @@
...
+static uint8_t g_page_count_total;
+
+/* Must be equal to page size for MPU alignment on ARMv7-M */
+#define UVISOR_PAGE_ALIGNMENT (g_page_size)
+
+
+void page_allocator_init(void * const heap_start, void * const heap_end, uint32_t page_size)
+{
+    /* Make sure the UVISOR_PAGE_UNUSED is definitely NOT a valid box id! */
+    if (vmpu_is_box_id_valid(UVISOR_PAGE_UNUSED)) {
```
meriac: @niklas-arm coolness.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,240 @@
...
+#   include "halt.h"
+#   include "context.h"
+#else
+#   include "uvisor-lib/uvisor-lib.h"
+#   define DPRINTF(...) {}
+#   define g_active_box 0
+#   define vmpu_is_box_id_valid(...) 0
+#   define HALT_ERROR(id, ...) DPRINTF(__VA_ARGS__)
+/* FIXME: Use real functions from `vmpu_unpriv_access.h` for unsupported too! */
+static uint32_t vmpu_unpriv_uint32_read(uint32_t addr)
```
niklas-arm: @AlessandroA This would require making `vmpu_unpriv_access.h` available in the API header files.

AlessandroA: Done. See #228

core/system/inc/page_allocator.h
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+#ifndef __PAGE_ALLOCATOR_H__
+#define __PAGE_ALLOCATOR_H__
+
+#include "api/inc/page_allocator_exports.h"
+
+/* Initialize the page allocator.
+ *
+ * The most importent argument to this function is the page size. It determines
```
AlessandroA: s/importent/important

api/inc/halt_exports.h
```diff
@@ -17,9 +17,15 @@
...
-#define UVISOR_ERROR_BOX_NAMESPACE_ANONYMOUS (-4)
+#define UVISOR_ERROR_INVALID_BOX_ID             (-2)
+#define UVISOR_ERROR_BUFFER_TOO_SMALL           (-3)
+#define UVISOR_ERROR_BOX_NAMESPACE_ANONYMOUS    (-4)
+
+
+#define UVISOR_ERROR_CLASS_MASK     (0xFFFF0000)
+#define UVISOR_ERROR_MASK           (0x0000FFFF)
+
+#define UVISOR_ERROR_CLASS_PAGE     (1ul << 16)
```
AlessandroA: s/1ul/1UL/

api/inc/page_allocator_exports.h
```diff
@@ -0,0 +1,62 @@
...
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_TABLE    (UVISOR_ERROR_CLASS_PAGE + 101)
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_SIZE     (UVISOR_ERROR_CLASS_PAGE + 102)
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_ORIGIN   (UVISOR_ERROR_CLASS_PAGE + 103)
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_OWNER    (UVISOR_ERROR_CLASS_PAGE + 104)
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_COUNT    (UVISOR_ERROR_CLASS_PAGE + 105)
+
+
+/* Must be a power of 2 for MPU alignment in ARMv7-M */
+#ifndef UVISOR_PAGE_SIZE
+#   define UVISOR_PAGE_SIZE ((size_t)16 * 1024)
```
AlessandroA: We never indented pre-processor directives. I'd stick to the following for now:

```C
#ifdef
#define
#endif
```

AlessandroA: s/`(size_t)16`/`(size_t) 16`/

api/inc/page_allocator_exports.h
```diff
@@ -0,0 +1,62 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef __UVISOR_API_PAGE_ALLOCATOR_EXPORTS_H__
+#define __UVISOR_API_PAGE_ALLOCATOR_EXPORTS_H__
+
+#include <stdint.h>
+#include <stddef.h>
+#include "halt_exports.h"
```
AlessandroA: Please use the full path here (we use this for all API header files). Could you also move the inclusion before the `std` ones? It helps ensuring that `halt_exports.h` is self-contained.

api/inc/page_allocator_exports.h
```diff
@@ -0,0 +1,62 @@
...
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_COUNT    (UVISOR_ERROR_CLASS_PAGE + 105)
+
+
+/* Must be a power of 2 for MPU alignment in ARMv7-M */
+#ifndef UVISOR_PAGE_SIZE
+#   define UVISOR_PAGE_SIZE ((size_t)16 * 1024)
+#endif
+
+/* Returns the rounded up number of pages required to hold `size` */
+#define UVISOR_PAGES_FOR_SIZE(size) \
```
AlessandroA: This can go in 1 line.

api/inc/page_allocator_exports.h
```diff
@@ -0,0 +1,62 @@
...
+/* Creates a page table with `count` many entries */
+#define UVISOR_PAGE_TABLE(count) \
+    struct { \
+        uint32_t page_size; \
+        uint32_t page_count; \
+        void * page_origins[count]; \
+    }
+
+/* Creates a page table with enough pages to hold `size` */
+#define UVISOR_PAGE_TABLE_FOR_SIZE(size) \
```
AlessandroA: This can go in 1 line.

core/system/inc/page_allocator.h
```diff
@@ -0,0 +1,51 @@
...
+ * the alignment of the start and end address, where the alignment size is the
+ * page size for ARMv7-M.
+ * The start address is rounded upwards, whereas the end address is rounded
+ * downwards to the next aligned border.
+ * This means that not all of the memory between start and end address can be
+ * used by the heap address, due to adhering to alignment restrictions.
+ *
+ * @param heap_start[in] start address of the page heap
```
AlessandroA: Capitalize the `@param` comments (3 of them here).

core/system/inc/page_allocator.h
```diff
@@ -0,0 +1,51 @@
...
+ * @returns non-zero on failure with failure class `UVISOR_ERROR_CLASS_PAGE`. See `UVISOR_ERROR_PAGE_*`.
+ */
+int page_allocator_malloc(UvisorPageTable * const table);
+
+/* Free the pages associated with the table, only if it passes validation.
+ * @returns non-zero on failure with failure class `UVISOR_ERROR_CLASS_PAGE`. See `UVISOR_ERROR_PAGE_*`.
+ */
+int page_allocator_free(const UvisorPageTable * const table);
+
+#endif /*__PAGE_ALLOCATOR_H__*/
```
AlessandroA:
```diff
-#endif /*__PAGE_ALLOCATOR_H__*/
+#endif /* __PAGE_ALLOCATOR_H__ */
```

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,240 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* This file can be compiled externally to provide the page allocator algorithm
+   for devices NOT supported by uVisor. For this purpose this file is copied as
+   is into uvisor-mbed-lib and compiled by the target build system. */
+#if defined(UVISOR_PRESENT) && (UVISOR_PRESENT == 1)
```
AlessandroA: See [here](https://github.com/ARMmbed/uvisor/pull/227/commits/af2a299e9d6f822e1187b50f5333b78175cf848c#r65154640).

core/system/inc/mpu/vmpu.h
```diff
@@ -162,6 +162,30 @@ extern int vmpu_box_id_self(void);
...
 extern int vmpu_box_namespace_from_id(int box_id, char *box_name, size_t length);
 
+/** Determines if the passed size can be mapped to an exact region size
+ * depending on underlying MPU implementation. Note that the size must be an
+ * exact match to a MPU region size.
+ * With the ARMv7-M MPU that would be 32 <= 2^N <= 512M.
+ * With the K64F MPU that would be 32 <= 32*N <= 512M.
+ *
+ * @note `size` is limited to 512M, even though larger sizes are allowed by the
+ *       MPU, but don't make sense from a security perspective.
```
niklas-arm: because?

api/inc/page_allocator_exports.h
```diff
@@ -0,0 +1,60 @@
...
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_OWNER    (UVISOR_ERROR_CLASS_PAGE + 104)
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_COUNT    (UVISOR_ERROR_CLASS_PAGE + 105)
+
+
+/* Must be a power of 2 for MPU alignment in ARMv7-M */
+#ifndef UVISOR_PAGE_SIZE
+#define UVISOR_PAGE_SIZE ((uint32_t) 16 * 1024)
+#endif
+
+/* Returns the rounded up number of pages required to hold `size` */
```
AlessandroA: Please use the imperative mood and punctuation:

`/* Return the rounded up number of pages required to hold `size`. */`

(It applies to all occurrences in this PR).

AlessandroA: To clarify, punctuation only applies if the comment is a full English sentence.

```C
/* Maximum context depth */
#define MAX_DEPTH 10
```

```C
/* Check that the value is valid. */
bool is_value_valid(uint32_t value);
```

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,249 @@
...
+    }
+    const uint32_t mask = size - 1;
+    return (addr + mask) & ~mask;
+}
+#endif
+
+/* By default a maximum of 16 pages are allowed. This can only
+   be decided by the porting engineer for the current platform! */
+#ifndef UVISOR_PAGE_TABLE_MAX_COUNT
+#define UVISOR_PAGE_TABLE_MAX_COUNT ((size_t)16)
```
AlessandroA:
```diff
-(size_t)16
+(size_t) 16
```

core/system/inc/linker.h
```diff
@@ -65,6 +65,10 @@ typedef struct {
     /* address of __uvisor_box_context */
     uint32_t **uvisor_box_context;
 
+    /* heap for page allocator */
+    uint32_t * page_start, * page_end;
+    uint32_t page_size;
```
meriac: @niklas-arm Isn't that a pointer to page_size ?

niklas-arm: yes… yes, it is…
_cries in corner_

niklas-arm:
I added pointer range checks for all input pointers passed to the page allocator init function.
Also added `vmpu_is_region_size_valid` and `vmpu_round_up_region` functions to make the page size and page alignment specialized on the MPU implementation.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,272 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* This file can be compiled externally to provide the page allocator algorithm
+   for devices NOT supported by uVisor. For this purpose this file is copied as
```
AlessandroA: I would remove the mbed-lib-specific sentence, but the comment is important — if you want to elaborate more in place of the mbed-lib part please do!

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,272 @@
...
+   is into uvisor-mbed-lib and compiled by the target build system. */
+#if defined(UVISOR_PRESENT) && (UVISOR_PRESENT == 1)
+#include <uvisor.h>
+#include "page_allocator.h"
```
AlessandroA: How does it work for `page_allocator.h`? It must be copied as well? Other header files (`*_exports.h` ones) are usually available for free (or copied altogether), but this would need to be copied separately.

AlessandroA: Sorry this doesn't apply, my mistake.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,272 @@
...
+    if (!vmpu_is_region_size_valid(size)) {
+        return 0;
+    }
+    const uint32_t mask = size - 1;
+    return (addr + mask) & ~mask;
+}
+#endif
+
+/* By default a maximum of 16 pages are allowed. This can only
+   be decided by the porting engineer for the current platform! */
```
AlessandroA:
```diff
-   be decided by the porting engineer for the current platform! */
+ * be decided by the porting engineer for the current platform! */
```

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,272 @@
...
+static const void * g_page_heap_start;
+/* Points to the end of the page heap */
+static const void * g_page_heap_end;
+/* Contains the number of free pages */
+static uint8_t g_page_count_free;
+/* Contains the total number of available pages */
+static uint8_t g_page_count_total;
+
+
+void page_allocator_init(void * const heap_start, void * const heap_end, uint32_t * const page_size)
```
AlessandroA: I guess page_size is a `uint32_t const * const` (constant pointer to constant uint32)?

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+}
+static uint32_t vmpu_round_up_region(uint32_t addr, uint32_t size)
+{
+    if (!vmpu_is_region_size_valid(size)) {
+        return 0;
+    }
+    const uint32_t mask = size - 1;
+    return (addr + mask) & ~mask;
+}
+static uint32_t page_table_read(uint32_t addr)
```
AlessandroA: Can this be inlined?

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+{
+    return *((uint32_t *) addr);
+}
+static void page_table_write(uint32_t addr, uint32_t data)
```
AlessandroA: Can this be inlined?

core/system/src/page_allocator.c
Patater: It's static. The compiler will likely inline it automatically. Check the disassembly to see.

core/system/inc/mpu/vmpu.h
```diff
@@ -162,6 +162,30 @@ extern int vmpu_box_id_self(void);
 extern int vmpu_box_id_caller(void);
 extern int vmpu_box_namespace_from_id(int box_id, char *box_name, size_t length);
 
+/** Determines if the passed size can be mapped to an exact region size
```
Patater: Use imperative mood.

core/system/inc/mpu/vmpu.h
```diff
@@ -162,6 +162,30 @@ extern int vmpu_box_id_self(void);
...
+ * @retval 0 The region size is not valid
+ */
+extern int vmpu_is_region_size_valid(uint32_t size);
+
+/**
+ * @param addr Input address to be rounded up
+ * @param size Region size chosen for alignment
+ *
+ * @retval >0 The passed address rounded up to the next possible alignment
+ * of the passed region size.
```
Patater: Remove period

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -717,3 +717,42 @@ void vmpu_arch_init_hw(void)
     );
 #endif /* !defined(UVISOR_IN_STANDALONE_SRAM) */
 }
+
+int vmpu_is_region_size_valid(uint32_t size)
+{
+    /* Get the MSB of the size */
```
Patater: Add period

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -717,3 +717,42 @@ void vmpu_arch_init_hw(void)
...
+    const int bits = vmpu_bits(size) - 1;
+    if (bits < ARMv7M_MPU_ALIGNMENT_BITS || 29 < bits) {
+        /* 2^5 == 32, which is the minimum region size */
+        /* 2^29 == 512M, which is the maximum region size */
+        return 0;
+    }
+    /* Compute the size from MSB */
```
Patater: Add period

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -717,3 +717,42 @@ void vmpu_arch_init_hw(void)
...
+    const int bit_size = (1UL << bits);
+    /* There is no round up */
```
Patater: Add period

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -717,3 +717,42 @@ void vmpu_arch_init_hw(void)
...
+    if (!vmpu_is_region_size_valid(size)) {
+        /* Region size must be valid! */
+        return 0;
+    }
+    /* Size is 2*N, can be used directly for alignment */
+    /* Create the LSB mask: Example 0x80 -> 0x7F */
+    const uint32_t mask = size - 1;
+    /* Mask is 31 <= mask <= (1 << 29) - 1 */
+
+    /* Adding the mask can overflow */
```
Patater: Add period

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -717,3 +717,42 @@ void vmpu_arch_init_hw(void)
...
+    const uint32_t rounded_addr = addr + mask;
+    /* Check for overflow */
```
Patater: Add period

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -717,3 +717,42 @@ void vmpu_arch_init_hw(void)
...
+    if (rounded_addr < addr) {
+        /* This means the address was too large to align */
```
Patater: Add period

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -717,3 +717,42 @@ void vmpu_arch_init_hw(void)
...
+        return 0;
+    }
+    /* Mask the rounded address to get the aligned address */
```
Patater: Add period

core/system/src/mpu/vmpu_freescale_k64.c
```diff
@@ -296,3 +295,36 @@ void vmpu_arch_init(void)
...
+int vmpu_is_region_size_valid(uint32_t size)
+{
+    /* Align size to 32B */
+    const uint32_t masked_size = size & ~31;
+    if (masked_size < 32 || (1 << 29) < masked_size) {
+        /* 2^5 == 32, which is the minimum region size */
+        /* 2^29 == 512M, which is the maximum region size */
+        return 0;
+    }
+    /* There is no rounding, we only care about an exact match! */
```
Patater: s/`,`/`.`
s/w/W

core/system/src/mpu/vmpu_freescale_k64.c
```diff
@@ -296,3 +295,36 @@ void vmpu_arch_init(void)
     /* init memory protection */
     vmpu_mem_init();
 }
+
+int vmpu_is_region_size_valid(uint32_t size)
+{
+    /* Align size to 32B */
```
Patater: Add period

core/system/src/mpu/vmpu_freescale_k64.c
```diff
@@ -296,3 +295,36 @@ void vmpu_arch_init(void)
...
+uint32_t vmpu_round_up_region(uint32_t addr, uint32_t size)
+{
+    if (!vmpu_is_region_size_valid(size)) {
+        /* Region size must be valid! */
+        return 0;
+    }
+    /* Alignment is always 32B */
+    const uint32_t mask = 31;
+
+    /* Adding the mask can overflow */
```
Patater: Add period

core/system/src/mpu/vmpu_freescale_k64.c
```diff
@@ -296,3 +295,36 @@ void vmpu_arch_init(void)
...
+    if (masked_size < 32 || (1 << 29) < masked_size) {
+        /* 2^5 == 32, which is the minimum region size */
+        /* 2^29 == 512M, which is the maximum region size */
+        return 0;
+    }
+    /* There is no rounding, we only care about an exact match! */
+    return (masked_size == size);
+}
+
+uint32_t vmpu_round_up_region(uint32_t addr, uint32_t size)
```
Patater: Can the implementation of `vmpu_round_up_region` be shared between these two MPUs? The code looks pretty much the same, other than which `vmpu_region_size_valid` function to call.

Patater:
Some commit organization things
 - Remove period from end of `Add is_region_valid and round_up_region to vmpu.`.
 - Squash `Add page allocator exports` with `Add page allocator SVC calls for API` and `Add page allocator implementation`. It's a three-way squash. All commits provide implementation of the page allocator API.
- Squash `Add page heap start and end address to config` into `Initialize page heap with start and end address`. It's implementation detail for initializing the page heap.

api/inc/page_allocator_exports.h
```diff
@@ -0,0 +1,60 @@
...
+#ifndef __UVISOR_API_PAGE_ALLOCATOR_EXPORTS_H__
+#define __UVISOR_API_PAGE_ALLOCATOR_EXPORTS_H__
+
+#include "api/inc/halt_exports.h"
+#include <stdint.h>
+#include <stddef.h>
+
+
+#define UVISOR_ERROR_PAGE_OK                    (0)
+#define UVISOR_ERROR_PAGE_OUT_OF_MEMORY         (UVISOR_ERROR_CLASS_PAGE + 100)
```
Patater: Since you have error classes now, it probably isn't necessary to start at 100, but if you want to that is fine.

api/inc/page_allocator_exports.h
```diff
@@ -0,0 +1,60 @@
...
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_TABLE    (UVISOR_ERROR_CLASS_PAGE + 101)
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_SIZE     (UVISOR_ERROR_CLASS_PAGE + 102)
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_ORIGIN   (UVISOR_ERROR_CLASS_PAGE + 103)
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_OWNER    (UVISOR_ERROR_CLASS_PAGE + 104)
+#define UVISOR_ERROR_PAGE_INVALID_PAGE_COUNT    (UVISOR_ERROR_CLASS_PAGE + 105)
+
+
+/* Must be a power of 2 for MPU alignment in ARMv7-M. */
```
Patater: Add a static assert to check this alignment.

api/inc/page_allocator_exports.h
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+/* Create a page table with enough pages to hold `size`. */
+#define UVISOR_PAGE_TABLE_FOR_SIZE(size) UVISOR_PAGE_TABLE(UVISOR_PAGES_FOR_SIZE(size))
+
+
+typedef struct {
+    uint32_t page_size;     /* The page size in bytes */
+    uint32_t page_count;    /* The number of pages in the page table */
+    void * page_origins[1]; /* Table of pointer to the origins of a page */
```
Patater: s/pointer/pointers

Because of `origins`, this reads as if there are multiple origins per page. Is that correct?

core/system/inc/page_allocator.h
```diff
@@ -0,0 +1,51 @@
...
+ * used by the heap address, due to adhering to alignment restrictions.
+ *
+ * @param heap_start[in] Start address of the page heap
+ * @param heap_end[in]   End address of the page heap
+ * @param page_size      Pointer to page size for the allocator
+ */
+void page_allocator_init(void * const heap_start, void * const heap_end, const uint32_t * const page_size);
+
+/* Allocate a number of requested pages with the requested page size.
+ * @param table.page_size[in]     Must be equal to the current page size.
```
Patater: Remove period.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+}
+
+#endif /* defined(UVISOR_PRESENT) && (UVISOR_PRESENT == 1) */
+
+/* By default a maximum of 16 pages are allowed. This can only
+ * be decided by the porting engineer for the current platform! */
+#ifndef UVISOR_PAGE_TABLE_MAX_COUNT
+#define UVISOR_PAGE_TABLE_MAX_COUNT ((size_t) 16)
+#endif
+/* The page box_id is the box id which is 8bit large  */
```
Patater: s/8bit/8-bit

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+typedef uint8_t page_owner_t;
+/* Maps the page to the owning box handle */
```
Patater: Use imperative mood.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+static page_owner_t g_page_owner_table[UVISOR_PAGE_TABLE_MAX_COUNT];
+/* define a unused value for the page table */
```
Patater: Capitalize comment beginnings.

core/system/src/mpu/vmpu_armv7m.c
niklas-arm: I don't think adding a period contributes at all to making this code more understandable or maintainable or more secure.
This is an internal comment, not API documentation.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+        HALT_ERROR(SANITY_CHECK_FAILED,
+            "Page heap start pointer (0x%08x) is not in sram memory.\n",
+            (unsigned int) heap_start);
+    }
+    if (!heap_end || !vmpu_sram_addr((uint32_t) heap_end)) {
+        HALT_ERROR(SANITY_CHECK_FAILED,
+            "Page heap end pointer (0x%08x) is not in sram memory.\n",
+            (unsigned int) heap_end);
+    }
+    if (heap_end < heap_start) {
```
Patater: Is having `heap_end == heap_start` (a size zero heap) not allowed?

core/system/src/mpu/vmpu_freescale_k64.c
niklas-arm: YES, I GET IT.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+        HALT_ERROR(SANITY_CHECK_FAILED,
+            "Page heap end pointer (0x%08x) is smaller than heap start pointer (0x%08x).\n",
+            (unsigned int) heap_end,
+            (unsigned int) heap_start);
+    }
+    if ((*page_size < 1024) || !vmpu_is_region_size_valid(*page_size)) {
```
Patater: What went into the decision to make 1024 the minimum page size? Add a comment in the code.

core/system/src/mpu/vmpu_freescale_k64.c
niklas-arm: I thought so too, but the mask is different. On ARMv7-M the mask is the (region size - 1), on K64F it's always (32 - 1), regardless of region size.

core/system/src/mpu/vmpu_freescale_k64.c
niklas-arm: `s/,/;`?

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+    /* Minimal 32B alignment */
+    return ((size & ~31) == size);
+}
+static inline uint32_t vmpu_round_up_region(uint32_t addr, uint32_t size)
+{
+    if (!vmpu_is_region_size_valid(size)) {
+        return 0;
+    }
+    const uint32_t mask = size - 1;
+    return (addr + mask) & ~mask;
```
Patater: What would happen if adding mask to addr overflowed?

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+    uint32_t start = vmpu_round_up_region((uint32_t) heap_start, *page_size);
+    if (start == 0) {
+        HALT_ERROR(SANITY_CHECK_FAILED,
+            "Page heap start address 0x%08x cannot be aligned with page size %uB!\n",
+            (unsigned int) heap_start,
+            (unsigned int) *page_size);
+    }
+
+    g_page_size = *page_size;
+    /* This is the page heap start address */
```
Patater: Add period

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+    g_page_heap_start = (void *) start;
+
+    /* How many pages can we fit in here? */
+    if ((start + g_page_size) > (uint32_t) heap_end) {
+        g_page_count_total = 0;
+    }
+    else {
```
Patater: Put the `}` on the same line as the `else {`.

api/inc/page_allocator_exports.h
niklas-arm: This is actually half true.
It must be power of 2 for the ARM MPU, but can be 32B-multiple for K64F MPU.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+        g_page_count_total = ((uint32_t) heap_end - start) / g_page_size;
+    }
+    /* Clamp page count to table size */
```
Patater: Grammars... :(

I'll stop bothering you about it.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+}
+static inline void page_table_write(uint32_t addr, uint32_t data)
+{
+    *((uint32_t *) addr) = data;
+}
+
+#endif /* defined(UVISOR_PRESENT) && (UVISOR_PRESENT == 1) */
+
+/* By default a maximum of 16 pages are allowed. This can only
+ * be decided by the porting engineer for the current platform! */
```
Patater: Add comments about the rationale or things a porting engineer must consider when decided on this value. Explain the rationale behind choosing 16.

core/system/src/mpu/vmpu_freescale_k64.c
niklas-arm: The period is clearly relevant to the security of this code.

api/inc/page_allocator_exports.h
niklas-arm: I can't add a static check. The page size is passed through the uvisor-input.S, so can only be checked at runtime.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+    /* point to the first entry in the table */
+    void * * page_table = table->page_origins;
+
+    /* Iterate through the page table and find the empty pages */
+    size_t page = 0;
+    for (; (page < g_page_count_total) && pages_required; page++) {
+        /* If the page is unused, it's entry is UVISOR_PAGE_UNUSED (not NULL!) */
+        if (g_page_owner_table[page] == UVISOR_PAGE_UNUSED) {
+            /* Marry this page to the box id */
+            g_page_owner_table[page] = box_id;
```
Patater: Is it possible that another call to `page_allocator_malloc` in a different thread or interrupt could claim a page after our check, but before we set it for this box_id? If this function assumes a lock is held before being called, please add a comment explaining the assumptions. If we don't assume a lock is held, we could use an atomic test and set built-in if any re-entrancy is required.

api/inc/page_allocator_exports.h
Patater: It would be a static check for programmer sanity, not security. We'd still need the check inside uVisor. Having a static check here earlier would help the programmer catch his mistake sooner and in an obvious way: he wouldn't have to pull out the debugger to see the uVisor halt sanity check failed message.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+        /* Range check the returned pointer */
+        if (page < g_page_heap_start || page >= g_page_heap_end) {
+            DPRINTF("uvisor_page_free: FAIL: Pointer 0x%08x does not belong to any page!\n\n", (unsigned int) page);
+            return UVISOR_ERROR_PAGE_INVALID_PAGE_ORIGIN;
+        }
+        /* Compute the index for the pointer */
+        size_t page_index = (page - g_page_heap_start) / g_page_size;
+        /* Check if the page belongs to the caller */
+        if (g_page_owner_table[page_index] == box_id) {
+            free_mask[page_index/32] |= (1 << (page_index & 31));
```
Patater: Place spaces around arithmetic operators, like `/`.

api/inc/page_allocator_exports.h
niklas-arm: Yes, but the static check would need to be specialized on MPU type.
ie. power of two for ARM MPU, 32 multiple for K64F.
Not sure if we want that in a header file.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,287 @@
...
+    /* Iterate over the bits in the free mask and actually free the pages */
+    size_t count = 0;
+    for (; count < g_page_count_total; count++) {
+        if (free_mask[count/32] & (1 << (count & 31))) {
+            g_page_owner_table[count] = UVISOR_PAGE_UNUSED;
+            DPRINTF("uvisor_page_free: Freeing page at index %u\n", count);
+        }
+    }
+    /* We have freed some pages */
+    g_page_count_free += page_count;
```
Patater: Updates to `g_page_count_free` must also be atomic.

api/inc/page_allocator_exports.h
Patater: Nah, wouldn't want that specialization just for a static assert. We could consider forcing everybody to use the least common denominator MPU capabilities (forcing everybody to use power of 2 alignment so their code is portable), but that's probably not worth it.

At least update your comment to reflect reality, or delete it so it doesn't confuse anybody. We'll have to provide more detailed rules about setting the `UVISOR_PAGE_SIZE` in our porting guide.

core/system/src/page_allocator.c
niklas-arm: This assumes being called in an SVC, which makes this mutually exclusive.
But, since this same file is used for unsupported, this does indeed lead to a race condition.

niklas-arm: Going to implement a mutex for alloc and free for unsupported only.

niklas-arm:
Addressed review and squashed commits.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,333 @@
...
+            return UVISOR_ERROR_PAGE_INVALID_PAGE_OWNER;
+        }
+    }
+    /* We now have validated the pages in the table.
+     * All pages that need to be freed are in the free_mask. */
+
+    /* Iterate over the bits in the free mask and actually free the pages. */
+    uint32_t count = 0;
+    for (; count < g_page_count_total; count++) {
+        if (free_mask[count/32] & (1 << (count & 31))) {
```
Patater: Place spaces around arithmetic operators, like `/`.

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,333 @@
...
+#define vmpu_public_flash_addr(...) 1
+#define vmpu_sram_addr(...) 1
+/* HALT_ERROR can be provided by the OS. For example using `assert`. */
+#ifndef HALT_ERROR
+#define HALT_ERROR(id, ...) DPRINTF(__VA_ARGS__)
+#endif
+/* Mutex implementation must be provided by the OS, since the page allocator
+ * is called directly from user code instead of through an SVC call. */
+#ifndef UVISOR_PAGE_ALLOCATOR_MUTEX_AQUIRE
+#define UVISOR_PAGE_ALLOCATOR_MUTEX_AQUIRE {}
```
Patater: Should we `#error` here instead of silently failing to acquire a mutex if the uVisor porting engineer forgot to implement this macro?

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    /* Get the calling box id. */
+    const page_owner_t box_id = g_active_box;
+    DPRINTF("uvisor_page_malloc: Requesting %u pages with size %uB for box %u\n", pages_required, page_size, box_id);
+
+    /* Update the free pages count. */
+    g_page_count_free -= pages_required;
+    /* Point to the first entry in the table. */
+    void * * page_table = table->page_origins;
```
meriac: @niklas-arm page_table_read ?

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,285 @@
...
+            (unsigned int) heap_start,
+            (unsigned int) *page_size);
+    }
+
+    g_page_size = *page_size;
+    /* This is the page heap start address. */
+    g_page_heap_start = (void *) start;
+
+    /* How many pages can we fit in here? */
+    if ((start + g_page_size) > (uint32_t) heap_end) {
```
AlessandroA: Is this guaranteed not to overflow?

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,285 @@
...
+            DPRINTF("uvisor_page_malloc: Found an empty page 0x%08x entry at index %u\n", (unsigned int) ptr, page);
+        }
+    }
+    DPRINTF("uvisor_page_malloc: %u free pages remaining.\n\n", g_page_count_free);
+
+    UVISOR_PAGE_ALLOCATOR_MUTEX_RELEASE;
+    return UVISOR_ERROR_PAGE_OK;
+}
+
+int page_allocator_free(const UvisorPageTable *const table)
```
AlessandroA: `* const`

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,285 @@
...
+    const page_owner_t box_id = g_active_box;
+    int table_size = page_count;
+
+    /* Contains the bit mask of all the pages we need to free. */
+    uint32_t free_mask[(UVISOR_PAGE_TABLE_MAX_COUNT + 31) / 32] = {0};
+    /* Iterate over the table and validate each pointer. */
+    void * const * page_table = table->page_origins;
+
+    for (; table_size > 0; page_table++, table_size--) {
+        void *page = (void *) page_table_read((uint32_t) page_table);
```
AlessandroA: `void * page`

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,285 @@
...
+            DPRINTF("uvisor_page_free: FAIL: Pointer 0x%08x does not belong to any page!\n\n", (unsigned int) page);
+            UVISOR_PAGE_ALLOCATOR_MUTEX_RELEASE;
+            return UVISOR_ERROR_PAGE_INVALID_PAGE_ORIGIN;
+        }
+        /* Compute the index for the pointer. */
+        uint32_t page_index = (page - g_page_heap_start) / g_page_size;
+        /* Check if the page belongs to the caller. */
+        if (g_page_owner_table[page_index] == box_id) {
+            free_mask[page_index / 32] |= (1 << (page_index & 31));
+        }
```
AlessandroA:
```diff
-        }
-        /* Abort if the page doesn't belong to the caller. */
-        else {
-            if (g_page_owner_table[page_index] == UVISOR_PAGE_UNUSED) {

+        } else {
+            /* Abort if the page doesn't belong to the caller. */
+            if (g_page_owner_table[page_index] == UVISOR_PAGE_UNUSED) {
```

core/system/src/page_allocator.c
```diff
@@ -0,0 +1,285 @@
...
+    /* Get the calling box id. */
+    const page_owner_t box_id = g_active_box;
+    int table_size = page_count;
+
+    /* Contains the bit mask of all the pages we need to free. */
+    uint32_t free_mask[(UVISOR_PAGE_TABLE_MAX_COUNT + 31) / 32] = {0};
+    /* Iterate over the table and validate each pointer. */
+    void * const * page_table = table->page_origins;
+
+    for (; table_size > 0; page_table++, table_size--) {
```
AlessandroA: Can you move `int table_size = page_count;` here (right before the loop)?

api/src/uvisor-input.S
```diff
@@ -84,6 +85,11 @@ uvisor_config:
     /* pointer to __uvisor_box_context */
     .long __uvisor_box_context
 
+    /* start and end address and page size of page heap */
```
AlessandroA: Can you replace this comment with something like `/* Page allocator heap boundaries and size */`? See [this commit](https://github.com/ARMmbed/uvisor/pull/233/commits/d1128a81846fed3f584b8677bbd46393a8ae651d) for reference.

core/system/inc/linker.h
```diff
@@ -65,6 +65,10 @@ typedef struct {
     /* address of __uvisor_box_context */
     uint32_t **uvisor_box_context;
 
+    /* heap for page allocator */
```
AlessandroA: Can you replace this comment with something like `/* Page allocator heap boundaries and size */`? See [this commit](https://github.com/ARMmbed/uvisor/pull/233/commits/d1128a81846fed3f584b8677bbd46393a8ae651d) for reference.

core/system/src/page_allocator.c
niklas-arm: Moved to https://github.com/ARMmbed/uvisor/pull/232/commits/2d9ae6bc52a54f919c65881e01d604135aa61f68#diff-c71a307c4fd911388ca9a0399570ac88R99

core/system/inc/linker.h
```diff
@@ -65,6 +65,10 @@ typedef struct {
...
+    uint32_t * page_start, * page_end;
```
AlessandroA: Please move on 2 lines for consistency with [this commit](https://github.com/ARMmbed/uvisor/pull/233/commits/d1128a81846fed3f584b8677bbd46393a8ae651d):

```
uint32_t * page_start;
uint32_t * page_end;
```
Thanks!

core/system/src/page_allocator.c
niklas-arm: Good point, fixed.

PR #228, opened 2016-05-31
Move vmpu_unpriv_access.h to the API folder

AlessandroA:
Needed for #227.

@niklas-arm @meriac @Patater


niklas-arm:
Is the GCC inline asm a problem for ARMCC or IARCC?

core/system/inc/mpu/vmpu_unpriv_access.h
```diff
@@ -17,6 +17,17 @@
 #ifndef __VMPU_UNPRIV_ACCESS_H__
 #define __VMPU_UNPRIV_ACCESS_H__
 
+/** Write an 8 bit value to an address as if in unprivileged mode.
```
Patater: s/8 bit/8-bit

core/system/inc/mpu/vmpu_unpriv_access.h
```diff
@@ -25,6 +36,17 @@ static UVISOR_FORCEINLINE void vmpu_unpriv_uint8_write(uint32_t addr, uint8_t da
     );
 }
 
+/** Write a 16 bit value to an address as if in unprivileged mode.
```
Patater: s/16 bit/16-bit

core/system/inc/mpu/vmpu_unpriv_access.h
```diff
@@ -33,13 +55,35 @@ static UVISOR_FORCEINLINE void vmpu_unpriv_uint16_write(uint32_t addr, uint16_t
     );
 }
 
+/** Write a 32 bit value to an address as if in unprivileged mode.
```
Patater: s/32 bit/32-bit

Patater:
LGTM

AlessandroA:
@niklas-arm Good point. Intrinsics should be the best option here. @meriac I found [this](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0472j/chr1359124997817.html) which implies the right opcode is used for the right pointer. I'll change the explicit asm inline into intrinsics.

PR #229, opened 2016-05-31
Add box index structure and initialize it

niklas-arm:
This introduces a box index in box bss memory, augmenting the concept of the `ctx` pointer with additional information such the box config, heaps, box id.
uVisor initializes the box index and allows for the user to declare additional reserved box index space by overwriting the `__uvisor_box_index_size` constant.

Note: This builds directly on top of #227, but does not directly depend on it.

cc @Patater @AlessandroA @meriac


niklas-arm:
Argh. Wrong branch.

PR #231, opened 2016-05-31
3. Add tier-2 allocator and initialization

niklas-arm:
This adds the required glue code for connecting the tier-2 with the tier-1 allocator incl. initialization of heaps on first allocation.
Also adds mutex protection to process heap via extended box index.

Note: This builds directly on top of #230 and depends on it!

cc @Patater @AlessandroA @meriac

PR #232, opened 2016-05-31
4. Add initialization of box index and page allocator for unsupported targets

niklas-arm:
This adds initialization for unsupported targets, where the page heap is placed inside the main heap, while leaving the main heap intact, which means this is compatible with existing CMSIS linker scripts.

Note: This builds on #231 and depends on it!

Actual diff:
- Move unsupported.c into rtx/src folder.
- Add uvisor_malloc_init for unsupported targets

cc @Patater @AlessandroA @meriac

### DIFF
--- a/api/rtx/src/unsupported_malloc.c
+++ b/api/rtx/src/unsupported_malloc.c
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2015-2016, ARM Limited, All Rights Reserved
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "uvisor-lib/uvisor-lib.h"
+
+#if !(defined(UVISOR_PRESENT) && (UVISOR_PRESENT == 1))
+
+#include <string.h> /* for memset */
+
+/* Forward declaration of the page allocator API. */
+extern void page_allocator_init(void * const heap_start, void * const heap_end, const uint32_t * const page_size);
+
+extern uint32_t __end__[];      /* __heap_start */
+extern uint32_t __HeapLimit[];  /* __heap_end   */
+
+extern uint32_t __StackLimit[];   /* bottom of stack */
+
+/* There is only one box index for box 0. */
+RtxBoxIndex * __uvisor_ps;
+
+static void box_index_init(void *box_bss, uint32_t heap_size)
+{
+    const uint32_t index_size = sizeof(RtxBoxIndex);
+    /* Adjust size for overhead of box index */
+    heap_size -= index_size;
+
+    /* The box index is at the beginning of the bss section */
+    RtxBoxIndex *const indexOS = box_bss;
+    /* Zero the _entire_ index, so that we know to initialize the mutex on
+     * first use! */
+    memset(box_bss, 0, index_size);
+    /* Initialize user context */
+    indexOS->index.ctx = NULL;
+    /* Initialize box heap */
+    indexOS->index.box_heap = box_bss + index_size;
+    indexOS->index.box_heap_size = heap_size;
+    /* Active heap pointer is NULL */
+    indexOS->index.active_heap = NULL;
+
+    /* There is no box config for unsupported! */
+    indexOS->index.config = NULL;
+
+    /* Set the index */
+    __uvisor_ps = indexOS;
+}
+
+void secure_malloc_init(void)
+{
+    /* get the main heap size from the linker script */
+    uint32_t heap_size = ((uint32_t) __HeapLimit -
+                          (uint32_t) __end__);
+    /* Main heap size is aligned to page boundaries n*UVISOR_PAGE_SIZE */
+    uint32_t heap_start = (uint32_t) __StackLimit - heap_size;
+    /* align the start address of the main heap to a page boundary */
+    heap_start &= ~(UVISOR_PAGE_SIZE - 1);
+    /* adjust the heap size to the new heap start address */
+    heap_size = (uint32_t) __StackLimit - heap_start;
+
+    /* page heap now extends from the previous main heap start address
+     * to the new main heap start address */
+    extern uint32_t __uvisor_page_size;
+    page_allocator_init(__end__, (void *) heap_start, &__uvisor_page_size);
+    box_index_init((void *) heap_start, heap_size);
+}
+
+#endif
--- a/api/rtx/src/rtx_malloc_wrapper.c
+++ b/api/rtx/src/rtx_malloc_wrapper.c
@@ -55,7 +55,12 @@ static int init_allocator()
 {
     int ret = 0;
     if (__uvisor_ps == NULL) {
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -98,6 +98,7 @@ SOURCES:=\
          $(CORE_SYSTEM_DIR)/src/context.c \
          $(CORE_SYSTEM_DIR)/src/halt.c \
          $(CORE_SYSTEM_DIR)/src/main.c \
+         $(CORE_SYSTEM_DIR)/src/page_allocator.c \
```
niklas-arm: test

api/rtx/src/unsupported_malloc.c
```diff
@@ -0,0 +1,146 @@
...
+ * Be aware that the page allocator is not re-entrant, so the OS must provide a
+ * mutex implementation to enable thread-safety!
+ */
+#define DPRINTF(...) {}
+#define g_active_box 0
+#define vmpu_is_box_id_valid(...) 0
+#define vmpu_public_flash_addr(...) 1
+#define vmpu_sram_addr(...) 1
+#define HALT_ERROR(id, ...) assert(__VA_ARGS__)
+#define UVISOR_PAGE_ALLOCATOR_MUTEX_AQUIRE {}
```
niklas-arm: Requires implementation still!

api/rtx/src/unsupported_malloc.c
```diff
@@ -0,0 +1,143 @@
...
+ */
+#include "uvisor-lib/uvisor-lib.h"
+
+#if !(defined(UVISOR_PRESENT) && (UVISOR_PRESENT == 1))
+
+#include <string.h>
+
+/* Note: This file is not included in the uVisor release library. Instead, the
+ *       host OS needs to compile it separately if a platform does not support
+ *       uVisor (but uVisor API header files are still used). */
```
Patater: uVisor API header files should not be used in unsupported applications. We need to think about how to deliver this code.

api/rtx/src/unsupported_malloc.c
```diff
@@ -0,0 +1,143 @@
...
+int page_allocator_free(const UvisorPageTable * const table);
+
+extern uint32_t __end__[];      /* __heap_start */
+extern uint32_t __HeapLimit[];  /* __heap_end   */
+
+extern uint32_t __StackLimit[];   /* bottom of stack */
+
+RtxBoxIndex * __uvisor_ps;
+
+static void box_index_init(void *box_bss, uint32_t heap_size)
```
Patater: This function seems like a copy of some code I reviewed earlier, for inside uVisor. Be sure it is in sync with that implementation, if any improvements were made in the uVisor implementation.

Patater:
> Move unsupported.c into rtx/src folder

What motivates this move? I would expect only RTX-specific files in the rtx/src folder, and I don't see any changes to the file in this PR that would cause unsupported.c to become RTX-specific

niklas-arm:
> What motivates this move?

Mostly to get it out of the src folder, since it is not compiled by uVisor and not part of the compiled uVisor archive.
It's not part of RTX either, so I understand that it doesn't really fit there either, but it's still the better choice to move it.

Patater:
Given that reasoning, I would prefer leaving it where it is. rtx should be just for rtx-specific things. The file is already being excluded from the .a file properly.

niklas-arm:
> The file is already being excluded from the .a file properly.

But it's still confusing and is mental overhead. It's not consistent from a build system PoV either, there are special rules to exclude it in the uVisor Makefile and special rules to include it in uVisor-Lib Makefile.
It feels very forced to leave it there.

Patater:
It's not important to fix it in this PR. File an issue to make a folder for unsupported that could be shared between multiple OSes and is excluded from the `.a`.

api/rtx/src/unsupported_malloc.c
```diff
@@ -0,0 +1,162 @@
...
+{
+    return *((uint32_t *) addr);
+}
+static inline void page_table_write(uint32_t addr, uint32_t data)
+{
+    *((uint32_t *) addr) = data;
+}
+
+/* Include the original page allocator source directly. */
+#include "source/page_allocator.c_inc"
```
niklas-arm: > I understand why you did this but it makes unicorns cry.

😭 

> Would it still be ok if you instead save this `rtx_malloc_wrapper.c` file as a .h file and then include it conditionally in `page_allocator.c`?

But then I have the same, just the other way around. It's not better, just because it's a `.h` file.
I'd still have to copy the `page_allocator.c` file over from uVisor core. but this way, no unsupported code is in uVisor core.

api/rtx/src/unsupported_malloc.c
```diff
@@ -0,0 +1,162 @@
...
+
+    /* page heap now extends from the previous main heap start address
+     * to the new main heap start address */
+    extern uint32_t __uvisor_page_size;
+    page_allocator_init(__end__, (void *) heap_start, &__uvisor_page_size);
+    box_index_init((void *) heap_start, heap_size);
+}
+
+
+int uvisor_page_malloc(UvisorPageTable *const table)
```
meriac: @AlessandroA @niklas-arm - needs to be renamed, too for consistency I presume (secure_*) ?

api/rtx/src/unsupported_malloc.c
```diff
@@ -0,0 +1,162 @@
...
+{
+    return page_allocator_free(table);
+}
+
+/* This is the fallback implementation for using the page allocator from uVisor
+ * inside an OS as a normal function.
+ * Be aware that the page allocator is not re-entrant, so the OS must provide a
+ * mutex implementation to enable thread-safety!
+ */
+#define DPRINTF(...) {}
```
meriac: @niklas-arm move these lines to the start?

api/rtx/src/unsupported_malloc.c
```diff
@@ -0,0 +1,162 @@
...
+        if (g_page_allocator_mutex_id == NULL) {
+            /* Mutex failed to be created. */
+            return;
+        }
+    }
+
+    osMutexWait(g_page_allocator_mutex_id, osWaitForever);
+}
+
+#define UVISOR_PAGE_ALLOCATOR_MUTEX_AQUIRE  page_allocator_mutex_aquire()
```
meriac: @niklas-arm - these, too ?

meriac:
@AlessandroA @niklas-arm @Patater : page_allocator.c could have a conditional .h include depending on whether the .c is compiled in a uvisor context or in unsupported. The unsupported header file does not need to be included in the supported tree and vice versa. Potentially the header file name can be identical for bot supported and unsupported ("page_allocator_config.h") and have different content for supported/unsupported.

api/rtx/src/unsupported_malloc.c
niklas-arm: This is the unsupported implementation of [the uVisor API](https://github.com/ARMmbed/uvisor/blob/dev/api/inc/page_allocator.h#L30-L35).
The uVisor enabled [implementation goes through an SVC](https://github.com/ARMmbed/uvisor/blob/dev/api/src/page_allocator.c#L21-L29), which is not available or useful for unsupported targets.

This is the first time we are implementing a core uVisor API.
Technically the user is not required to use the tier-2 allocator (ie. the API `secure_allocator.h`), but can call `uvisor_page_malloc` directly.

api/rtx/src/unsupported_malloc.c
niklas-arm: Maybe I should move this stuff into it's own file `unsupported_page_allocator.c`?

AlessandroA:
LGTM.

AlessandroA:
cc @meriac

meriac:
@niklas-arm unsupported.c lives in {mbedmicro/mbed}/uvisor

Patater:
Other than period at end of commit message, LGTM, now that we've dropped the rtx folder moving commit.

meriac:
Deferring issue about *.inc include of source code and merging now - see https://github.com/ARMmbed/uvisor/issues/239

Issue #90, opened 2015-10-28
Add symbol support to uvisor_read & uvisor_write APIs


meriac:
- currently uvisor_read & *_write do not accept structures member addresses
- affects user experience as addresses need to be calculated on the fly

Example:
```C
uvisor_read(&UART0->BAUD_RATE)
```
needs currently to be written as:
```C
uvisor_read(UART0_ADDRESS + UART_BAUD_RATE_OFFSET)
```

rainierwolfcastle:
ARM Internal Ref: IOTSFW-1121

meriac:
Resolved by the new Register Level Access API: https://github.com/mbedmicro/mbed/blob/master/hal/targets/cmsis/core_cmSecureAccess.h

PR #233, opened 2016-06-01
Register level gateway

AlessandroA:
This PR introduces the Register Gateway implementation.
- Added the issue #241 to track the runtime OR operation on the gateway function pointer independently of this pull request.

Done so far:

* Remove `uvisor_{read,write}` APIs restrictions (see #90) by using  a `struct` instead of SVCall metadata.
* All operations implemented.
* Added APIs for common bit manipulations.
* All operations tested.

Please concentrate on the last 2 commits.

@meriac @Patater @niklas-arm

### DIFF
--- a/api/inc/register_gateway.h
+++ b/api/inc/register_gateway.h
@@ -32,8 +32,8 @@
 #error "Unsupported instruction set. The ARM Thumb-2 instruction set must be supported."
 #endif /* __thumb__ && __thumb2__ */
 
-/* register gateway operations */
-/* note: do not use special characters as these numbers will be stringified */
+/* Register gateway operations */
+/* Note: Do not use special characters as these numbers will be stringified. */
 #define UVISOR_OP_READ(op)  (op)
 #define UVISOR_OP_WRITE(op) ((1 << 4) | (op))
 #define UVISOR_OP_NOP       0x0
@@ -41,17 +41,17 @@
 #define UVISOR_OP_OR        0x2
 #define UVISOR_OP_XOR       0x3
 
-/* default mask for whole register operatins */
+/* Default mask for whole register operatins. */
 #define __UVISOR_OP_DEFAULT_MASK 0x0
 
-/* register gateway metadata */
+/* Register gateway metadata */
 #if defined(__CC_ARM)
 
 /* TODO/FIXME */
 
 #elif defined(__GNUC__)
 
-/* 1 argument: simple read, no mask */
+/* 1 argument: Simple read, no mask */
 #define __UVISOR_REGISTER_GATEWAY_METADATA1(src_box, addr) \
     "b.n skip_args%=\n" \
     ".word " UVISOR_TO_STRING(UVISOR_REGISTER_GATEWAY_MAGIC) "\n" \
@@ -61,7 +61,7 @@
     ".word " UVISOR_TO_STRING(__UVISOR_OP_DEFAULT_MASK) "\n" \
     "skip_args%=:\n"
 
-/* 2 arguments: simple write, no mask */
+/* 2 arguments: Simple write, no mask */
 #define __UVISOR_REGISTER_GATEWAY_METADATA2(src_box, addr, val) \
     "b.n skip_args%=\n" \
     ".word " UVISOR_TO_STRING(UVISOR_REGISTER_GATEWAY_MAGIC) "\n" \
@@ -71,7 +71,7 @@
     ".word " UVISOR_TO_STRING(__UVISOR_OP_DEFAULT_MASK) "\n" \
     "skip_args%=:\n"
 
-/* 3 arguments: masked read */
+/* 3 arguments: Masked read */
 #define __UVISOR_REGISTER_GATEWAY_METADATA3(src_box, addr, op, mask) \
     "b.n skip_args%=\n" \
     ".word " UVISOR_TO_STRING(UVISOR_REGISTER_GATEWAY_MAGIC) "\n" \
@@ -81,7 +81,7 @@
     ".word " UVISOR_TO_STRING(mask) "\n" \
     "skip_args%=:\n"
 
-/* 4 arguments: masked write */
+/* 4 arguments: Masked write */
 #define __UVISOR_REGISTER_GATEWAY_METADATA4(src_box, addr, val, op, mask) \
     "b.n skip_args%=\n" \
     ".word " UVISOR_TO_STRING(UVISOR_REGISTER_GATEWAY_MAGIC) "\n" \
@@ -98,10 +98,11 @@
                                               __UVISOR_REGISTER_GATEWAY_METADATA3, \
                                               __UVISOR_REGISTER_GATEWAY_METADATA2, \
                                               __UVISOR_REGISTER_GATEWAY_METADATA1, \
-                                              /* no macro for 0 args */          )(src_box, ##__VA_ARGS__)
+                                              /* No macro for 0 args */          )(src_box, ##__VA_ARGS__)
 
-/* register-level gateway - read */
-/* FIXME currently only a hardcoded 32bit constant can be used for the addr field */
+/* Rregister-level gateway - Read */
+/* FIXME: Currently only a hardcoded 32bit constant can be used for the addr
+ *        field. */
 #define uvisor_read(src_box, ...) \
     ({ \
         uint32_t res = UVISOR_SVC(UVISOR_SVC_ID_REGISTER_GATEWAY, \
@@ -109,7 +110,7 @@
         res; \
     })
 
-/* register-level gateway - write */
+/* Register-level gateway - Write */
 #define uvisor_write(src_box, addr, val, ...) \
     ({ \
         UVISOR_SVC(UVISOR_SVC_ID_REGISTER_GATEWAY, \
--- a/api/inc/register_gateway.h
+++ b/api/inc/register_gateway.h
@@ -17,105 +17,212 @@
 #ifndef __UVISOR_API_REGISTER_GATEWAY_H__
 #define __UVISOR_API_REGISTER_GATEWAY_H__
 
+#include "api/inc/register_gateway_exports.h"
 #include "api/inc/uvisor_exports.h"
 #include <stdint.h>
 
-/* The following magic is used to verify that a register gateway structure. It
- * has been generated starting from the ARM Thumb/Thumb-2 invalid opcode.
- * ARM Thumb (w/o the Thumb-2 extension): 0xF7FxAxxx (32 bits)
- * The 'x's can be freely chosen (they have been chosen randomly here).
- * This requires the Thumb-2 extensions because otherwise only 16 bits opcodes
- * are accepted.*/
-#if defined(__thumb__) && defined(__thumb2__)
-#define UVISOR_REGISTER_GATEWAY_MAGIC 0xF7F3A89E
-#else
-#error "Unsupported instruction set. The ARM Thumb-2 instruction set must be supported."
-#endif /* __thumb__ && __thumb2__ */
+/** Get the offset of a struct member.
+ * @internal
+ */
+#define __UVISOR_OFFSETOF(type, member) ((uint32_t) (&(((type *)(0))->member)))
 
-/* Register gateway operations */
-/* Note: Do not use special characters as these numbers will be stringified. */
-#define UVISOR_OP_READ(op)  (op)
-#define UVISOR_OP_WRITE(op) ((1 << 4) | (op))
-#define UVISOR_OP_NOP       0x0
-#define UVISOR_OP_AND       0x1
-#define UVISOR_OP_OR        0x2
-#define UVISOR_OP_XOR       0x3
+/** Generate the opcode of the 16-bit Thumb-2 16-bit T2 encoding of the branch
+ *  instruction.
+ * @internal
+ * @note The branch instruction is encoded according to the Thumb-2 immediate
+ * encoding rules:
+ *     <instr_addr>: B.N <label>
+ *     imm = (<label>_addr - PC) / 2
+ * Where:
+ *     PC = <instr>_addr + 4
+ * The +4 is to account for the pipelined PC at the time of the branch
+ * instruction. See ARM DDI 0403E.b page A4-102 for more details.
+ * @param instr[in] Address of the branch instruction
+ * @param label[in] Address of the label
+ * @returns the 16-bit encoding of the B.N <label> instruction.
+ */
+#define BRANCH_OPCODE(instr, label) \
+    (uint16_t) (0xE000 | (uint8_t) ((((uint32_t) (label) - ((uint32_t) (instr) + 4)) / 2) & 0xFF))
 
-/* Default mask for whole register operatins. */
-#define __UVISOR_OP_DEFAULT_MASK 0x0
+/** `BX LR` encoding
+ * @internal
+ */
+#define BXLR                           ((uint16_t) 0x4770)
 
-/* Register gateway metadata */
-#if defined(__CC_ARM)
-
-/* TODO/FIXME */
-
-#elif defined(__GNUC__)
-
-/* 1 argument: Simple read, no mask */
-#define __UVISOR_REGISTER_GATEWAY_METADATA1(src_box, addr) \
-    "b.n skip_args%=\n" \
-    ".word " UVISOR_TO_STRING(UVISOR_REGISTER_GATEWAY_MAGIC) "\n" \
-    ".word " UVISOR_TO_STRING(addr) "\n" \
-    ".word " UVISOR_TO_STRING(src_box) "_cfg_ptr\n" \
-    ".word " UVISOR_TO_STRING(UVISOR_OP_READ(UVISOR_OP_NOP)) "\n" \
-    ".word " UVISOR_TO_STRING(__UVISOR_OP_DEFAULT_MASK) "\n" \
-    "skip_args%=:\n"
-
-/* 2 arguments: Simple write, no mask */
-#define __UVISOR_REGISTER_GATEWAY_METADATA2(src_box, addr, val) \
-    "b.n skip_args%=\n" \
-    ".word " UVISOR_TO_STRING(UVISOR_REGISTER_GATEWAY_MAGIC) "\n" \
-    ".word " UVISOR_TO_STRING(addr) "\n" \
-    ".word " UVISOR_TO_STRING(src_box) "_cfg_ptr\n" \
-    ".word " UVISOR_TO_STRING(UVISOR_OP_WRITE(UVISOR_OP_NOP)) "\n" \
-    ".word " UVISOR_TO_STRING(__UVISOR_OP_DEFAULT_MASK) "\n" \
-    "skip_args%=:\n"
-
-/* 3 arguments: Masked read */
-#define __UVISOR_REGISTER_GATEWAY_METADATA3(src_box, addr, op, mask) \
-    "b.n skip_args%=\n" \
-    ".word " UVISOR_TO_STRING(UVISOR_REGISTER_GATEWAY_MAGIC) "\n" \
-    ".word " UVISOR_TO_STRING(addr) "\n" \
-    ".word " UVISOR_TO_STRING(src_box) "_cfg_ptr\n" \
-    ".word " UVISOR_TO_STRING(UVISOR_OP_READ(op)) "\n" \
-    ".word " UVISOR_TO_STRING(mask) "\n" \
-    "skip_args%=:\n"
-
-/* 4 arguments: Masked write */
-#define __UVISOR_REGISTER_GATEWAY_METADATA4(src_box, addr, val, op, mask) \
-    "b.n skip_args%=\n" \
-    ".word " UVISOR_TO_STRING(UVISOR_REGISTER_GATEWAY_MAGIC) "\n" \
-    ".word " UVISOR_TO_STRING(addr) "\n" \
-    ".word " UVISOR_TO_STRING(src_box) "_cfg_ptr\n" \
-    ".word " UVISOR_TO_STRING(UVISOR_OP_WRITE(op)) "\n" \
-    ".word " UVISOR_TO_STRING(mask) "\n" \
-    "skip_args%=:\n"
-
-#endif /* __CC_ARM or __GNUC__ */
-
-#define __UVISOR_REGISTER_GATEWAY_METADATA(src_box, ...) \
-     __UVISOR_MACRO_SELECT(_0, ##__VA_ARGS__, __UVISOR_REGISTER_GATEWAY_METADATA4, \
-                                              __UVISOR_REGISTER_GATEWAY_METADATA3, \
-                                              __UVISOR_REGISTER_GATEWAY_METADATA2, \
-                                              __UVISOR_REGISTER_GATEWAY_METADATA1, \
-                                              /* No macro for 0 args */          )(src_box, ##__VA_ARGS__)
-
-/* Rregister-level gateway - Read */
-/* FIXME: Currently only a hardcoded 32bit constant can be used for the addr
- *        field. */
-#define uvisor_read(src_box, ...) \
+/** Register Gateway - Read operation
+ *
+ * This macro provides an API to perform 32-bit read operations on restricted
+ * registers. Such accesses are assembled into a read-only flash structure that
+ * is read and validated by uVisor before performing the operation.
+ *
+ * @warning These APIs currently only support link-time known values. This means
+ * that an access of this type is not supported:
+ * ```
+ * static uint32_t const const_mask = 0x0000F000;
+ * a = uvisor_read(&SYSCFG->CTRL, UVISOR_RGW_OP_READ, const_mask); // Yes
+ * uint32_t var_mask = rand();
+ * a = uvisor_read(&SYSCFG->CTRL, UVISOR_RGW_OP_READ, var_mask); // No
+ * ```
+ *
+ * @param box_name[in]  The name of the source box as decalred in
+ *                      `UVISOR_BOX_CONFIG`.
+ * @param addr[in]      The address for the data access.
+ * @param operation[in] The operation to perform at the address for the read. It
+ *                      is chosen among the `UVISOR_RGW_OP_*` macros.
+ * @param mask[in]      The mask to apply for the read operation.
+ * @returns The value read from address using the operation and mask provided
+ * (or their respective defaults if they have not been provided).
+ */
+#define uvisor_read(box_name, addr, op, msk) \
     ({ \
-        uint32_t res = UVISOR_SVC(UVISOR_SVC_ID_REGISTER_GATEWAY, \
-                                  __UVISOR_REGISTER_GATEWAY_METADATA(src_box, ##__VA_ARGS__)); \
-        res; \
+        /* Instanstiate the gateway. This gets resolved at link-time. */ \
+        __attribute__((aligned(4))) static TRegisterGateway const register_gateway = { \
+            .svc_opcode = UVISOR_SVC_OPCODE(UVISOR_SVC_ID_REGISTER_GATEWAY), \
+            .branch     = BRANCH_OPCODE(__UVISOR_OFFSETOF(TRegisterGateway, branch), \
+                                        __UVISOR_OFFSETOF(TRegisterGateway, bxlr)), \
+            .magic      = UVISOR_REGISTER_GATEWAY_MAGIC, \
+            .box_ptr    = (uint32_t) & box_name ## _cfg_ptr, \
+            .address    = (uint32_t) addr, \
+            .value      = 0, \
+            .mask       = msk, \
+            .operation  = op, \
+            .bxlr       = BXLR  \
+        }; \
+        \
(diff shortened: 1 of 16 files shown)


api/inc/register_gateway.h
```diff
@@ -17,105 +17,146 @@
...
+            .box_ptr    = (uint32_t) & box_name ## _cfg_ptr, \
+            .address    = (uint32_t) addr, \
+            .value      = 0, \
+            .mask       = SELECT_MASK(__VA_ARGS__), \
+            .operation  = SELECT_READ_OP(__VA_ARGS__), \
+            .bxlr       = BXLR  \
+        }; \
+        \
+        /* Call the actual gateway. */ \
+        uint32_t result = ((uint32_t (*)(void)) ((uint32_t) &register_gateway | 1))(); \
```
Patater: So, just casting to the function pointer doesn't set the thumb bit. We have to do this manually as done here. Correct?

api/inc/register_gateway.h
```diff
@@ -17,105 +17,146 @@
...
+ * This macro provides a variadic API to perform 32-bit read operations on
+ * restricted registers. Such accesses are assembled into a read-only flash
+ * structure that is read and validated by uVisor before performing the
+ * operation.
+ *
+ * The following APIs can be used:
+ * void uvisor_read(src_box, address);
+ * void uvisor_read(src_box, address, operation, mask);
+ *
+ * @warning These APIs currently only support link-time known values. This means
```
Patater: What does this warning mean for the `uvisor_read` operation? The following example is for `uvisor_write`. Consider either removing the warning to making the example illustrate what the warning means.

api/inc/register_gateway_exports.h
```diff
@@ -0,0 +1,59 @@
...
+
+/** Register gateway magic
+ *
+ * The following magic is used to verify that a register gateway structure. It
+ * has been generated starting from the ARM Thumb/Thumb-2 invalid opcode.
+ * ARM Thumb (w/o the Thumb-2 extension): 0xF7FxAxxx (32 bits)
+ * The 'x's can be freely chosen (they have been chosen randomly here).
+ * This requires the Thumb-2 extensions because otherwise only 16 bits opcodes
+ * are accepted.*/
+#if defined(__thumb__) && defined(__thumb2__)
```
Patater: Why do we have to check both __thumb__ and __thumb2__? Why not just check __thumb2__?

api/inc/register_gateway.h
```diff
@@ -17,105 +17,146 @@
...
-    ".word " UVISOR_TO_STRING(UVISOR_REGISTER_GATEWAY_MAGIC) "\n" \
-    ".word " UVISOR_TO_STRING(addr) "\n" \
-    ".word " UVISOR_TO_STRING(src_box) "_cfg_ptr\n" \
-    ".word " UVISOR_TO_STRING(UVISOR_OP_WRITE(op)) "\n" \
-    ".word " UVISOR_TO_STRING(mask) "\n" \
-    "skip_args%=:\n"
-
-#endif /* __CC_ARM or __GNUC__ */
+/** Both `uvisor_read` and `uvisor_write` can have 2 additional input arguments -
+ *  mask and operation. These macros select the right argument from
```
Patater: I'd rather not we have optional macro arguments, for code maintainability sake. We need to balance API ease of understanding and use with code maintenance and future refactoring efforts.

I'd prefer named macros like `uvisor_write_mask` that takes a mask and `uvisor_write_op` that takes an op, and `uvisor_write_mask_op` that takes  both. The macros that accept fewer parameters can be implemented in terms of the macros that take more parameters; this maintains the aspect of your current implementation where the default values for parameters easy to identify.

core/system/src/register_gateway.c
```diff
@@ -0,0 +1,95 @@
...
+#include <uvisor.h>
+#include "halt.h"
+#include "register_gateway.h"
+#include "vmpu.h"
+
+/** Perform the validation of a register gateway.
+ * @internal
+ * @warning This function does not verify that the operation, value and mask
+ * make sense together. The code that performs the operation should check for
+ * this and throw an error accordingly.
```
Patater: s/throw/return

core/system/src/mpu/vmpu.c
```diff
@@ -487,3 +482,9 @@ int vmpu_box_namespace_from_id(int box_id, char *box_namespace, size_t length)
 
     return copy_box_namespace(box_cfgtbl[box_id]->box_namespace, box_namespace);
 }
+
+/* TODO */
+void vmpu_register_gateway_check(uint32_t address, uint32_t operation, uint32_t mask)
```
Patater: Can we delete this function? What would we need it for?

api/inc/register_gateway.h
```diff
@@ -17,104 +17,146 @@
...
+ *  `__VA_ARGS__` and provide a sensible default if none is provided.
+ * @internal
+ */
+#define SELECT_MASK(...)       __UVISOR_MACRO_SELECT(_0, ##__VA_ARGS__, , , _MASK , , _DEFAULT_NULL)(__VA_ARGS__)
+#define SELECT_READ_OP(...)    __UVISOR_MACRO_SELECT(_0, ##__VA_ARGS__, , , _OP , , _DEFAULT_READ)(__VA_ARGS__)
```
niklas-arm: Note that the `@internal` only applies to `SELECT_MASK(...)`. You can group it (`@{`,`@}`).

api/inc/register_gateway.h
meriac: @AlessandroA I agree with Jaeden here - that API won't be directly exposed to the user anyway. It's much clearer if we move the macro indirection.

api/inc/register_gateway.h
AlessandroA: Yes. At the moment I've tried the following:

* Just cast to a fn pointer. Doesn't work (Thumb bit not set).
* Save `gtw_ptr | 1` in a `static const` and use that as fn ptr. Doesn't work (`gtw_ptr` not known until link time, the `| 1` is done at runtime.

Please feel free to suggest other approaches!

api/inc/register_gateway.h
AlessandroA: Updated.

api/inc/register_gateway_exports.h
AlessandroA: It is to make sure to select the configuration we know at the moment (Thumb with Thumb-2 extension). Currently that is identical to just having `__thumb2__`, but I thought it was better to spell it out.

api/inc/register_gateway.h
AlessandroA: I removed the variadic macros and only left the full ones. On the other hand, I provided new APIs to perform common operations as set, clear, toggle bits.

api/inc/register_gateway.h
AlessandroA: Right! Thanks!

AlessandroA:
@meriac @Patater @niklas-arm 

* Addressed comments from first round of review.
* Added implementation of all operations.
* Added APIs for common bit manipulations.

Remaining:

* Writing tests for all operations.

**Note**: A linker script update is required.

core/system/inc/linker.h
```diff
@@ -19,62 +19,69 @@
...
 
-        /* boxes configuration tables */
-    uint32_t *cfgtbl_start, *cfgtbl_end;
+        /* Secure boxes configuration tables */
+    uint32_t * cfgtbl_start;
+    uint32_t * cfgtbl_end;
 
-        /* pointers to boxes configuration tables */
-    uint32_t *cfgtbl_ptr_start, *cfgtbl_ptr_end;
+        /* Pointers to the secure boxes configuration tables */
```
Patater: Do we want comments not aligned with the code still, or should we fix that in this commit, too?

api/src/uvisor-input.S
```diff
@@ -37,54 +40,49 @@ uvisor_init:
...
     .long __uvisor_mode
 
-    /* start and end address of protected bss */
+    /* Protected BSS section */
     .long __uvisor_bss_start
     .long __uvisor_bss_end
 
-    /* start and end address of uvisor main bss */
+        /* uVisor own BSS section */
     .long __uvisor_bss_main_start
```
Patater: Do we want comments to line up with the code here or not?

api/src/uvisor-input.S
```diff
@@ -37,54 +40,49 @@ uvisor_init:
...
+    /* uVisor mode of operation */
     .long __uvisor_mode
 
-    /* start and end address of protected bss */
+    /* Protected BSS section */
     .long __uvisor_bss_start
     .long __uvisor_bss_end
 
-    /* start and end address of uvisor main bss */
+        /* uVisor own BSS section */
```
Patater: Do we want comments to line up with the code here or not?

api/src/uvisor-input.S
```diff
@@ -93,14 +91,23 @@ uvisor_config:
     /* Privileged system IRQ hooks */
     .long __uvisor_priv_sys_irq_hooks
 
+/* uVisor mode of operation
+ * 0: Disabled
+ * 1: Permissive
+ * 2: Enabled */
```
Patater: What if we change these values elsewhere? We'd forget to update this comment, probably. For the sake of maintainability, consider removing the values here and reference `enum` names or `#define` names instead, if you want to reference the values that `__uvisor_mode` can be.

api/src/uvisor-input.S
```diff
@@ -93,14 +91,23 @@ uvisor_config:
...
 __uvisor_mode:
-    /* uvisor default mode - user can override weak reference */
+    /* uVisor default mode: Disabled.
+     * The user can override this weak reference. */
     .long 0
 
 .section .bss
+/* Pointer to the secure boxes context. It is update at runtime by uVisor to
```
Patater: s/update/updated

api/inc/register_gateway.h
```diff
@@ -17,105 +17,208 @@
...
-/* Register gateway metadata */
-#if defined(__CC_ARM)
-
-/* TODO/FIXME */
-
-#elif defined(__GNUC__)
+/** Get the offset of a struct member.
+ * @internal
+ */
+#define OFFSETOF(type, member) ((uint32_t) (&(((type *)(0))->member)))
```
Patater: Can we use `offsetof` instead of this macro?

core/system/inc/linker.h
AlessandroA: I removed the indentation, I think it's cleaner! Thanks for spotting it

api/src/uvisor-input.S
AlessandroA: I removed the indentation, I think it's cleaner! Thanks for spotting it

api/src/uvisor-input.S
AlessandroA: Yeah unfortunately the names would be UVISOR_ENABLED, _DISABLED, _PERMISSIVE and they're not an enum, but just preprocessor definitions. It would be equally outdated in case of change. I'd rather remove the comment altogether, what do you think?

api/inc/register_gateway.h
AlessandroA: `offsetof` is a GCC builtin and it's not available on ARMCC that I know of. Since this is user code I wanted to make it more general.

api/inc/register_gateway.h
```diff
@@ -17,105 +17,208 @@
...
+        }; \
+        \
+        /* Pointer to the register gateway we just created. The pointer is
+         * located in a discoverable linker section. */ \
+        __attribute__((section(".keep.uvisor.register_gateway_ptr"))) \
+        static uint32_t const register_gateway_ptr = (uint32_t) &register_gateway; \
+        (void) register_gateway_ptr; \
+        \
+        /* Call the actual gateway. */ \
+        uint32_t result = ((uint32_t (*)(void)) ((uint32_t) &register_gateway | 1))(); \
```
Patater: I'm fine with this implementation. File an issue for getting rid of this `| 1`. We can then prioritize the optimization appropriately. I wouldn't block this PR based on having the run-time `| 1` here, and it isn't a priority to get rid of it right now.

api/inc/register_gateway_exports.h
```diff
@@ -0,0 +1,57 @@
...
+    uint32_t mask;
+    uint16_t operation;
+    uint16_t bxlr;
+} UVISOR_PACKED TRegisterGateway;
+
+/** Register gateway operations
+ * @warning The operation value must be hardcoded. */
+#define UVISOR_RGW_OP_READ          0 /**< value = *address                      */
+#define UVISOR_RGW_OP_READ_AND      1 /**< value = *address & mask               */
+#define UVISOR_RGW_OP_WRITE         4 /**< *address = value                      */
```
Patater: Is there some undocumented scheme here behind the numbers? Why no `2` or `3`. Please comment why in the code, if this is on purpose.

api/src/uvisor-input.S
```diff
@@ -79,6 +79,10 @@ uvisor_config:
     .long __uvisor_cfgtbl_ptr_start
     .long __uvisor_cfgtbl_ptr_end
 
+        /* Pointers to the secure boxes register gateways */
```
Patater: Are we sure we don't want to align comments with code?

api/src/uvisor-input.S
Patater: I've fine having the comment use the `#define` names. Those are part of our API can shouldn't change.

I'm also fine removing the comment, but there must be some value in having the comment as you felt there was a reason to add it.

api/inc/register_gateway.h
Patater: That's unfortunate, for ARMCC.

api/inc/register_gateway.h
```diff
@@ -17,104 +17,208 @@
...
+ *                      `UVISOR_BOX_CONFIG`.
+ * @param addr[in]      The address for the data access.
+ * @param operation[in] The operation to perform at the address for the read. It
+ *                      is chosen among the `UVISOR_RGW_OP_*` macros.
+ * @param mask[in]      The mask to apply for the read operation.
+ * @returns The value read from address using the operation and mask provided
+ * (or their respective defaults if they have not been provided).
+ */
+#define uvisor_read(box_name, addr, op, msk) \
+    ({ \
```
Patater: Do all our compilers support the `({` syntax okay?

api/inc/register_gateway.h
```diff
@@ -17,104 +17,208 @@
...
+            .svc_opcode = UVISOR_SVC_OPCODE(UVISOR_SVC_ID_REGISTER_GATEWAY), \
+            .branch     = BRANCH_OPCODE(OFFSETOF(TRegisterGateway, branch), OFFSETOF(TRegisterGateway, bxlr)), \
+            .magic      = UVISOR_REGISTER_GATEWAY_MAGIC, \
+            .box_ptr    = (uint32_t) &box_name ## _cfg_ptr, \
+            .address    = (uint32_t) addr, \
+            .value      = val, \
+            .mask       = msk, \
+            .operation  = op, \
+            .bxlr       = BXLR  \
+        }; \
```
Patater: Note that you didn't need to add `__attribute__((aligned(2))` here, because you are doing `| 1` and the object might (slight chance) not be aligned to a 2 byte boundary in which case no change to the address of the object needs to change for it to be a valid thumb-2 target.

api/inc/register_gateway_exports.h
```diff
@@ -0,0 +1,57 @@
...
+    uint16_t svc_opcode;
+    uint16_t branch;
+    uint32_t magic;
+    uint32_t box_ptr;
+    uint32_t address;
+    uint32_t value;
+    uint32_t mask;
+    uint16_t operation;
+    uint16_t bxlr;
+} UVISOR_PACKED TRegisterGateway;
```
Patater: Add a comment explaining why this must be packed, including the reason that there must be no padding before `svc_opcode`.

api/inc/register_gateway.h
AlessandroA: Yup. ARMCC is not officially supported yet by uVisor but it has the same extension (it's a GNU extension).

api/inc/register_gateway_exports.h
AlessandroA: Just a typo. Fixed

api/inc/register_gateway.h
AlessandroA: Sorry I don't get the comment!

AlessandroA: Ack, now I get it. Yes, initially I was thinking about the old 4-byte alignment requirements and added an attribute to enforce it, but this is not required for the reason you mentioned.

api/inc/register_gateway_exports.h
AlessandroA: Done.

AlessandroA:
* Addressed second round of review by @Patater

api/inc/register_gateway.h
Patater: So, we aren't using `offsetof` because of ARMCC. But we are using `({`, even though ARMCC, because ARMCC is not supported. Both these uses are in api/inc headers.

api/inc/register_gateway.h
niklas-arm: Hm, the ARMv7-M ARM states "All instruction fetches must be halfword-aligned." (page 48). We are not returning from an exception here therefore:
> - bit [0] of the value is ignored when loading the PC using an ADD or MOV instruction
> - a BLX, BX, LDR to the PC, POP or LDM including the PC instruction will cause an INVSTATE UsageFault if bit [0] of the value loaded is zero
> - loading the PC with a value from a memory location whose address is not word aligned is UNPREDICTABLE.

Since we cannot be sure of the actual mechanism the compiler uses to execute that struct (probably `blx` though), I'd actually make it 4-byte aligned.

api/inc/register_gateway.h
niklas-arm: Good, I'm going to need this macro for the tier-2 allocator then ;-)

api/inc/register_gateway.h
Patater: Actually, yes. It is required still. Ignore my nonsense. The CPU would read from a memory address 0x1234566 for a jump target of 0x1234567. This obviously won't work right if the CPU instruction actually starts at 0x1234567. I'd recommend you add an explicit alignment on the `TRegisterGateway`, `__attribute__((aligned(4))`.

api/inc/register_gateway_exports.h
Patater: Force an alignment here. This struct must always be located at an executable location, which means `__attribute__((aligned(4))`.

api/inc/register_gateway.h
AlessandroA: What I meant is that ARMCC supports the GNU extensions, hence it supports `({ ... })`. It does not include the GCC intrinsics/builtins though, so it does not have `offsetof`.

api/inc/register_gateway.h
AlessandroA: Thanks, you're right @niklas-arm and @Patater ! Fixed.

api/inc/register_gateway_exports.h
AlessandroA: Done.

AlessandroA:
* Addressed third round of review by @Patater and @niklas-arm.

api/inc/register_gateway_exports.h
AlessandroA: Actually, I put it into the typedef as you suggested: It's a requirement of the gateway to be always aligned.

AlessandroA:
* Rebased after #227 has been merged.

Note: The commit list is shown out-of-order by GH, but the actual branch commit order is correct.

meriac:
@AlessandroA @niklas-arm @Patater : page_allocator.c could have a conditional .h include depending on whether the .c is compiled in a uvisor context or in unsupported. The unsupported header file does not need to be included in the supported tree and vice versa. Potentially the  header file name can be identical for bot supported and unsupported ("page_allocator_config.h") and have different content for supported/unsupported.

api/inc/register_gateway.h
```diff
@@ -17,104 +17,210 @@
...
-    ".word " UVISOR_TO_STRING(UVISOR_REGISTER_GATEWAY_MAGIC) "\n" \
-    ".word " UVISOR_TO_STRING(addr) "\n" \
-    ".word " UVISOR_TO_STRING(src_box) "_cfg_ptr\n" \
-    ".word " UVISOR_TO_STRING(UVISOR_OP_READ(UVISOR_OP_NOP)) "\n" \
-    ".word " UVISOR_TO_STRING(__UVISOR_OP_DEFAULT_MASK) "\n" \
-    "skip_args%=:\n"
+/** Get the offset of a struct member.
+ * @internal
+ */
+#define OFFSETOF(type, member) ((uint32_t) (&(((type *)(0))->member)))
```
meriac: @AlessandroA : OFFSET_OF instead ?

niklas-arm: You may have to remove this [`#define OFFSETOF`](https://github.com/ARMmbed/uvisor/blob/dev/api/rtx/src/secure_allocator.c#L33).

api/inc/register_gateway.h
```diff
@@ -17,104 +17,210 @@
...
+ * Where:
+ *     PC = <instr>_addr + 4
+ * The +4 is to account for the pipelined PC at the time of the branch
+ * instruction. See ARM DDI 0403E.b page A4-102 for more details.
+ * @param instr[in] Address of the branch instruction
+ * @param label[in] Address of the label
+ * @returns the 16-bit encoding of the B.N <label> instruction.
+ */
+#define BRANCH_OPCODE(instr, label) \
+    (uint16_t) (0xE000 | (uint8_t) ((((uint32_t) (label) - (uint32_t) (instr + 4)) / 2) & 0xFF))
```
meriac: @AlessandroA : Shouldn't that be "(uint32_t) instr + 4" ?

api/inc/register_gateway.h
AlessandroA: I used `OFFSETOF` because it's the macro implementation of the GCC built-in `offsetof`. Do you prefer the other name? It's only used here anyway.

api/inc/register_gateway.h
meriac: Please mention that issue in the comment of the pull request.

api/inc/register_gateway.h
AlessandroA: Good point! Fixed.

AlessandroA:
* Added the issue #241 to track the runtime OR operation on the gateway function pointer.

api/inc/register_gateway.h
meriac: @AlessandroA we needs to namespace it anyway for uVisor, right - UVISOR_OFFSET_OF ?

core/system/src/register_gateway.c
```diff
@@ -0,0 +1,125 @@
...
+        HALT_ERROR(SANITY_CHECK_FAILED,
+                   "The pointer (0x%08X) in the register gateway 0x%08X is not a valid box configuration pointer",
+                   register_gateway->box_ptr, (uint32_t) register_gateway);
+    }
+
+    /* Verify that the register gateway is registered in the discoverable
+     * register gateway pointers section. */
+    uint32_t * ptr = __uvisor_config.register_gateway_ptr_start;
+    bool found = false;
+    for (; ptr < __uvisor_config.register_gateway_ptr_end; ptr++) {
```
meriac: @AlessandroA : I hear unicorns crying... what about the both-way links ?

AlessandroA: Yeah I know. The circular links didn't work unfortunately.

meriac: How about we remove it then and rethink it ?

core/system/src/register_gateway.c
```diff
@@ -0,0 +1,125 @@
...
+    /* Note: We assume here that TRegisterGateway is a multiple of 32 bit. If
+     * that is not true there might be false negatives with an unaligned secure
+     * gateway. */
+    int word = 0;
+    for (; word < sizeof(TRegisterGateway) / sizeof(uint32_t); word++) {
+        if (!vmpu_public_flash_addr(((uint32_t) register_gateway) + sizeof(uint32_t) * word)) {
+            HALT_ERROR(SANITY_CHECK_FAILED, "Register gateway 0x%08X is not in public flash.",
+                       (uint32_t) register_gateway);
+        }
+    }
```
meriac: @AlessandroA: too heavy-handed - please just verify the alignment of the start and that the start is a vmpu_public_flash_addr

core/system/src/register_gateway.c
```diff
@@ -0,0 +1,127 @@
...
+        HALT_ERROR(NOT_ALLOWED, "Register gateway is owned by box %d, while the active box is %d.",
+                   box_id, g_active_box);
+    }
+
+    /* Verify that the address is in one of the allowed memory sections:
+     *   0x40000000 - 0x400FFFFF: Peripheral
+     *   0x42000000 - 0x43FFFFFF: Peripheral alias
+     *   0xE00FF000 - 0xE00FFFFF: Custom ROM Table */
+    uint32_t address = register_gateway->address;
+    if (((address & VMPU_PERIPH_MASK) != VMPU_PERIPH_START) &&
```
meriac: @Alessandro you can combine both into one rule -and ignore the 1MB hole inbetween.

core/system/src/register_gateway.c
```diff
@@ -0,0 +1,127 @@
...
+/** Perform the validation of a register gateway.
+ * @internal
+ * @warning This function does not verify that the operation, value and mask
+ * make sense together. The code that performs the operation should check for
+ * this and return an error accordingly.
+ * @param register_gateway[in]  Pointer to the register gateway structure in
+ *                              flash.
+ * @returns the extracted address of the register gateway.
+ */
+static TRegisterGateway * register_gateway_check(TRegisterGateway * register_gateway)
```
meriac: @AlessandroA Please move the HALT_ERRORS out of this function, create an approriate error.

core/system/src/register_gateway.c
```diff
@@ -0,0 +1,127 @@
...
+    uint32_t address = register_gateway->address;
+    if (((address & VMPU_PERIPH_MASK) != VMPU_PERIPH_START) &&
+        ((address & VMPU_PERIPH_BITBAND_MASK) != VMPU_PERIPH_BITBAND_START) &&
+        ((address & VMPU_ROMTABLE_MASK) != VMPU_ROMTABLE_START)) {
+        HALT_ERROR(NOT_ALLOWED, "Register gateways can only target the peripheral or ROM Table memory regions.");
+    }
+
+    /* If none of the above checks failed, the same pointer to the register
+     * gateway is returned to the caller. */
+    return register_gateway;
```
meriac: just return the error code - it's confusing to the parameter. Please make the input parameters const const.

core/system/src/register_gateway.c
```diff
@@ -0,0 +1,127 @@
...
+}
+
+/* Perform a register gateway operation. */
+void register_gateway_perform_operation(uint32_t svc_sp, uint32_t svc_pc)
+{
+    /* Check if the SVCall points to a register gateway and extract the
+     * destination address from it. */
+    /* This function halts if it finds an error. */
+    TRegisterGateway * register_gateway = register_gateway_check((TRegisterGateway *) svc_pc);
+
```
meriac: Do the halt_error here including the result - access denied is fine.

AlessandroA:
* Addressed latest comments from @meriac. Now `register_gateway_check()` returns a status code. It's the caller that halts the system with a `NOT_ALLOWED` error.

core/system/src/register_gateway.c
AlessandroA: Done.

core/system/src/register_gateway.c
AlessandroA: Done.

core/system/src/register_gateway.c
AlessandroA: Done.

core/system/src/register_gateway.c
AlessandroA: Done.

core/system/inc/mpu/vmpu.h
```diff
@@ -101,8 +101,17 @@ static UVISOR_FORCEINLINE int vmpu_sram_addr(uint32_t addr)
...
+#define VMPU_PERIPH_MASK          0xFFF00000
 #define VMPU_PERIPH_BITBAND_START 0x42000000
 #define VMPU_PERIPH_BITBAND_END   0x43FFFFFF
+#define VMPU_PERIPH_BITBAND_MASK  0xFE000000
+#define VMPU_PERIPH_FULL_MASK     0xFC000000
+
+/* ROM Table region boundaries */
+#define VMPU_ROMTABLE_START 0xE00FF000UL
+#define VMPU_ROMTABLE_END   0xE00FFFFFUL
+#define VMPU_ROMTABLE_MASK  0xFFFFF000UL
```
meriac: @AlessandroA remove unused  VMPU_PERIPH_END and VMPU_PERIPH_BITBAND_END and VMPU_ROMTABLE_END symbols ? Create the wrong impression that the end can be non-aligned.

PR #247, opened 2016-06-09
Add osEvents notifications for unsupported

niklas-arm:
This adds the osEvents notification and allocator switching code to uVisor for the unsupported case.

cc @meriac @AlessandroA @Patater


Patater:
I don't like that the CMSIS file is duplicated, but I don't have any better ideas for now. We are duplicating the "core" files from CMSIS already.

PR #248, opened 2016-06-09
Kinetis: Fix the SRAM_OFFSET configuration parameter

AlessandroA:
A previous FIXME in the code was referring to a mismatch between the
mbed 3.0 linker script configuration (0x200) and the original offset as
set by the official Freescale/NXP linker scripts.

Now that mbed features 0x400 again, we can remove the FIXME message and
use that value.

@meriac @patater @niklas-arm

### DIFF
--- a/platform/kinetis/inc/configurations.h
+++ b/platform/kinetis/inc/configurations.h
@@ -36,10 +36,7 @@
 /* Memory boundaries */
 #define FLASH_ORIGIN 0x00000000
 #define FLASH_OFFSET 0x410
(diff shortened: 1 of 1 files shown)


AlessandroA:
* Marked as "don't merge yet". I want to test all modifications (here, and in mbed) before merging.

AlessandroA:
Tested, it can now be merged.

AlessandroA:
@meriac

Patater:
LGTM

PR #250, opened 2016-06-09
Remove deprecated assertions on bss_size == 0

AlessandroA:
The assertions were used in the `vmpu_acl_stack()` functions (Kinetis and
ARM MPUs) to check that box 0 has no context.

Now the concept of context has been extended to a more general BSS, and
the BSS for a box (including box 0) has always a size greater than 0.

@meriac @Patater @niklas-arm


niklas-arm:
Do we want to add an [`assert(bss != 0)` like here](https://github.com/AlessandroA/uvisor/blob/bc089ebc38c9bb51ccb05b20074ca55d549b9547/core/system/src/mpu/vmpu_freescale_k64.c#L206)?

AlessandroA:
I did not put it there because anyway we already check (somewhere else) that the `bss_size` contains enough space for (at least) a `UvisorBoxIndex`. The original check was here because it was a way to prove that box 0 wasn't misconfigured, but now it's more complicated than just a check.

PR #252, opened 2016-06-10
Register gateway: Improve the API generality

AlessandroA:
With the new API:

* A gateway can be marked as shared. If so, there is no check on the
  current box being the same as the gateway owner.
* The value for a write operation can be runtime-defined.
* The gateway performs 8-, 16-, 32-bit operations, depending on the size
  of the value pointed by the address field of the gateway.

@meriac @Patater @niklas-arm 

**Note**:

The new API looks like this:

```C
uint16_t value = uvisor_read(my_box, &PERIPH->REG, UVISOR_RGW_OP_READ, true, 0xFFFFFFFF);
```

The `true` means that the gateway is shared. This is how generally we will set all gateways in the legacy code (SDKs, OSs, etc.) for shared boxes to use their low-level APIs.

### DIFF
--- a/core/system/src/register_gateway.c
+++ b/core/system/src/register_gateway.c
@@ -56,7 +56,7 @@ static int register_gateway_check(TRegisterGateway const * const register_gatewa
     }
 
     /* Verify that the box configuration pointer is in the dedicated flash
-     * region and that it corresponds to the currently active box. */
+     * region. */
     /* Note: We only check that the box pointer is larger than the start of the
      * pointer region. The subsequent substraction that we do to calculate the
      * box ID is then guaranteed to be sane. */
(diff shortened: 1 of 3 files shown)


core/system/src/register_gateway.c
```diff
@@ -100,38 +104,74 @@ void register_gateway_perform_operation(uint32_t svc_sp, uint32_t svc_pc)
...
-    vmpu_unpriv_uint32_write(svc_sp, result);
+    /* Store the result at the target address.
+     * The code runs here only if the register gateway performs a write
+     * operation. */
+    switch(width) {
+    case 4:
+        *((uint32_t *) address) = (uint32_t) result;
+        break;
+    case 2:
+        *((uint16_t *) address) = (uint16_t) (result & 0xFFFF);
```
meriac: @AlessandroA the masking seems unnecessary ?

Patater:
LGTM

PR #253, opened 2016-06-10
Register gateway: Provide macros for the shared bit

AlessandroA:
@meriac @patater @niklas-arm 

The API would now look like this:

```C
uint16_t val = uvisor_read(my_box, UVISOR_RGW_EXCLUSIVE, address, UVISOR_RGW_OP_READ, 0xFFFFFFFFUL);
```

### DIFF
--- a/api/inc/register_gateway.h
+++ b/api/inc/register_gateway.h
@@ -60,16 +60,17 @@
  *
  * @param box_name[in]  The name of the source box as decalred in
  *                      `UVISOR_BOX_CONFIG`.
+ * @param shared[in]    Whether the gateway can be shared with other boxes or
+ *                      not. Two values are available: UVISOR_RGW_SHARED,
+ *                      UVISOR_RGW_EXCLUSIVE.
  * @param addr[in]      The address for the data access.
  * @param operation[in] The operation to perform at the address for the read. It
  *                      is chosen among the `UVISOR_RGW_OP_*` macros.
- * @param shared[in]    True if the gateway can be performed by any box. In this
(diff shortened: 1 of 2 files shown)


Patater:
LGTM

api/inc/register_gateway.h
```diff
@@ -143,77 +145,95 @@
...
+ *                     not. Two values are available: UVISOR_RGW_SHARED,
+ *                     UVISOR_RGW_EXCLUSIVE.
  * @param address[in]  Address at which to check the bits
  * @param mask[in]     Bits to select out of the target address
  * @returns The value `(bool) (*address & mask) == mask)`.
  */
-#define UVISOR_BITS_CHECK(box_name, address, mask) \
-    ((bool) (UVISOR_BITS_GET(box_name, address, mask) == mask))
+#define UVISOR_BITS_CHECK(box_name, shared, address, mask) \
+    ((bool) (UVISOR_BITS_GET(box_name, shared, address, mask) == mask))
```
niklas-arm: Is `bool` acceptable here? The user would have to have `#include <stdbool.h>` somewhere, but we cannot guarantee this, unless we include it here as well.

niklas-arm:
👍  for moving the shared bit forward.

niklas-arm:
Apart from `bool` usage, LTGM.

api/inc/register_gateway.h
AlessandroA: Good point! I removed the `bool`. The macro can still effectively be used in conditional statements.

AlessandroA:
* Addressed @niklas-arm's comments.

PR #256, opened 2016-06-14
fixup! export-table: Add export table

Patater:
The original implementation of the export table had a bug where, only in
release builds, the export table was optimized out by the compiler. Make
the export table non-static so that the compiler will not optimize out the
export table in release builds.

Fixes: 7408972eddd3 ("export-table: Add export table")

### DIFF
--- a/core/system/src/export_table.c
+++ b/core/system/src/export_table.c
@@ -20,7 +20,7 @@
  * table can be exported correctly. Placing this table into the .export_table
  * section locates this table at the end of the uVisor binary. */
 __attribute__((section(".export_table")))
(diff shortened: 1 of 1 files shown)


AlessandroA:
Do we want to keep the `fixup!` tag in the commit message? Other than that, LGTM.

Tested with K64F_SECURE and the CMSIS-DAP firmware.

Patater:
This is targeted at the dev branch and fixes a bug introduced only to the dev branch, so it makes sense to leave the fixup for now, and apply the fixup before merging from dev to master.

PR #258, opened 2016-06-15
Add thread event implementations for osEvents

salkinium:
Required to get correct allocator switching in supported mode.
Note: I had to increase the flash size, so I cherry-picked Jaedens commit directly.

This PR will be closed in favour of https://github.com/ARMmbed/uvisor/pull/259


Patater:
LGTM

api/src/uvisor-lib.c
```diff
@@ -37,5 +38,7 @@ int uvisor_lib_init(void)
         return -1;
     }
 
+    osRegisterForOsEvents(&uvisor_export_table->os_event_observer);
```
meriac: Please add a comment that osEventObs won't allow a second call. For systems that don't make use of osRegisterForOsEvents we recommend to osRegisterForOsEvents(NULL) to disable further registrations (backdoor). Please make sure to submit a patch to osRegisterForOsEvents that implements and documents that. @AlessandroA @Patater @niklas-arm

core/system/src/export_table.c
```diff
@@ -14,7 +14,105 @@
...
+    uint32_t ii = 0;
+    for (; ii < UVISOR_EXPORT_TABLE_THREADS_MAX_COUNT; ii++) {
+        if (thread[ii].allocator == NULL)
+            break;
+    }
+    if (ii < UVISOR_EXPORT_TABLE_THREADS_MAX_COUNT) {
+        /* Remember the process id for this thread. */
+        thread[ii].process_id = g_active_box;
+        /* Fall back to the process heap if ctx is NULL. */
+        thread[ii].allocator = context ? context : index->box_heap;
```
meriac: @niklas-arm @AlessandroA that looks insecure: using the user-controllable box context pointer in global memory instead of using the one from the internal table.

meriac: Maybe better covered by another privcall to uvisor ? @AlessandroA @niklas-arm

core/system/src/export_table.c
```diff
@@ -14,7 +14,105 @@
...
+    /* If the thread is inside another process, switch into it. */
+    if (context->process_id != g_active_box) {
+        context_switch_in(CONTEXT_SWITCH_UNBOUND_THREAD, context->process_id, 0, 0);
+    }
+    /* Copy the thread allocator into the (new) box index. */
+    index = (UvisorBoxIndex *) *(__uvisor_config.uvisor_box_context);
+    if (context->allocator) {
+        /* If the active_heap is NULL, then the process heap needs to be
+         * initialized yet. The initializer sets the active heap itself. */
+        if (index->active_heap) {
```
meriac: @niklas-arm - if (!index->active_heap) ?

core/uvisor-config.h
```diff
@@ -17,7 +17,7 @@
 #ifndef  __UVISOR_CONFIG_H__
 
 #define UVISOR_MAGIC        0x2FE539A6
-#define UVISOR_FLASH_LENGTH 0x8000
+#define UVISOR_FLASH_LENGTH 0x8800
```
meriac: 0x9000 ? Should be uncritical as its kindof UVISOR_FLASH_MAX and wont waste space when not used.

core/system/src/export_table.c
niklas-arm: nope.

core/system/src/export_table.c
niklas-arm: But `context` is passed by the user anyway.

meriac: context is not the problem here as it's only passed on. If I correctly understand we deal with privileged code here. Thus we can't use the variable in this way.

meriac: We should reconsider and save vital parameters like that into protected memories - even if that means duplication @niklas-arm @AlessandroA

core/system/src/export_table.c
meriac: I see what you are doing. I don't like very much overloading of functionality into a single variable. That smells like trouble when extending code in future. How about an ->initialized flag instead ?

core/system/src/export_table.c
niklas-arm: Nothing stops you from passing `index->box_heap` as `context` in `osThreadCreateContext`.
The entire box index is only protected against access from other processes, not it's own.

meriac:
Closed in favour of https://github.com/ARMmbed/uvisor/pull/260

PR #262, opened 2016-06-20
Switch allocators for uVisor supported; Add uVisor box main feature

Patater:
I've cleaned up the commits we have in unstable.

These PRs were similar to this one, but not as clean or as far along in implementing previous feedback.
https://github.com/ARMmbed/uvisor/pull/260
https://github.com/ARMmbed/uvisor/pull/258

The following is an excerpt from the example-uvisor-threaded-blinky application, illustrating the use of the new box main feature. For the complete working example, see https://github.com/Patater/example-uvisor-threaded-blinky
```C++
static void led1_main(const void *);

UVISOR_BOX_NAMESPACE(NULL);
UVISOR_BOX_HEAPSIZE(8192);
UVISOR_BOX_MAIN(led1_main, osPriorityNormal, UVISOR_BOX_STACK_SIZE);
UVISOR_BOX_CONFIG(box_led1, acl, UVISOR_BOX_STACK_SIZE, box_context);
```

After this lands in the uVisor dev branch, meriac/mbed-os needs to commit the results of a fresh run of the importer script. Then, https://github.com/ARMmbed/example-uvisor-threaded-blinky/pull/2 needs to be updated to point to this new version of meriac/mbed-os.

### DIFF
--- a/api/inc/svc_exports.h
+++ b/api/inc/svc_exports.h
@@ -114,8 +114,10 @@
 #define UVISOR_SVC_ID_PAGE_FREE             UVISOR_SVC_CUSTOM_TABLE(23)
 
 /* SVC immediate values for hardcoded table (call from unprivileged) */
-#define UVISOR_SVC_ID_UNVIC_OUT        UVISOR_SVC_FIXED_TABLE(0, 0)
-#define UVISOR_SVC_ID_REGISTER_GATEWAY UVISOR_SVC_FIXED_TABLE(3, 0)
+#define UVISOR_SVC_ID_UNVIC_OUT           UVISOR_SVC_FIXED_TABLE(0, 0)
+/* Deprecated: UVISOR_SVC_ID_CX_IN(nargs) UVISOR_SVC_FIXED_TABLE(1, nargs) */
+/* Deprecated: UVISOR_SVC_ID_CX_OUT       UVISOR_SVC_FIXED_TABLE(2, 0) */
+#define UVISOR_SVC_ID_REGISTER_GATEWAY    UVISOR_SVC_FIXED_TABLE(3, 0)
 
 /* SVC immediate values for hardcoded table (call from privileged) */
 #define UVISOR_SVC_ID_UNVIC_IN         UVISOR_SVC_FIXED_TABLE(0, 0)
--- a/api/inc/svc_exports.h
+++ b/api/inc/svc_exports.h
@@ -118,6 +118,8 @@
 /* Deprecated: UVISOR_SVC_ID_CX_IN(nargs) UVISOR_SVC_FIXED_TABLE(1, nargs) */
 /* Deprecated: UVISOR_SVC_ID_CX_OUT       UVISOR_SVC_FIXED_TABLE(2, 0) */
 #define UVISOR_SVC_ID_REGISTER_GATEWAY    UVISOR_SVC_FIXED_TABLE(3, 0)
+#define UVISOR_SVC_ID_BOX_INIT_FIRST      UVISOR_SVC_FIXED_TABLE(4, 0)
+#define UVISOR_SVC_ID_BOX_INIT_NEXT       UVISOR_SVC_FIXED_TABLE(5, 0)
 
 /* SVC immediate values for hardcoded table (call from privileged) */
 #define UVISOR_SVC_ID_UNVIC_IN         UVISOR_SVC_FIXED_TABLE(0, 0)
--- a/api/inc/vmpu_exports.h
+++ b/api/inc/vmpu_exports.h
@@ -169,6 +169,10 @@ typedef struct {
     /* Contains user provided size of box heap without guards of buffers. */
     uint32_t heap_size;
 
+    /* Opaque-to-uVisor data that potentially contains uvisor-lib-specific or
+     * OS-specific per-box configuration */
+    const void * const lib_config;
+
     const char * box_namespace;
     const UvisorBoxAclItem * const acl_list;
     uint32_t acl_count;
--- a/api/rtx/src/box_init.c
+++ b/api/rtx/src/box_init.c
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2016, ARM Limited, All Rights Reserved
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "uvisor-lib/uvisor-lib.h"
+#include "mbed_interface.h"
+#include "cmsis_os.h"
+#include <stdint.h>
+#include <string.h>
+
+/* This function is called by uVisor in unprivileged mode. On this OS, we
+ * create box main threads for the box. */
+void __uvisor_lib_box_init(void * lib_config)
+{
+    osThreadId thread_id;
+    osThreadDef_t * flash_thread_def = lib_config;
+    osThreadDef_t thread_def;
+
+    /* Copy thread definition from flash to RAM. The thread definition is most
+     * likely in flash, so we need to copy it to box-local RAM before we can
+     * modify it. */
+    memcpy(&thread_def, flash_thread_def, sizeof(thread_def));
+
+    /* Note that the box main thread stack is separate from the box stack. This
+     * is because the thread must be created to use a different stack than the
+     * stack osCreateThread() is called from, as context information is saved
+     * to the thread stack by the call to osCreateThread(). */
+    /* Allocate memory for the main thread from the process heap (which is
+     * private to the process). This memory is never freed, even if the box's
+     * main thread exits. */
+    thread_def.stack_pointer = malloc_p(thread_def.stacksize);
+
+    if (thread_def.stack_pointer == NULL) {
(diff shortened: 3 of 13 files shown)


core/system/src/box_main.c
```diff
@@ -0,0 +1,166 @@
...
+    /* If the currently active box is not the one we expected from our running
+     * counter, then the recursive chain of context switches has been broken.
+     * */
+    if (g_active_box != g_box_main_counter) {
+        HALT_ERROR(NOT_ALLOWED, "The chain of box initializations has been broken.");
+    }
+
+    /* Get the handler from the table of exported handlers. */
+    /* Note: We assume this table has already been verified. */
+    uint32_t const dst_fn = (uint32_t) __uvisor_config.box_main_handler;
```
niklas-arm: Shouldn't this also check if the function address is in flash?

Patater: It should be in flash or box-local SRAM.

core/system/src/export_table.c
```diff
@@ -14,7 +14,105 @@
...
+static void * thread_create(int id, void * c)
+{
+    (void) id;
+    UvisorThreadContext * context = c;
+    const UvisorBoxIndex * const index =
+            (UvisorBoxIndex * const) *(__uvisor_config.uvisor_box_context);
+    /* Search for a free slot in the tasks meta data. */
+    uint32_t ii = 0;
+    for (; ii < UVISOR_EXPORT_TABLE_THREADS_MAX_COUNT; ii++) {
+        if (thread[ii].allocator == NULL)
```
AlessandroA: Add missing braces.

Patater: Done

core/system/src/export_table.c
```diff
@@ -14,7 +14,105 @@
...
+            "thread context (%08x) is invalid!\n",
+            context);
+    }
+}
+
+static void thread_switch(void * c)
+{
+    UvisorThreadContext * context = c;
+    UvisorBoxIndex * index;
+    if (context == NULL) return;
```
AlessandroA: Add missing braces. `return` goes in its own line.

core/system/src/export_table.c
```diff
@@ -14,7 +14,105 @@
...
+    if (context->process_id != g_active_box) {
+        context_switch_in(CONTEXT_SWITCH_UNBOUND_THREAD, context->process_id, 0, 0);
+    }
+    /* Copy the thread allocator into the (new) box index. */
+    index = (UvisorBoxIndex *) *(__uvisor_config.uvisor_box_context);
+    if (context->allocator) {
+        /* If the active_heap is NULL, then the process heap needs to be
+         * initialized yet. The initializer sets the active heap itself. */
+        if (index->active_heap) {
+            index->active_heap = context->allocator;
```
AlessandroA: Do we trust `context->allocator`? Who sets it?

core/system/src/export_table.c
Patater: Done

core/system/src/export_table.c
niklas-arm: We don't need to trust it.
It is passed by the user in `thread_create` and stored in the `thread` array.
uVisor does not care what it is, it never looks at it.

core/system/src/svc.c
```diff
@@ -137,7 +138,7 @@ void UVISOR_NAKED SVCall_IRQn_Handler(void)
         ".word  __svc_not_implemented\n"
         ".word  register_gateway_perform_operation\n"
         ".word  __svc_not_implemented\n"
-        ".word  __svc_not_implemented\n"
+        ".word  box_init_next\n"
```
AlessandroA: Why is this the 5th and not the 4th?

api/inc/svc_exports.h
```diff
@@ -116,6 +116,7 @@
 /* SVC immediate values for hardcoded table (call from unprivileged) */
 #define UVISOR_SVC_ID_UNVIC_OUT        UVISOR_SVC_FIXED_TABLE(0, 0)
 #define UVISOR_SVC_ID_REGISTER_GATEWAY UVISOR_SVC_FIXED_TABLE(3, 0)
+#define UVISOR_SVC_ID_BOX_INIT_NEXT    UVISOR_SVC_FIXED_TABLE(5, 0)
```
AlessandroA: See [here](https://github.com/ARMmbed/uvisor/pull/262/commits/0b723065cf8c398700fb841655684f9f3608847b#r68368139).

Patater: This was done to start a discussion, but it looks like the XXX was missing.

What were 1 and 2? Those were skipped, so maybe I should skip 4 for similar reasons.

AlessandroA: When we removed the secure gateway handlers we decided not to shift the existing ones, but to leave the unused and then continue with the available slots. We will reuse 1 and 2 when we wrap. The next one to use is 4 (has never been used before).

Patater: I would like to add a comment in the code about the previously used slots to avoid any confusion in the future. What was 1 used for? What was 2 used for (different from 1)?

AlessandroA: Secure gateway handlers. See [here](https://github.com/Patater/uvisor/commit/3f4e9f03a1322048c1216538cddf40be1985f419#diff-af8f010563b390003c76aa7d9829538aR133). Yup a comment can do, just a line for every unused slot, something like:

```C
#define UVISOR_SVC_ID_UNVIC_OUT        UVISOR_SVC_FIXED_TABLE(0, 0)
/* Deprecated: UVISOR_SVC_ID_CX_IN     UVISOR_SVC_FIXED_TABLE(1, x) */
/* Deprecated: UVISOR_SVC_ID_CX_OUT    UVISOR_SVC_FIXED_TABLE(2, x) */
#define UVISOR_SVC_ID_REGISTER_GATEWAY UVISOR_SVC_FIXED_TABLE(3, 0)
```

and 

```C
".word  unvic_gateway_out\n"
/* Deprecated:  secure_gateway_in */
/* Deprecated:  secure_gateway_out */
".word  vmpu_register_gateway\n"
```

api/inc/vmpu_exports.h
```diff
@@ -158,6 +158,10 @@ typedef struct {
     /* Contains user provided size of box heap without guards of buffers. */
     uint32_t heap_size;
 
+    /* Opaque-to-uVisor data that potentially contains uvisorlib-specific
+     * or OS-specific per-box configuration */
```
AlessandroA: `uvisor-lib-specific` and `configuration.`

api/src/uvisor-input.S
```diff
@@ -106,6 +106,9 @@ uvisor_config:
     /* Privileged system IRQ hooks */
     .long __uvisor_priv_sys_irq_hooks
 
+    /* uVisor Lib Functions for use by uVisor in unprivileged mode */
```
AlessandroA: This comment is a bit misleading, shouldn't it say something like "Default box initialization handler"? It's the system-wide function that is run by every box at init time.

core/system/inc/linker.h
```diff
@@ -95,6 +95,9 @@ typedef struct {
 
     /* Privileged system IRQ hooks */
     UvisorPrivSystemIRQHooks const * const priv_sys_irq_hooks;
+
+    /* uVisor Lib Functions for use by uVisor in unprivileged mode */
+    void (*lib_box_init)(void * lib_config);
```
AlessandroA: Same as [here](https://github.com/ARMmbed/uvisor/pull/262/commits/45bf15bd72c6e0df55df0e8200d1fd46a09cecf7#r68381186)

api/rtx/src/box_init.c
```diff
@@ -0,0 +1,63 @@
...
+     * box-local RAM before we can modify it. Allocate some RAM for storing the
+     * main thread's thread definition. This memory is never freed. */
+    thread_def = malloc_p(sizeof(*thread_def));
+    if (thread_def == NULL) {
+        /* No process heap memory available */
+        mbed_die();
+    }
+
+    /* Copy thread definition from flash to RAM. */
+    memcpy(thread_def, flash_thread_def, sizeof(*thread_def));
```
AlessandroA: Is the `thread_def` changed at runtime by the OS? Or is it used as-is?

api/src/uvisor-input.S
Patater: The comment is for this entire section, not just the one function. This function is called by uVisor in unprivileged mode; this function is provided by uVisor Lib; I'm not sure what is misleading.

api/rtx/src/box_init.c
Patater: Below the code you quoted, you can see that uVisor Lib modifies the provided `thread_def` before giving it to the RTOS (by allocating a stack from the tier-2 allocator). The RTOS uses the `thread_def` after our modifications. The RTX implementation of CMSIS RTOS uses the `thread_def` to call `rt_tsk_create`, which populates a separate in-ram TCB with various options.

AlessandroA:
LGTM.

@meriac Ready for review. Note: This PR is against `dev`!

api/rtx/src/box_init.c
```diff
@@ -0,0 +1,63 @@
...
+    if (thread_def->stack_pointer == NULL) {
+        /* No process heap memory available */
+        mbed_die();
+    }
+
+    thread_id = osThreadCreate(thread_def, NULL);
+
+    if (thread_id == NULL) {
+        /* Failed to create thread */
+        mbed_die();
```
meriac: @Patater @AlessandroA : Please replace malloc_p by using a stack allocation (that is freed when leaving the function, allocas are avoided).

Patater: I verified that the RTX doesn't use the thread definition after the call to `osThreadCreate`.

However, this isn't a guarantee at the CMSIS RTOS level, so far as I can see; as in, there is nothing stopping some other CMSIS-RTOS-compliant OS from using the thread definition after the call to `osThreadCreate`.

Patater:
We now have two SVCs for box-init. One to initiate the recursion, and one for the recursion.

PR #264, opened 2016-06-20
Swap the revision and variant fields of CPUID

AlessandroA:
When verifying the core version against the minimum requirements, the
variant and revision fields in the CPUID register where swapped.

If a core version if r2p0, then the variant is r2, the revision is p0.

Reported and fixed by @fvincenzo.

@meriac @Patater @niklas-arm


AlessandroA:
@Patater Added commit to fix the typo.

Patater:
LGTM

PR #265, opened 2016-06-23
Build artifacts end up in separate folders

AlessandroA:
This PR makes sure that build artifacts end up in separate folders from the source files paths.

We have 2 classes of build artifacts:

* For the uVisor core build. Artifacts now end up in `$(CONFIGURATION)/$(BUILD_MODE)/core/`.
* For the uVisor APIs build. Artifacts now end up in `$(CONFIGURATION)/$(BUILD_MODE)/api/`.

The source files in the `core` and `api` folders can still have the same name, without generating collisions (an example is `benchmark.c` in both `api/src/` and `core/system/src/`).

As a bonus, the core and APIs builds are now separated, so in the future one can be changed without affecting the other.

@meriac @Patater @niklas-arm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ CONFIGURATION_PREFIX:=$(PLATFORM_DIR)/$(PLATFORM)/$(BUILD_MODE)/$(CONFIGURATION_
 # Release library definitions
 API_ASM_HEADER:=$(API_DIR)/src/uvisor-header.S
 API_ASM_INPUT:=$(API_DIR)/src/uvisor-input.S
-API_ASM_OUTPUT:=$(API_ASM_INPUT).$(CONFIGURATION_LOWER).$(BUILD_MODE).s
+API_ASM_OUTPUT:=$(CONFIGURATION_PREFIX)/$(API_DIR)/uvisor-output.s
 API_RELEASE:=$(API_DIR)/lib/$(PLATFORM)/$(BUILD_MODE)/$(CONFIGURATION_LOWER).a
 API_VERSION:=$(API_DIR)/lib/$(PLATFORM)/$(BUILD_MODE)/$(CONFIGURATION_LOWER).txt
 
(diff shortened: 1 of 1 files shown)


Patater:
It appears that some non-comment changes made it into your "Makefile: Improve comments" commit on accident.

AlessandroA:
You mean [this one](https://github.com/ARMmbed/uvisor/pull/265/commits/ffe82ec97d84352db94718f7429a2ed0a8fa6b38#diff-b67911656ef5d18c4ae36cb6741b7965R228)? It's just a renaming made for clarity, doesn't really change functionality. If you want I can move it to the next commit.

Patater:
It wouldn't belong in the next commit, either. I suppose if you renamed your first commit to "Makefile: Improve clarity", then it'd be okay. But, It still sort of hides the change from `make release` to `make api_release`, which is best happening all in its own commit; being separate makes the change obvious.

Makefile
```diff
@@ -224,55 +247,68 @@ endif
...
 
-# Pre-process and compile an S file into an object file.
-%.S.$(CONFIGURATION_LOWER).$(BUILD_MODE).o: %.S.$(CONFIGURATION_LOWER).$(BUILD_MODE).s
+# Pre-process and compile a release library C file into an object file.
+$(CONFIGURATION_PREFIX)/$(API_DIR)/%.o: %.c
 	$(CC) $(CPPFLAGS) $(CFLAGS) -c $< -o $@
 
+# Generate the output asm file from the asm input file and the INCBIN'ed binary.
+$(API_ASM_OUTPUT:.s=.o): $(CONFIGURATION_PREFIX).bin $(API_ASM_INPUT)
+	cp -f $(API_ASM_HEADER) $(API_ASM_OUTPUT)
```
meriac: @AlessandroA why would you use `cp -f` - isn't the failure of copy an rightful indication that something is seriously wrong ?

meriac: @AlessandroA : ping - please clarify

AlessandroA: @meriac Fixed! Now it's only `cp`, the `-f` was not needed. If the file is absent we get an error, which is better.

PR #271, opened 2016-06-28
fixup! Add tier-2 allocator init and wrapper functions

Patater:
Previously, we weren't acquiring the memory mutex if the box_heap was
the active heap. This allowed multiple threads to clobber each other's
memories when the active heap was the box_heap.


AlessandroA:
LGTM

niklas-arm:
👍

PR #272, opened 2016-06-29
Docs: Add the quick-start guide

AlessandroA:
This new document is a step-by-step guide to:

* Enable uVisor on an existing app.
* Add a secure box to the app, and:
    * Protect static and dynamic memories.
    * Gain exclusive ownership of peripherals and interrupts.

All the snippets in the guide have been tested with the latest mbed version (using the CLI tools).

Missing:

* Instructions on setting an API entry point (using RPC semantics).

@meriac @Patater @niklas-arm

### DIFF
--- a/docs/QUICKSTART.md
+++ b/docs/QUICKSTART.md
@@ -1,0 +1,413 @@
+# Quick-Start Guide for uVisor on mbed OS
+
+This guide will help you get started with uVisor on mbed OS by walking you through creating a sample application for the NXP FRDM-K64F board.
+
+The uVisor provides sandboxed environments and resources protection for applications built for ARM Cortex-M3 and Cortex-M4 devices. Here we will show you how to enable the uVisor and configure a secure box to get hold of some exclusive resources (memory, peripherals, interrupts). For more information on the uVisor design philosophy, please check out our the uVisor [introductory document](../README.md).
+
+## Overview
+
+To get a basic `blinky` application running on mbed OS with uVisor enabled, you will need the following:
+
+* A platform and a toolchain supported by uVisor on mbed OS. You can verify this on [the official list](../README.md#supported-platforms). Please note that uVisor might support some platform internally, but not on mbed OS. Generally this means that the porting process has only been partially completed. If you want to port your platform to uVisor and enable it on mbed OS, please follow the [uVisor porting guide](PORTING.md).
+* git. It will be used to download the mbed codebase.
+* The mbed command-line tools, mbed-cli. You can run `pip install mbed-cli` to install them.
+
+For the remainder of this guide we will assume the following:
+
+* You are developing on a \*nix machine, in the `~/code` folder.
+* You are building the app for the [NXP FRDM-K64F](http://developer.mbed.org/platforms/FRDM-K64F/) target, with the [GCC ARM Embedded](https://launchpad.net/gcc-arm-embedded) toolchain.
+
+The instructions provided can be easily generalized to the case of other targets on other host OSs.
+
+## Start with the `blinky` app
+[Go to top](#overview)
+
+To create a new mbed application called `uvisor-example` just run the following commands:
+
+```bash
+$ cd ~/code
+$ mbed new uvisor-example
+```
+
+The mbed-cli tools will automatically fetch the mbed codebase for you. By default, git will be used to track your code changes, so your application will be ready to be pushed to a git server, if you want to.
+
+Once the import process is finished, create a `source` folder:
+```bash
+mkdir ~/code/uvisor-example/source
+```
+and place a new file `main.cpp` in it:
+
+```C
+/* ~/code/uvisor-example/source/main.cpp */
+
+#include "mbed.h"
+#include "rtos.h"
+
+DigitalOut led(LED1);
+
+int main(void)
+{
+    while (true) {
+        led = !led;
+        Thread::wait(500);
+    }
+}
+```
+
+This simple application just blinks an LED from the main thread, which is created by default by the OS.
+
+---
+
+**Checkpoint**
+
+Compile the application:
+
+```bash
+$ mbed compile -m K64F -t GCC_ARM
+```
+
+The resulting binary will be located at:
+
+```bash
+~/code/uvisor-example/.build/K64F/GCC_ARM/uvisor-example.bin
+```
+
+Drag-and-drop it onto the USB device mounted on your computer in order to flash the device. When the flashing process is completed, press the reset button on the device. You should see the device LED blinking.
+
+---
+
+In the next sections you will see:
+
+* How to [enable uVisor](#enable-uvisor) on the `uvisor-example` app.
+* How to [add a secure box](#add-a-secure-box) to the `uvisor-example` app with exclusive access to a timer, to a push-button interrupt, and to static and dynamic memories.
+
+## Enable uVisor
+[Go to top](#overview)
+
+To enable the uVisor on the app, just add the following lines at the beginning of the `main.cpp` file:
+
+```C
+/* ~/code/uvisor-example/source/main.cpp */
+
+#include "mbed.h"
+#include "rtos.h"
+#include "uvisor-lib/uvisor-lib.h"
+
+/* Register privleged system IRQ hooks.
+ * This is a system-wide configuration and it is independent from the app, but
+ * for the moment it needs to be specified in the app. This will change in a
+ * later version: The configuration will be provided by the OS. */
+extern "C" void SVC_Handler(void);
+extern "C" void PendSV_Handler(void);
+extern "C" void SysTick_Handler(void);
+UVISOR_SET_PRIV_SYS_IRQ_HOOKS(SVC_Handler, PendSV_Handler, SysTick_Handler);
+
+/* Main box Access Control Lists (ACLs). */
+/* Note: These are specific to the NXP FRDM-K64F board. See the section below
+ *       for more information. */
+static const UvisorBoxAclItem g_main_box_acls[] = {
+    /* For the LED */
+    {SIM,   sizeof(*SIM),   UVISOR_TACLDEF_PERIPH},
+    {PORTB, sizeof(*PORTB), UVISOR_TACLDEF_PERIPH},
+
+    /* For messages printed on the serial port. */
+    {OSC,   sizeof(*OSC),   UVISOR_TACLDEF_PERIPH},
+    {MCG,   sizeof(*MCG),   UVISOR_TACLDEF_PERIPH},
+    {UART0, sizeof(*UART0), UVISOR_TACLDEF_PERIPH},
+};
+
+/* Enable uVisor, using the ACLs we just created. */
+UVISOR_SET_MODE_ACL(UVISOR_ENABLED, g_main_box_acls);
+
+/* Rest of the existing app code */
+...
+```
+
+In the code above we specified 3 elements:
+
+1. System-wide uVisor configurations: `UVISOR_SET_PRIV_SYS_IRQ_HOOKS`. Application authors currently need to specify the privileged system IRQ hooks at the application level with this macro, but in the future the operating system will register the privileged system IRQ hooks on its own.
+1. Main box Access Control Lists (ACLs). Since with uVisor enabled everything runs in unprivileged mode, we need to make sure that peripherals that are accessed by the OS and the main box are allowed. These peripherals are specified using a list like the one in the snippet above. For the purpose of this example we provide you the list of all the ACLs that we know you will need. For other platforms or other applications you need to determine those ACLs following a process that is described in a [section](#the-main-box-acls) below.
+1. App-specific uVisor configurations: `UVISOR_SET_MODE_ACL`. This macro sets the uVisor mode (enabled) and associates the list of ACLs we just created with the main box.
+
+Before compiling, we need to add a custom target that inherits all the features of the original `K64F` target, and enables the uVisor feature. To do so, add the file `~/code/uvisor-example/mbed_app.json` with the following content:
+
+```JSON
+{
+    "custom_targets": {
+        "K64F_SECURE": {
+            "inherits": ["K64F"],
+            "extra_labels_add":["K64F", "UVISOR_SUPPORTED"],
+            "features_add": ["UVISOR"]
+        }
+    }
+}
+```
+
+---
+
+**Checkpoint**
+
+Compile the application again, but this time targeting `K64F_SECURE`:
+
+```bash
+$ mbed compile -m K64F_SECURE -t GCC_ARM
+```
+
+The binary will be located at:
+
+```bash
+~/code/uvisor-example/.build/K64F_SECURE/GCC_ARM/uvisor-example.bin
+```
+
+Re-flash the device and press the reset button. The device LED should be blinking as in the previous case.
+
+---
+
+If you enable uVisor in the `blinky` app as it was written above, you will not get any particular security feature. All code and resources share the same security context, which we call the *main box*.
+
+A lot happens under the hood, though. All the user code now runs in unprivileged mode, and the systems services like the `NVIC` APIs or the OS SVCalls are routed through the uVisor.
+
+## Add a secure box
+[Go to top](#overview)
+
+Now that uVisor is enabled, we can finally add a *secure box*.
+
+A secure box is a special compartment that is granted exclusive access to peripherals, memories and interrupts. Private resources are only accessible when the *context* of the secure box is active. The uVisor is the only one that can enable a secure box context, for example upon thread switching or interrupt handling.
+
+Code that belongs to a box is not obfuscated by uVisor, so it is still readable and executable from outside of the box. In addition, declaring an object in the same file that configures a secure box does not protect that object automatically.
+
+Instead, we provide specific APIs to instruct the uVisor to protect a private resource. Here we will show how to use these APIs in the `uvisor-example` app.
+
+### Configure the secure box
+
+For this example, we want to create a secure box called `private_timer`. The `private_timer` box will be configured to have exclusive access to the PIT timer and to the GPIO PORT C on the NXP FRDM-K64F board, which means that other boxes will be prevented from accessing these peripherals.
+
+Each secure box must have at least one thread, which we call the box's main thread. In our `private_timer` box we will only use this thread throughout the whole program. The thread will constantly save the current timer value in a private buffer. In addition, we want to print the content of the buffered timer values whenever we press the `SW2` button on the board.
+
+We want the box to have exclusive access to the following resources:
+
+* The timer and push-button peripherals (as specified by a peripheral ACL). Nobody else should be able to read the timer values.
+* The push-button interrupt (as specified by an IRQ ACL). We want the button IRQ to be re-routed to our box-specific ISR.
+* The buffer that holds the timer samples (as specified by a dynamic memory ACL).
+* The static memory that holds information about the timer buffer (as specified by a static memory ACL).
+
+Create a new source file, `~/code/uvisor-example/source/secure_box.cpp`. We will configure the secure box inside this file. The secure box name for this example is `private_timer`.
+
+```C
+/* ~/code/uvisor-example/source/secure_box.cpp */
+
+#include "mbed.h"
+#include "rtos.h"
+#include "uvisor-lib/uvisor-lib.h"
+
+/* Private static memory for the secure box */
+typedef struct {
+    uint32_t * buffer;
+    int index;
+} PrivateTimerStaticMemory;
+
+/* ACLs list for the secure box: Timer (PIT). */
+static const UvisorBoxAclItem g_private_timer_acls[] = {
+    {PIT,   sizeof(*PIT),   UVISOR_TACLDEF_PERIPH}
+    {PORTC, sizeof(*PORTC), UVISOR_TACLDEF_PERIPH},
+};
+
+static void private_timer_main_thread(const void *);
+
+/* Secure box configuration */
+UVISOR_BOX_NAMESPACE(NULL);                   /* We won't specify a box namespace for this example. */
+UVISOR_BOX_HEAPSIZE(4096);                    /* Heap size for the secure box */
+UVISOR_BOX_MAIN(private_timer_main_thread,    /* Main thread for the secure box */
+                osPriorityNormal,             /* Priority of the secure box's main thread */
+                1024);                        /* Stack size for the secure box's main thread */
+UVISOR_BOX_CONFIG(private_timer,              /* Name of the secure box */
+                  g_private_timer_acls,       /* ACLs list for the secure box */
+                  1024,                       /* Stack size for the secure box */
+                  PrivateTimerStaticMemory);  /* Private static memory for the secure box. */
+```
+
+### Create the secure box's main thread function
+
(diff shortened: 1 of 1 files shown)


docs/QUICKSTART.md
```diff
@@ -0,0 +1,403 @@
+# Quick-Start Guide for uVisor on mbed OS
+
+This guide will help you get started with uVisor on mbed OS.
+
+The uVisor creates sandboxed environments and resources protection for application built for ARM Cortex-M3 and Cortex-M4 devices. Here we will show you how to enable the uVisor and configure a secure box to get hold of some exclusive resources (memory, peripherals, interrupts). For more information on the uVisor design philosophy, please check out our the uVisor [introductory document](../README.md).
```
Patater: Sometimes you say uVisor, sometimes "the uVisor". We should pick one way and stick with it. Probably we should be just "uVisor", and never "the uVisor"; this seems to be the most common usage.

Patater:
Use this link to review rendered markdown https://github.com/AlessandroA/uvisor/blob/add_quickstart_guide/docs/QUICKSTART.md

docs/QUICKSTART.md
```diff
@@ -0,0 +1,403 @@
...
+
+* You are developing on a \*nix machine, in the `~/code` folder.
+* You are building the app for the [NXP FRDM-K64F](http://developer.mbed.org/platforms/FRDM-K64F/) target, with the [GCC ARM Embedded](https://launchpad.net/gcc-arm-embedded) toolchain.
+
+The instructions provided can be easily generalized to the case of other targets on other host OSs.
+
+## Start with the `blinky` app
+[Go to top](#overview)
+
+To create a new mbed application called `uvisor-example`just run the following commands:
```
Patater: Add a comma and a space after `uvisor-example`

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+```bash
+$ cd ~/code
+$ mbed new uvisor-example
+```
+
+The mbed-cli tools will automatically fetch the mbed code-base for you. By default, git will be used to track your code changes, so your application will be ready to be pushed to a git server, if you want to.
```
Patater: s/code-base/codebase

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+};
+
+/* Enable uVisor, using the ACLs we just created. */
+UVISOR_SET_MODE_ACL(UVISOR_ENABLED, g_main_box_acls);
+
+/* Rest of the existing app code */
+...
+```
+
+If you enable uVisor in the `blinky` app as it was written above, you will not get any particular security feature. All code and resources share the same security context, which we call the *main box*.
```
Patater: I'm not sure when we want to start calling the _main box_ the _legacy box_. Maybe now?

docs/QUICKSTART.md
AlessandroA: I like "uVisor" much more than "the uVisor", but am not sure if there was a "marketing" decision to use the article. @meriac?

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+A lot happens under the hood, though. All the user code now runs in unprivileged mode, and the systems services like the `NVIC` APIs or the OS SVCalls are routed through the uVisor.
+
+In the code above we specified 3 elements:
+
+1. System-wide uVisor configurations: `UVISOR_SET_PRIV_SYS_IRQ_HOOKS`. This macro is only needed temporarily, as it will be set by the operating system in the future.
```
Patater: > This macro is only needed temporarily, as it will be set by the operating system in the future.
Application authors currently need to specify the privileged system IRQ hooks at the application level with this macro, but in the future, the operating system will register the privileged system IRQ hooks on its own.

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+/* Enable uVisor, using the ACLs we just created. */
+UVISOR_SET_MODE_ACL(UVISOR_ENABLED, g_main_box_acls);
+
+/* Rest of the existing app code */
+...
+```
+
+If you enable uVisor in the `blinky` app as it was written above, you will not get any particular security feature. All code and resources share the same security context, which we call the *main box*.
+
+A lot happens under the hood, though. All the user code now runs in unprivileged mode, and the systems services like the `NVIC` APIs or the OS SVCalls are routed through the uVisor.
```
Patater: This doesn't happen quite yet. Perhaps you should move your discussion of adding the secure target to right before this paragraph. Specifically, if they build the application here, without `UVISOR_SUPPORTED`, what you wrote is not yet true.

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+A secure box is a special compartment that is granted exclusive access to peripherals, memories and interrupts. Private resources are only accessible when the *context* of the secure box is active. The uVisor is the only one that can enable a secure box context, for example upon thread switching or interrupt handling.
+
+Code that belongs to a box is not obfuscated by uVisor, so it is still readable and executable from outside of the box. In addition, declaring an object in the same file that configures a secure box does not protect that object automatically.
+
+Instead, we provide specific APIs to instruct the uVisor to protect a private resource. Here we will show how to use these APIs in the `uvisor-example` app.
+
+### Configure the secure box
+
+For this example, we want to create a secure box called `private_timer`. The `private_timer` box has exclusive access to the PIT timer and to the GPIO PORT C on the NXP FRDM-K64F board. These peripherals cannot be accessed from outside the secure box.
```
Patater: > These peripherals cannot be accessed from outside the secure box.

We intend to create a secure box that will prevent these peripherals from being accessed from outside the secure box.

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+Each secure box must have at least one thread, which we call the box main thread. In our `private_timer` box we will only use this thread throughout the whole program. The thread will constantly save the current timer value in a private buffer. In addition, we want to print the content of the buffered timer values whenever we press the button on the board.
```
Patater: s/box main thread/box's main thread

Patater: > whenever we press the button on the board

`the button` means which button? Maybe `the user button`.

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+We want the box to have exclusive access to the following resources:
+
+* The timer and push-button peripherals (peripheral ACLs). Nobody else should be able to read the timer values.
```
Patater: > (peripheral ACLs)

(as specified by peripheral ACLs)

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+/* Register privleged system IRQ hooks.
+ * This is a system-wide configuration and it is independent from the app, but
+ * for the moment it needs to be specified in the app. This will change in a
+ * later version: The configuration will be provided by the OS. */
+extern "C" void SVC_Handler(void);
+extern "C" void PendSV_Handler(void);
+extern "C" void SysTick_Handler(void);
+UVISOR_SET_PRIV_SYS_IRQ_HOOKS(SVC_Handler, PendSV_Handler, SysTick_Handler);
+
+/* Main box ACLs. */
```
Patater: This is the first mention of ACL in this document. It may be worth explicitly introducing the term someplace in this document, to keep the quick start self-contained (avoiding making the user read the entire README first).

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+For this example, we want to create a secure box called `private_timer`. The `private_timer` box has exclusive access to the PIT timer and to the GPIO PORT C on the NXP FRDM-K64F board. These peripherals cannot be accessed from outside the secure box.
+
+Each secure box must have at least one thread, which we call the box main thread. In our `private_timer` box we will only use this thread throughout the whole program. The thread will constantly save the current timer value in a private buffer. In addition, we want to print the content of the buffered timer values whenever we press the button on the board.
+
+We want the box to have exclusive access to the following resources:
+
+* The timer and push-button peripherals (peripheral ACLs). Nobody else should be able to read the timer values.
+* The push-button interrupt (IRQ ACL). We want the button IRQ to be re-routed to our box-specific ISR.
+* The buffer that holds the timer samples (dynamic memory ACL).
```
Patater: > (dynamic memory ACL)

(as specified by a dynamic memory ACL)

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+### Configure the secure box
+
+For this example, we want to create a secure box called `private_timer`. The `private_timer` box has exclusive access to the PIT timer and to the GPIO PORT C on the NXP FRDM-K64F board. These peripherals cannot be accessed from outside the secure box.
+
+Each secure box must have at least one thread, which we call the box main thread. In our `private_timer` box we will only use this thread throughout the whole program. The thread will constantly save the current timer value in a private buffer. In addition, we want to print the content of the buffered timer values whenever we press the button on the board.
+
+We want the box to have exclusive access to the following resources:
+
+* The timer and push-button peripherals (peripheral ACLs). Nobody else should be able to read the timer values.
+* The push-button interrupt (IRQ ACL). We want the button IRQ to be re-routed to our box-specific ISR.
```
Patater: > (IRQ ACL)

(as specified by an IRQ ACL)

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+/* ACLs list for the secure box: Timer (PIT). */
+static const UvisorBoxAclItem g_private_timer_acls[] = {
+    {PIT,   sizeof(*PIT),   UVISOR_TACLDEF_PERIPH}
+    {PORTC, sizeof(*PORTC), UVISOR_TACLDEF_PERIPH},
+};
+
+static void private_timer_main_thread(const void *);
+
+/* Secure box configuration */
+UVISOR_BOX_NAMESPACE(NULL);                   /* Box namespace. It can be NULL for this example. */
```
Patater: > `It can be NULL for this example.`

It can be NULL for any other example, or any other application, too.

`We won't specify a namespace for this example.`

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+UVISOR_BOX_HEAPSIZE(4096);                    /* Heap size for the secure box */
+UVISOR_BOX_MAIN(private_timer_main_thread,    /* Secure box main thread function */
```
Patater: s/Secure box main thread function/Main thread for the secure box

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+                1024);                        /* Stack size for the secure box main thread */
+UVISOR_BOX_CONFIG(private_timer,              /* Name of the secure box */
+                  g_private_timer_acls,       /* ACLs list for the secure box */
+                  1024,                       /* Stack size for the secure box */
+                  PrivateTimerStaticMemory);  /* Private static memory for the secure box. */
+```
+
+### Create the secure box main thread function
+
+In general, you can decide what to do in your box main thread. You can run it once and then kill it, use it to configure memories, peripherals, or to create other threads. In this app, the box main thread is the only thread for the `private_timer` box, and it will run throughout the whole program.
```
Patater: s/box main thread/box's main thread

Patater: > You can run it once and then kill it

Did you test that? Does it work as intended?

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+The `private_timer_main_thread` function configures the PIT timer, allocates the dynamic buffer to hold the timer values and initializes its private static memory `PrivateTimerStaticMemory`. A spinning loop is used to update the values in the buffer every time the thread is reactivated.
```
Patater: Add comma after `private static memory`

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+```C
+/* Number of timer samples we will use */
+#define PRIVATE_TIMER_BUFFER_COUNT 256
+
+/* For debug purposes: Print the buffer values when a button is pressed. */
```
Patater: s/Print/print

Patater: This also says "a button", whereas before we said "the button". Be consistent.

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+static void private_timer_button_on_press(void)
+{
+    for (int i = 0; i < PRIVATE_TIMER_BUFFER_COUNT; i++) {
+        printf("buffer[%03d] = %lu\r\n", i, uvisor_ctx->buffer[i]);
+    }
+}
+
+/* Secure box main thread function */
```
Patater: s/Secure box main thread function/Main thread for the secure box

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
+# Quick-Start Guide for uVisor on mbed OS
```
Patater: We should probably explicitly mention that this guide's code examples and mbed target descriptions are K64F specific, as we use specific peripherals (SW2, SIM, PORTC, etc.) and extend the existing K64F target to add UVISOR_SUPPORTED.

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+* The content of the private memory `PrivateTimerStaticMemory` can be accessed using the `PrivateTimerStaticMemory * uvisor_ctx` pointer, which is maintained by uVisor.
+* The `InterruptIn` object triggers the registration of an interrupt slot. Since that code is run in the context of the `private_timer` box, then the push-button IRQ belongs to that box. If you want to use the IRQ APIs directly, read the [section](#the-nvic-apis) below.
+* Even if the `private_timer_button_on_press` function runs in the context of `private_timer`, we can still use the `printf` function, which accesses the `UART0` peripheral, owned by the main box. This is because all ACLs declared in the main box are by default shared with all the other secure boxes. This also means that the messages we are printing on the serial port are not secure, because other boxes have access to that peripheral.
+
+> **Note**: Instantiating an object in the `secure_box.cpp` global scope will automatically map it to the main box context (heap), not the `private_timer` one. If you want an object to be private to a box, you need to instantiate it inside the code that will run in the context of that box.
```
Patater: Maybe this should be stronger than a `Note`. Perhaps a `Caution` or `Look out!` or whatever is standard in mbed documentation for `You are about to make a mistake!`.

Patater: > you need to instantiate it inside the code that will run in the context of that box.

or inside the `BoxContext` (in this example, `PrivateTimerStaticMemory`)

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+**Checkpoint**
+
+Compile the application again, re-flash the device, and press the reset button. The device LED should be blinking as in the previous case.
+
+If you don't see the LED blinking, it means that the application halted somewhere, probably because uVisor captured a fault. You can setup the uVisor debug messages to see if there is any problem. Follow the [Debugging uVisor](DEBUGGING.md) document for a step-by-step guide.
+
+If the LED is blinking, it means that the app is running fine. If you now press the `SW2` button on the NXP FRDM-K64F board, the `private_timer_button_on_press` function will be executed, printing the values in the timer buffer. You can observe these values by opening a serial port connection to the device, with a baud rate of 9600. When the print is completed, you should see the LED blinking again.
+
+### Allow other boxes to talk to the secure box
```
Patater: Boxes can insecurely talk to each other through the _main box_. Perhaps rename this section to `Allow other boxes to securely communicate to the secure box`.

Why `secure box` and not `secure boxes`? If because there is only one secure box in this example, then consider saying `main box` instead of `other boxes` as that's the only other box.

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+## Appendix
+[Go to top](#overview)
+
+This section contains additional information that you might find useful when setting up a secure box.
+
+### The NVIC APIs
+
+The ARM CMSIS header files provide APIs to configure, enable and disable IRQs in the NVIC module. These APIs are all prefixed with `NVIC_` and can be found in the `core_cm*.h` files in your CMSIS module.
+
+In addition, the CMSIS header also provide APIs to set and get an interrupt vector at runtime. This require the vector table, which is usually located in flash, to be relocated to SRAM.
```
Patater: s/require/requires

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+When the uVisor is enabled, all NVIC APIs are re-routed to the corresponding uVisor vIRQ APIs, which virtualise the interrupt module. The uVisor interrupts model has the following features:
```
Patater: s/virtualise/virtualize

Patater: s/interrupts model/interrupt model

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+* The uVisor owns the vector table.
```
Patater: interrupt vector table

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+ * Without uVisor: Relocate the vector table to SRAM and set my_isr as the ISR for MY_IRQ.
+ * With    uVisor: Register MY_IRQ for the current box with my_isr as ISR. */
+NVIC_SetVector(MY_IRQ, &my_isr);
+
+/* Change the IRQ state. */
+NVIC_SetPriority(MY_IRQ, 3);
+NVIC_EnableIRQ(MY_IRQ);
+```
+
+> **Note**: In this model a call to `NVIC_SetVector` must always happen before an IRQ state is changed. In platforms that don't relocate the vector table such call might be originally absent and must be added to work with uVisor.
```
Patater: s/such call/such a call

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+For more information on the uVisor APIs, checkout the [uVisor APIs documentation](APIS.md).
```
Patater: This appears to be a dead link. Consider omitting this link until the link is no longer dead.

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+### The *main box* ACLs
+
+The code samples that we provide in this guide give you a ready-made list of ACLs for the main box. The list includes peripherals that we already know will be necessary to make the example app work, and it is specific to the NXP FRDM-K64F target.
```
Patater: Consider moving this paragraph more to the top of the document. This is not something that should be buried in end matter.

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+```
+***********************************************************
+                    BUS FAULT
+***********************************************************
+
+...
+
+* MEMORY MAP
+  Address:           0x4004800C
+  Region/Peripheral: SIM
```
Patater: Really? It says SIM and not like AIPS Slot 17 or something?

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+Now that you know which peripheral is causing the fault (the `SIM` peripheral, in this example), you can add its entry to the ACLs list:
+
+
+```C
+static const UvisorBoxAclItem g_main_box_acls[] = {
+    {SIM, sizeof(*SIM), UVISOR_TACLDEF_PERIPH},
+};
+```
+
+> **Note**: If the fault debug screen does not show the name of the peripheral, you need to look it up in the target device reference manual.
```
Patater: This is more likely, in my experience.

docs/QUICKSTART.md
```diff
@@ -0,0 +1,410 @@
...
+
+For readability, do not use the hard-coded addresses of your peripherals, but rather use the symbols provided by the target CMSIS module.
+
+The process is then repeated multiple times until all ACLs have been added to the list. When no other ACL is needed any more, the system will run without hitting a uVisor fault.
```
Patater: Don't use passive voice here

Patater:
Very nice work. The guide is quite readable and has a good narrative; it doesn't jump around too much. I could follow along pretty easily.

Most of my comments (especially for re-ordering things) are recommendations to improve clarity and technical accuracy (without compromising the ease of understanding, as technically accurate documents often do).

docs/QUICKSTART.md
AlessandroA: I rephrased the title altogether.

docs/QUICKSTART.md
AlessandroA: It's done at [line 18](https://github.com/ARMmbed/uvisor/pull/272/files#diff-e3d1d3587b83098597a158f42f54b071R18).

Patater: Listing the assumptions is good.

Consider expanding `This guide will help you get started with uVisor on mbed OS.` to:
`This guide will help you get started with uVisor on mbed OS by walking you through creating a sample application for the K64F.`

docs/QUICKSTART.md
AlessandroA: Didn't test, but it *shouldn't be a problem*™. The box's main thread is never used in any other way.

docs/QUICKSTART.md
AlessandroA: The doc will be in very soon (working on it), I'd leave it here so I don't need to make changes here later.

docs/QUICKSTART.md
AlessandroA: With the debug box it would show the peripheral name (*and* the AIPS slot on K64F, but I limited the output to what is relevant for the parsing).

docs/QUICKSTART.md
AlessandroA: Yes I know, but I still want to show here the golden way, which we want to implement soon anyway. Readers have immediate access to this **Note** right at the moment they realize they don't have the peripheral name, so it shouldn't confuse them.

docs/QUICKSTART.md
Patater: I'm not sure how forward looking this document should be. If too forward looking, or if it hard for readers to tell what is forward looking, they might get frustrated doing something not yet possible.

docs/QUICKSTART.md
AlessandroA: I prefer *main box*. *Legacy* implies that if you do something new you shouldn't use it (reminds me a bit of "deprecated"), but actually it's reasonable to leave a lot of stuff in the main box.

docs/QUICKSTART.md
AlessandroA: Sure but they have immediate access here to the "alternative" solution. It think the guide is nicer with the peripheral names 😄

...And when you don't see them on your screen and come back to the guide to complain, you'll see that there's a note addressing your problem already 😉

AlessandroA: (For example, on the line of "forward looking" I still left the priv hooks macro in, even if we want to move it somewhere else, and I still left some ACLs in that shouldn't be needed (utest + greentea issue we found yesterday. In those cases the guide really needs them)

Patater: Yes, they have access to the "alternative" solution. It's just a bit misleading is all. No real confusion should come of it, assuming the reader notices the "Note" box.

AlessandroA:
All comments from @Patater should now be addressed. The following are still an open question:

* [uVisor vs. The uVisor](https://github.com/ARMmbed/uvisor/pull/272#discussion_r68922847).
* [main box vs. legacy box](https://github.com/ARMmbed/uvisor/pull/272#discussion_r68924872).

PR #281, opened 2016-07-06
ARMv7-M MPU: Halt on a non-recoverable fault

AlessandroA:
Faults were being returned from in the ARMv7-M vMPU module, which led
to very ambiguous debug sessions.

@meriac @Patater @niklas-arm


Patater:
LGTM

PR #283, opened 2016-07-12
Add initial clang-format file

Patater:
This file attempts to describe our coding standard in a way that
clang-format understands. It isn't perfect yet, but it is a nice
starting point.

See <http://clang.llvm.org/docs/ClangFormat.html> for usage
instructions, including how to integrate clang-format with your favorite
text editor.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,95 @@
+# http://clang.llvm.org/docs/ClangFormatStyleOptions.html
+---
+Language: Cpp
+BasedOnStyle: Google
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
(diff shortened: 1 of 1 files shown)


AlessandroA:
Thanks! I think it's a great start, I've been using it lately. Could you add here (in the PR description and in the commit description) a one-line how-to to use it on a particular file/set of files?

When the file is mature enough we can think of ways to hook it in the contribution process or in the build process.

Patater:
Updated commit description to point to clang-format documentation.

Patater:
I've been using clang-format with my text editor to, from time to time, auto-format highlighted blocks of code.

meriac:
@Patater : please rebase

Patater:
@meriac: Rebased

PR #284, opened 2016-07-12
Fix 2 ARMv7-M vMPU bugs

AlessandroA:
This PR fixes 2 bugs:

* The box 0 ACLs (which are shared by default) were never used when handling a fault coming from `g_active_box != 0`.
* The stack ACL for `g_active_box !=0` was overwritten in the round-robin ACL application, which turns out to be a bug in a special case.

You can read more about the bugs and their fixes in the related commit messages.

@meriac @Patater @niklas-arm


Patater:
LGTM

PR #285, opened 2016-07-12
RPC Gateway

Patater:
Implement an in-flash, verifiable, callable RPC structure that nominates a function as suitable RPC target.

Please give a thorough review. This PR targets unstable, but is intended to be dev-branch quality, with the exception of the reliance on the `Dummy-implement rpc_fncall and fpc_fncall_async` commit, and with the exception that we don't yet have a complete end-to-end prototype to verify this design as implemented.

Note that this enables building uvisor-lib out of C++ files, if we so desire. We don't have a need for that yet, as the dummy RPC implementation in this PR could be in C, but I anticipate we might want to use C++ in implementing some portion of RPC. If it's a requirement that uvisor-lib must never use C++ ever, then we can easily drop those commits and reimplement the RPC dummy in C.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,11 +68,13 @@ API_VERSION:=$(API_DIR)/lib/$(PLATFORM)/$(BUILD_MODE)/$(CONFIGURATION_LOWER).txt
 # Release library source files
 # Note: We explicitly remove unsupported.c from the list of source files. It
 #       will be compiled by the host OS in case uVisor is not supported.
-API_SOURCES:=$(wildcard $(API_DIR)/src/*.c)
+API_SOURCES:=$(wildcard $(API_DIR)/src/*.c) $(wildcard $(API_DIR)/src/*.cpp)
 API_SOURCES:=$(filter-out $(API_DIR)/src/unsupported.c, $(API_SOURCES))
 
 # Release library object files
-API_OBJS:=$(foreach API_SOURCE, $(API_SOURCES), $(CONFIGURATION_PREFIX)/$(API_DIR)/$(notdir $(API_SOURCE:.c=.o))) \
+API_OBJS:=$(API_SOURCES:.c=.o) # Replace .c with .o
+API_OBJS:=$(API_OBJS:.cpp=.o) # Replace .cpp with .o
+API_OBJS:=$(foreach API_SOURCE, $(API_OBJS), $(CONFIGURATION_PREFIX)/$(API_DIR)/$(notdir $(API_SOURCE))) \
           $(API_ASM_OUTPUT:.s=.o)
 
 # Make the ARMv7-M MPU driver the default.
@@ -129,8 +131,9 @@ OBJS:=$(foreach SOURCE, $(SOURCES), $(CONFIGURATION_PREFIX)/$(CORE_DIR)/$(notdir
 # build the core or the release library (so the two can have the same names
 # without collisions).
 ifneq ($(MAKECMDGOALS),build_core)
-vpath %.c $(API_DIR)/src
-vpath %.s $(API_DIR)/src
+vpath %.c   $(API_DIR)/src
+vpath %.cpp $(API_DIR)/src
+vpath %.s   $(API_DIR)/src
 else
 vpath %.c $(CORE_SYSTEM_DIR)/src:\
           $(CORE_SYSTEM_DIR)/src/mpu:\
@@ -189,7 +192,7 @@ CFLAGS_PRE:=\
 
 CFLAGS:=$(FLAGS_CM4) $(CFLAGS_PRE)
 CPPFLAGS:=
-CXXFLAGS:=-fno-exceptions
+CXXFLAGS:=$(CFLAGS) -fno-exceptions
 
 LINKER_CONFIG:=\
     -D$(CONFIGURATION) \
@@ -293,6 +296,10 @@ $(CONFIGURATION_PREFIX)/$(CORE_DIR)/%.o: %.c
 # Pre-process and compile a release library C file into an object file.
 $(CONFIGURATION_PREFIX)/$(API_DIR)/%.o: %.c
 	$(CC) $(CPPFLAGS) $(CFLAGS) -c $< -o $@
+
+# Pre-process and compile a release library C++ file into an object file.
+$(CONFIGURATION_PREFIX)/$(API_DIR)/%.o: %.cpp
+	$(CXX) $(CPPFLAGS) $(CXXFLAGS) -c $< -o $@
 
 # Generate the output asm file from the asm input file and the INCBIN'ed binary.
 $(API_ASM_OUTPUT:.s=.o): $(CONFIGURATION_PREFIX).bin $(API_ASM_INPUT)
--- a/Makefile
+++ b/Makefile
@@ -188,6 +188,7 @@ CFLAGS_PRE:=\
         -fdata-sections
 
 CFLAGS:=$(FLAGS_CM4) $(CFLAGS_PRE)
+CPPFLAGS:=
 CXXFLAGS:=-fno-exceptions
 
 LINKER_CONFIG:=\
@@ -287,17 +288,17 @@ $(CONFIGURATION_PREFIX).linker: $(CORE_LINKER_DIR)/default.h
 
 # Pre-process and compile a core C file into an object file.
 $(CONFIGURATION_PREFIX)/$(CORE_DIR)/%.o: %.c
-	$(CC) $(CXXFLAGS) $(CFLAGS) -c $< -o $@
+	$(CC) $(CPPFLAGS) $(CFLAGS) -c $< -o $@
 
 # Pre-process and compile a release library C file into an object file.
 $(CONFIGURATION_PREFIX)/$(API_DIR)/%.o: %.c
-	$(CC) $(CXXFLAGS) $(CFLAGS) -c $< -o $@
+	$(CC) $(CPPFLAGS) $(CFLAGS) -c $< -o $@
 
 # Generate the output asm file from the asm input file and the INCBIN'ed binary.
 $(API_ASM_OUTPUT:.s=.o): $(CONFIGURATION_PREFIX).bin $(API_ASM_INPUT)
 	cp $(API_ASM_HEADER) $(API_ASM_OUTPUT)
 	$(CPP) -w -P $(API_CONFIG) $(API_ASM_INPUT) >> $(API_ASM_OUTPUT)
-	$(CC) $(CXXFLAGS) $(CFLAGS) -c $(API_ASM_OUTPUT) -o $@
+	$(CC) $(CPPFLAGS) $(CFLAGS) -c $(API_ASM_OUTPUT) -o $@
 
 ctags: source.c.tags
 
--- a/Makefile
+++ b/Makefile
@@ -188,7 +188,7 @@ CFLAGS_PRE:=\
         -fdata-sections
 
 CFLAGS:=$(FLAGS_CM4) $(CFLAGS_PRE)
-CPPFLAGS:=-fno-exceptions
+CXXFLAGS:=-fno-exceptions
 
 LINKER_CONFIG:=\
     -D$(CONFIGURATION) \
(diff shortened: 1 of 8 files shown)


api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,224 @@
...
+ * @param fn_name[in]  The function being designated as an RPC target
+ * @param fn_type[in]  The return type of the function being designated as an
+ *                     RPC target
+ * @param __VA_ARGS__  The type of each parameter passed to the target
+ *                     function. There can be up to 4 parameters. Each
+ *                     parameter must be no more than uint32_t in size.
+ */
+#define UVISOR_BOX_RPC_GATEWAY_SYNC(box_name, gw_name, fn_name, fn_type, ...) \
+    UVISOR_STATIC_ASSERT(sizeof(fn_type) <= sizeof(uint32_t), gw_name ## _return_type_too_big); \
+    _UVISOR_BOX_RPC_GATEWAY_ARG_CHECK(gw_name, __VA_ARGS__) \
```
AlessandroA: It needs to be `##__VA_ARGS__` if the variable arguments can be 0.

api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,224 @@
...
+    _UVISOR_BOX_RPC_GATEWAY_SYNC_CALLER(fn_name, __VA_ARGS__) \
```
AlessandroA: It needs to be `##__VA_ARGS__` if the variable arguments can be 0.

api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,224 @@
...
+ * @param fn_name[in]  The function being designated as an RPC target
+ * @param fn_type[in]  The return type of the function being designated as an
+ *                     RPC target
+ * @param __VA_ARGS__  The type of each parameter passed to the target
+ *                     function. There can be up to 4 parameters. Each
+ *                     parameter must be no more than uint32_t in size.
+ */
+#define UVISOR_BOX_RPC_GATEWAY_ASYNC(box_name, gw_name, fn_name, fn_type, ...) \
+    UVISOR_STATIC_ASSERT(sizeof(fn_type) <= sizeof(uint32_t), gw_name ## _return_type_too_big); \
+    _UVISOR_BOX_RPC_GATEWAY_ARG_CHECK(gw_name, __VA_ARGS__) \
```
AlessandroA: It needs to be `##__VA_ARGS__` if the variable arguments can be 0.

api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,224 @@
...
+    _UVISOR_BOX_RPC_GATEWAY_ASYNC_CALLER(fn_name, __VA_ARGS__) \
```
AlessandroA: It needs to be `##__VA_ARGS__` if the variable arguments can be 0.

api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,224 @@
...
+        .function = (uint32_t) _sgw_async_ ## fn_name, \
+    }; \
+    \
+    /* Pointer to the gateway we just created. The pointer is located in a
+     * discoverable linker section. */ \
+    __attribute__((section(".keep.uvisor.rpc_gateway_ptr"))) \
+    static uint32_t const gw_name ## _rpc_gateway_ptr = (uint32_t) &gw_name ## _rpc_gateway; \
+    \
+    /* Declare the actual gateway. */ \
+    UVISOR_EXTERN int (*gw_name)(uvisor_rpc_result_t *, __VA_ARGS__) __attribute__((alias(UVISOR_TO_STRING(gw_name ## _rpc_gateway))));
```
AlessandroA: It needs to be `##__VA_ARGS__` if the variable arguments can be 0.

api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,224 @@
...
+
+#define _UVISOR_BOX_RPC_GATEWAY_ARG_CHECK(gw_name, ...) \
+    __UVISOR_BOX_MACRO(__VA_ARGS__, _UVISOR_BOX_RPC_GATEWAY_ARG_CHECK_4, \
+                                    _UVISOR_BOX_RPC_GATEWAY_ARG_CHECK_3, \
+                                    _UVISOR_BOX_RPC_GATEWAY_ARG_CHECK_2, \
+                                    _UVISOR_BOX_RPC_GATEWAY_ARG_CHECK_1, \
+                                    _UVISOR_BOX_RPC_GATEWAY_ARG_CHECK_0)(gw_name, __VA_ARGS__)
```
AlessandroA: It needs to be `##__VA_ARGS__` if the variable arguments can be 0.

api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,224 @@
...
+    UVISOR_STATIC_ASSERT(sizeof(p1_type) <= sizeof(uint32_t), gw_name ## _param_1_too_big); \
+    UVISOR_STATIC_ASSERT(sizeof(p2_type) <= sizeof(uint32_t), gw_name ## _param_2_too_big); \
+    UVISOR_STATIC_ASSERT(sizeof(p3_type) <= sizeof(uint32_t), gw_name ## _param_3_too_big);
+
+#define _UVISOR_BOX_RPC_GATEWAY_SYNC_CALLER(fn_name, ...) \
+    __UVISOR_BOX_MACRO(__VA_ARGS__, _UVISOR_BOX_RPC_GATEWAY_SYNC_CALLER_4, \
+                                    _UVISOR_BOX_RPC_GATEWAY_SYNC_CALLER_3, \
+                                    _UVISOR_BOX_RPC_GATEWAY_SYNC_CALLER_2, \
+                                    _UVISOR_BOX_RPC_GATEWAY_SYNC_CALLER_1, \
+                                    _UVISOR_BOX_RPC_GATEWAY_SYNC_CALLER_0)(fn_name, __VA_ARGS__)
```
AlessandroA: It needs to be `##__VA_ARGS__` if the variable arguments can be 0.

api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,224 @@
...
+        TFN_Ptr fp = (TFN_Ptr) fn_name; \
+        return rpc_fncall(p0, p1, p2, p3, fp); \
+    }
+
+#define _UVISOR_BOX_RPC_GATEWAY_ASYNC_CALLER(fn_name, ...) \
+    __UVISOR_BOX_MACRO(__VA_ARGS__, _UVISOR_BOX_RPC_GATEWAY_ASYNC_CALLER_4, \
+                                    _UVISOR_BOX_RPC_GATEWAY_ASYNC_CALLER_3, \
+                                    _UVISOR_BOX_RPC_GATEWAY_ASYNC_CALLER_2, \
+                                    _UVISOR_BOX_RPC_GATEWAY_ASYNC_CALLER_1, \
+                                    _UVISOR_BOX_RPC_GATEWAY_ASYNC_CALLER_0)(fn_name, __VA_ARGS__)
```
AlessandroA: It needs to be `##__VA_ARGS__` if the variable arguments can be 0.

api/inc/rpc_gateway.h
AlessandroA: I'll add the reason here for the [comment above](https://github.com/ARMmbed/uvisor/pull/285/commits/a319f5049b8bea8b391a44ab424c6cdd9e0c6ecb#r71174204), which applies to all other comments.

If you are passing `__VA_ARGS__` to a macro that also takes fixed arguments, like in this example:

```C
#define SOME_MACRO(a, ...) __SOME_MACRO(a, __VA_ARGS__)
```

then, when `__VA_ARGS__` is empty, the macro si resolved to `__SOME_MACRO(a, )`, which is a macro with 2 arguments (one of them being the *empty token*).

To avoid this, the GNU extensions (also available on armcc and IAR), allow the use of `##__VA_ARGS__`, which swallows a leading comma when `__VA_ARGS__` is the empty token.

See [here](https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html) for more details.

api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,224 @@
...
+
+#define _UVISOR_BOX_RPC_GATEWAY_SYNC_CALLER_4(fn_name, ...) \
+    static uint32_t __sgw_sync_ ## fn_name(uint32_t p0, uint32_t p1, uint32_t p2, uint32_t p3) \
+    { \
+        TFN_Ptr fp = (TFN_Ptr) fn_name; \
+        return rpc_fncall(p0, p1, p2, p3, fp); \
+    }
+
+#define _UVISOR_BOX_RPC_GATEWAY_ASYNC_CALLER(fn_name, ...) \
+    __UVISOR_BOX_MACRO(__VA_ARGS__, _UVISOR_BOX_RPC_GATEWAY_ASYNC_CALLER_4, \
```
AlessandroA: If this macro allows an empty `__VA_ARGS__` then the macro to use should actually be [this one](https://github.com/ARMmbed/uvisor/blob/master/api/inc/uvisor_exports.h#L62), as used [here](https://github.com/ARMmbed/uvisor/blob/master/api/inc/uvisor_exports.h#L71).

For a long time I've wanted to unify our variadic macros creation, as I'd like to remove duplications like this one (`__UVISOR_BOX_MACRO`).

Makefile
```diff
@@ -188,7 +188,7 @@ CFLAGS_PRE:=\
         -fdata-sections
 
 CFLAGS:=$(FLAGS_CM4) $(CFLAGS_PRE)
-CPPFLAGS:=-fno-exceptions
+CXXFLAGS:=-fno-exceptions
```
AlessandroA: Yes, these flags were for C++, not the C pre-processor.

Please modify this commit so that all rules that build a C file apply `$(CPPFLAGS) $(CFLAGS)`, and all the rules that build C++ files apply `$(CPPFLAGS) $(CXXFLAGS)`.

The flags will be defined as follows:

```make
CPPFLAGS:=
CXXFLAGS:=-fno-exceptions
```

___


This could probably be done in 2 commits, but I think you would then need to re-shuffle the previous commit that introduces compilation for C++ files. I think it's easier to just modify this one.

api/inc/rpc_gateway.h
Patater: I'm quite familiar with ##__VA_ARGS__. I try to avoid GCC-specific extensions where possible in user/library code. (GCC-specific stuff is fine in uVisor itself.)

For the RPC creating gateway macros, we can get by without GCC-specific extensions as we do require 1 argument. If the function you are writing a gateway to doesn't take any parameters, specify `void`.

I'll add a comment in the doxygen notes that at least one parameter is required.

api/inc/rpc_gateway.h
Patater: I'm also passing along the fn_name separately from the `__VA_ARGS__`. The macros are a bit different, but probably could be refactored.

Makefile
Patater: It was easy to do in 2 commits, and I think it makes sense to separate the change into two commits in order to keep functional changes separate from cosmetic ones; it's easier to see that "nothing is up my sleeve".

api/inc/rpc_gateway.h
AlessandroA: Do we require 1 argument as a consequence of this restriction or because we really want to? Relying on the GNU extensions is reasonable enough as all our toolchains support them (and there are ways anyways to workaround the need of `##` in case we wanted to).

I want to give priority to the API cleanliness and friendliness: If it's valuable to have 0 args then let's use `##`.

Otherwise yes, please add it to the docs. Likely, having 0 arguments makes the macro you used for variadic selection not work, so you get the error "for free".

api/inc/rpc_gateway_exports.h
```diff
@@ -0,0 +1,59 @@
...
+
+/** RPC gateway structure
+ *
+ * This struct is packed because we must ensure that the `ldr_pc` field has no
+ * padding before itself and will be located at a valid instruction location,
+ * and that the `function` field is at a pre-determined offset from the
+ * `ldr_pc` field.
+ */
+typedef struct {
+    uint32_t ldr_pc;
```
AlessandroA: Just to make sure that this is tested: The function that is loaded into the `pc` becomes a direct jump and never returns to the gateway, since `lr` is never updated. Besides, even if that function returned, it would then step over literal pool data (the metadata).

Has it been tested/verified in ASM that the destination function (the one that posts the RPC) correctly returns to the original caller (the caller of the gateway)?

Patater: Yes, this entire PR, and specifically the `ldr pc, [pc, #immediate]`, has been tested with https://github.com/Patater/example-uvisor-threaded-blinky/tree/rpc-demo-gateway

AlessandroA:
The docs for `UVISOR_BOX_RPC_MAXIMUM_INCOMING_RPC` are missing right?

Patater:
The docs for `UVISOR_BOX_RPC_MAXIMUM_INCOMING_RPC` are in the "Limiting the maximum number of incoming RPC messages" section. Did you want something more added in that section?

AlessandroA:
Sorry my mistake, I meant the header file. I couldn't find that API here. It should be in the user-facing RPC header (or in the box-config one, up to us).

Patater:
Ah, `UVISOR_BOX_RPC_MAXIMUM_INCOMING_RPC` isn't part of this PR. It isn't necessary that `UVISOR_BOX_RPC_MAXIMUM_INCOMING_RPC` be implemented for the gateways to work.

AlessandroA:
True, but I'd like to deliver for today the whole set of user-facing APIs, which also include this one. If we manage to review and merge this PR by the end of today, the API signatures will be merged into master! 😃

Patater:
I'm not so comfortable merging this work to master yet. dev would be okay.

AlessandroA:
But it's just API signatures right? No implementation. We need to be confident about these APIs as they are anyway part of the public release (which will point to master). It's fine if those API will change slightly in the future, although of course we're trying to get them right from the beginning  😉

Patater:
There may be other changes we want to make to the APIs as we progress with implementation. These APIs are 100% ready for the release, and not ready to commit these APIs to master, until that implementation work is complete.

There are also commits I still want to rebase before they become unrebasable. Specifically, I'd like to have the real implementation of the dummies (instead of the dummy implementation) in history in a sane progression.

I dummy-implemented the `UVISOR_BOX_RPC_MAXIMUM_INCOMING_RPC` macro in `rpc.h` in a new commit.

Patater:
Closing this PR to re-open another against dev instead of against unstable.

PR #273, opened 2016-06-29
Use bitmaps to map page ownership to box id.

niklas-arm:
This changes the page owner table to a page owner bitmap. This allows
to query the ownership of a page in O(1), rather than linear search, and
makes it possible to OR together box 0 and active box page bitmap to get
the active pages in O(1).

This is a preparation for porting the page allocator protection to the ARMv7-M MPU.
By storing the ownership information in a bitmap, it makes using the MPU subregions much easier (literally just masking the right 8 bits).

This has been tested on K64F without any failures.

cc @meriac @AlessandroA @Patater

PR #286, opened 2016-07-18
Port page heap access switching to ARMv7-M MPU

niklas-arm:
This adds the necessary changes to use subregions to access the page heap.
Changes include accessing the page owner bitmap efficiently to map it directly to MPU subregions and moving the page heap to behind the uVisor `.bss` section and aligning it's end with a subregion of the SRAM access region.

Note: This builds on #284 and cherry-picks #280 and #273 from `unstable` to `dev`.


niklas-arm:
Closed in favor of #294, since it includes all these changes.

PR #287, opened 2016-07-19
RPC Gateway

Patater:
Implement an in-flash, verifiable, callable RPC structure that nominates a function as suitable RPC target.

This PR targets dev. However, we don't yet have a complete end-to-end prototype to verify this design. Changes could be made to the API as a result of further development.


Patater:
I dropped the `uvisor-lib: Enable building of C++ files` commit because we don't need it yet. I'll add it to whatever PR ends up actually needing it.

AlessandroA:
@meriac Ready for review!

Patater:
Please hold off merging while I re-test the code after most recent modifications.

Patater:
Testing done

api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,237 @@
...
+ * is read and validated by uVisor before performing the operation.
+ *
+ * Create function with following signature:
+ * UVISOR_EXTERN fn_type fn_name(uint32_t a, uint32_t b);
+ *
+ * @param box_name[in] The name of the source box as declared in
+ *                     `UVISOR_BOX_CONFIG`
+ * @param gw_name[in]  The new, callable function pointer for performing RPC
+ * @param fn_name[in]  The function being designated as an RPC target
+ * @param fn_type[in]  The return type of the function being designated as an
```
meriac: @Patater please change to fn_ret_type or similar (indicating that's just the type of the return vs. type of fn)

Patater: Fixed

api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,237 @@
...
+#define _UVISOR_BOX_RPC_GATEWAY_ASYNC_CALLER_4(fn_name, ...) \
+    static uint32_t _sgw_async_ ## fn_name(uvisor_rpc_result_t result, uint32_t p0, uint32_t p1, uint32_t p2, uint32_t p3) \
+    { \
+        TFN_Ptr fp = (TFN_Ptr) fn_name; \
+        return rpc_fncall_async(p0, p1, p2, p3, fp, result); \
+    }
+
+/* This function is private to uvisor-lib, but needs to be publicly visible for
+ * the RPC gateway creation macros to work. */
+UVISOR_EXTERN uint32_t rpc_fncall(uint32_t p0, uint32_t p1, uint32_t p2, uint32_t p3, const TFN_Ptr fn);
```
meriac: @Patater : please rename rpc_fncall to rpc_fncall_sync

Patater: Renamed

Patater:
I am working on eliminating the issue that RPC gateways will only be callable outside of the  translation unit where they are declared. That was too much of a pain in the butt.

Patater:
That limitation has been eliminated.

api/inc/rpc_gateway.h
```diff
@@ -0,0 +1,240 @@
...
+ * Create function with following signature:
+ * UVISOR_EXTERN fn_ret gw_name(uint32_t a, uint32_t b);
+ *
+ * @param box_name[in] The name of the source box as declared in
+ *                     `UVISOR_BOX_CONFIG`
+ * @param gw_name[in]  The new, callable function pointer for performing RPC
+ * @param fn_name[in]  The function being designated as an RPC target
+ * @param fn_ret[in]   The return type of the function being designated as an
+ *                     RPC target
+ * @param __VA_ARGS__  The type of each parameter passed to the target
```
AlessandroA: This does not state the minimum. For the current implementation, `__VA_ARGS__` cannot be empty. So the use must always specify at least one type (which is fine, that would be `void`).

AlessandroA:
LGTM

PR #293, opened 2016-07-21
Refactor the documentation into two folders

AlessandroA:
The net results of this PR are the following:

* All docs split in 2 folders, `api` and `core`.
* The `docs` folder has a dedicated `README.md` which de-multiplexes the right document to the right purpose.
    * The `README.md` file is always rendered when you open the `docs` folder on GitHub.

All docs references have been fixed.

@meriac @niklas-arm @Patater


Patater:
Enabling uVisor debug messages seems like a `core` sort of activity to me.

AlessandroA:
Not sure, I would expect an app user/writer to want to see debug messages when uVisor halts. Keep in mind that the document is from the mbed OS perspective (and that with the automatic debug-build selection is fairly easy if you have a GDB-enabled board). @meriac What do you think?

Patater:
I'll note that I like the agreement between the source code folder structure and the proposed documentation folder structure.

PR #294, opened 2016-07-22
Add interface for MPU region management and scheduling

niklas-arm:
This adds a common interface for the implementation of the MPU region storage and scheduling.
Tested on both K64F and EFM32.

This build on top of #286.
cc @AlessandroA @meriac @Patater.


core/system/inc/page_allocator_config.h
```diff
@@ -42,4 +42,33 @@ typedef uint8_t page_owner_t;
...
+    map[page / 32] &= ~(1UL << (page % 32));
+}
+
+/** Check if the page bit is set int the page map array.
+ * @param map   an array of `uint32_t` containing the page map
+ * @param page  the index of the page to be set
+ * @retval 0    if page bit is not set
+ * @retval !0   if page bit is set
+ */
+static inline uint32_t page_allocator_map_get(const uint32_t * const map, uint8_t page)
```
meriac: return should be "int" & of values 0/1 (to fit it later to bitband operations)

core/system/src/page_allocator.c
```diff
@@ -177,9 +177,19 @@ int page_allocator_malloc(UvisorPageTable * const table)
     uint32_t page = 0;
     for (; (page < g_page_count_total) && pages_required; page++) {
         /* If the page is unused, it's entry is UVISOR_PAGE_UNUSED (not NULL!). */
```
AlessandroA: I guess this comment is not true any more?

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -31,6 +31,30 @@
...
 #endif/*ARMv7M_MPU_REGIONS*/
 
+/* The ARMv7-M MPU has 8 MPU regions plus one background region.
+ * Region 0 and 1 are used to unlock Application RAM and Flash.
+ * When switching into a secure box, region 2 is used to protect the boxes
+ * stack and context.
+ * If a box uses the page heap, the next region is used to protect it.
+ * This leaves 4 to 6 MPU regions for round robin scheduling:
+ *
+ *      8      <-- End of MPU regions, K64F_MPU_REGIONS_MAX
```
AlessandroA: Can you change `K64F_MPU_REGIONS_MAX` into something non-K64-specific?

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -31,6 +31,30 @@
...
+ * +---------+
+ * |    7    |
+ * |   ...   |
+ * |  2/3/4  | <-- Start of round robin
+ * +---------+
+ * |   2/3   | <-- Optional Box Pages
+ * +---------+
+ * |    2    | <-- Secure Box Stack + Context, K64F_MPU_REGIONS_STATIC
```
AlessandroA: Can you change `K64F_MPU_REGIONS_STATIC` into something non-K64-specific?

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -704,45 +728,69 @@ void vmpu_arch_init_hw(void)
...
+    const uint32_t total_size = (1 << vmpu_region_bits((uint32_t) __uvisor_config.sram_end - (uint32_t) __uvisor_config.sram_start));
+    /* The alignment is 1/8th of the region size = rounded up SRAM size. */
+    const uint32_t subregions_size = total_size / 8;
+    const uint32_t protected_size = (uint32_t) __uvisor_config.page_end - (uint32_t) __uvisor_config.sram_start;
+    /* The protected size must be aligned to the subregion size. */
+    if (protected_size % subregions_size != 0) {
         HALT_ERROR(SANITY_CHECK_FAILED,
-                   "The __uvisor_bss_end symbol (0x%08X)is not aligned to an MPU subregion boundary.",
-                   (uint32_t) __uvisor_config.bss_end);
+                   "The __uvisor_page_end symbol (0x%08X)is not aligned to an MPU subregion boundary.",
```
AlessandroA:
```diff
-(0x%08X)is
+(0x%08X) is
```

core/system/src/mpu/vmpu_freescale_k64_mem.c
```diff
@@ -336,8 +336,8 @@ void vmpu_mem_init(void)
     MPU_Region * region;
 
     /* rest of SRAM, accessible to mbed - non-executable for uvisor */
-    res = vmpu_mem_add_int(0, __uvisor_config.bss_end,
-        UVISOR_REGION_ROUND_UP((uint32_t) __uvisor_config.page_start) - ((uint32_t) __uvisor_config.bss_end),
+    res = vmpu_mem_add_int(0, __uvisor_config.page_end,
+        (uint32_t) __uvisor_config.sram_end - ((uint32_t) __uvisor_config.page_end),
```
AlessandroA: Why did you remove the rounding?

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -704,45 +728,69 @@ void vmpu_arch_init_hw(void)
...
+    /* Unlock the upper SRAM subregion only. */
     vmpu_acl_static_region(
-        2,
-        (void *) SRAM_ORIGIN,
-        rounded_size,
-        UVISOR_TACL_SREAD | UVISOR_TACL_SWRITE | UVISOR_TACL_SUBREGIONS(subregions_mask)
+        1,
+        __uvisor_config.sram_start,
+        total_size,
+        UVISOR_TACLDEF_DATA | UVISOR_TACL_EXECUTE | UVISOR_TACL_SUBREGIONS(subregions_disable_mask)
```
AlessandroA: Is the execute permission required?

core/system/inc/page_allocator_config.h
```diff
@@ -26,29 +26,53 @@
...
 /** Sets the page bit in the page map array.
  * @param map   an array of `uint32_t` containing the page map
  * @param page  the index of the page to be set
  */
 static inline void page_allocator_map_set(uint32_t * const map, uint8_t page)
 {
+#if UVISOR_PAGE_OWNER_MAP_LEFT_ALIGNED
+    page += UVISOR_PAGE_MAP_COUNT * 32 - g_page_count_total;
+#endif
+#if UVISOR_PAGE_MAP_COUNT == 1
```
AlessandroA: I wouldn't add conditional code just for this corner case. The net result is the same.

core/system/inc/page_allocator_config.h
```diff
@@ -26,29 +26,53 @@
...
+    map[0] |= (1UL << page);
+#else
     map[page / 32] |= (1UL << (page % 32));
```
AlessandroA: Does it still work if `page` is 0?

core/system/inc/page_allocator_config.h
```diff
@@ -57,7 +81,14 @@ static inline void page_allocator_map_set(uint32_t * const map, uint8_t page)
  */
 static inline void page_allocator_map_clear(uint32_t * const map, uint8_t page)
 {
+#if UVISOR_PAGE_OWNER_MAP_LEFT_ALIGNED
+    page += UVISOR_PAGE_MAP_COUNT * 32 - g_page_count_total;
+#endif
+#if UVISOR_PAGE_MAP_COUNT == 1
```
AlessandroA: Same comments as above.

core/system/inc/page_allocator_config.h
```diff
@@ -68,7 +99,14 @@ static inline void page_allocator_map_clear(uint32_t * const map, uint8_t page)
  */
 static inline int page_allocator_map_get(const uint32_t * const map, uint8_t page)
 {
+#if UVISOR_PAGE_OWNER_MAP_LEFT_ALIGNED
+    page += UVISOR_PAGE_MAP_COUNT * 32 - g_page_count_total;
+#endif
+#if UVISOR_PAGE_MAP_COUNT == 1
```
AlessandroA: Same comments as above.

core/system/inc/page_allocator_faults.h
```diff
@@ -30,16 +30,36 @@ uint32_t page_allocator_get_faults(uint8_t page);
...
  * @returns Non-zero on failure with failure class `UVISOR_ERROR_CLASS_PAGE`. See `UVISOR_ERROR_PAGE_*`.
  */
 int page_allocator_get_active_region_for_address(uint32_t address, uint32_t * start_addr, uint32_t * end_addr, uint8_t * page);
 
+/** Map an address to an 8-bit page mask.
+ * If the address is not part of any page, or the page does not belong to the
+ * active box or box 0 an error is returned and `mask`, `index` and `page` are invalid.
+ *
+ * Example with 12 pages:
+ *  - right-aligned map: 0b00000000'00000000'00001111'11111111
```
AlessandroA: In general, is there really a strong case about having 2 implementations for the right- and left-aligned cases? It might a premature optimization that we don't need. What's the runtime difference?

core/system/inc/page_allocator_config.h
```diff
@@ -25,21 +25,88 @@
...
+/** Sets the page bit in the page map array.
+ * @param map   an array of `uint32_t` containing the page map
+ * @param page  the index of the page to be set
+ */
+static inline void page_allocator_map_set(uint32_t * const map, uint8_t page)
+{
+#if UVISOR_PAGE_OWNER_MAP_LEFT_ALIGNED
+    page += UVISOR_PAGE_MAP_COUNT * 32 - g_page_count_total;
+#endif
+#if UVISOR_PAGE_MAP_COUNT == 1
```
meriac: @niklas-arm why sis there a special case for UVISOR_PAGE_MAP_COUNT == 1 ? Isn't that covered by the general case below?

core/system/inc/page_allocator_config.h
```diff
@@ -25,21 +25,88 @@
...
+    map[0] |= (1UL << page);
+#else
```
meriac: @niklas-arm Please avoid the optimisation now for one dword-only map. I would like to reduce the test complexity for now. The overhead we save is neglegible compared to the rest of the call (2nd tier etc.)

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -146,60 +58,61 @@ static const TMpuRegion* vmpu_fault_find_region(uint32_t fault_addr)
...
-            return UVISOR_TACL_UWRITE | UVISOR_TACL_UREAD;
-        default:
-            /* translate fault_addr into its physical address if it is in the bit-banding region */
-            if (fault_addr >= VMPU_PERIPH_BITBAND_START && fault_addr <= VMPU_PERIPH_BITBAND_END) {
-                fault_addr = VMPU_PERIPH_BITBAND_ALIAS_TO_ADDR(fault_addr);
-            }
-            else if (fault_addr >= VMPU_SRAM_BITBAND_START && fault_addr <= VMPU_SRAM_BITBAND_END) {
-                fault_addr = VMPU_SRAM_BITBAND_ALIAS_TO_ADDR(fault_addr);
-            }
+    if (fault_addr == (uint32_t) &SCB->SCR) {
```
meriac: @niklas-arm @AlessandroA SCB->SCR access is strictly speaking a stability-backdoor. We must change that into a RLA gateway in mbed-os as we have it now. Please add a fat fixme here.

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -146,60 +58,61 @@ static const TMpuRegion* vmpu_fault_find_region(uint32_t fault_addr)
...
+    if (page_allocator_get_active_mask_for_address(fault_addr, &mask, &index, &page) == UVISOR_ERROR_PAGE_OK)
+    {
+        /* Remember this fault. */
+        page_allocator_register_fault(page);
 
-    /* In a secure box the first available region is for the stack, so it must
-     * be excluded from the round-robin. */
-    if (g_active_box != 0 && mpu_slot == ARMv7M_MPU_RESERVED_REGIONS) {
-        mpu_slot++;
+        vmpu_mem_push_page_acl_iterator(mask, UVISOR_PAGE_MAP_COUNT * 4 - 1 - index);
```
meriac: @niklas-arm - the "-index" - isn't that a double negation of [#234](https://github.com/ARMmbed/uvisor/pull/294/files#diff-d5e475afa26cdc172546bc52c5d788abR234) ?

core/system/src/mpu/vmpu_armv7m.c
niklas-arm: Yes, those were the permission from before. The user is allowed to execute code from RAM after all. Maybe we can turn it into `UVISOR_TACL_UEXECUTE`? I don't think we ever need to execute priviledged code from RAM.

core/system/src/mpu/vmpu_armv7m.c
niklas-arm: Yes, but code reuse.

niklas-arm:
Updated code to address review.

AlessandroA:
Could you please add here a comment specifying how we need to change the app and/or mbed OS to accommodate the new changes?

From what I've seen so far, I think at least the following is needed:

* Update the linker script of all uVisor-supported platforms (K64F, EFM32GG_STK3700, Beetle).
* Update the app to include the new config macro.
* Update the docs for porting (linker script) and quick-start (config macro).

Anything else I've missed?

core/system/src/page_allocator_faults.c
```diff
@@ -94,3 +119,30 @@ uint8_t page_allocator_iterate_active_pages(int (*callback)(uint32_t start_addr,
 
     return count;
 }
+
+uint8_t page_allocator_iterate_active_page_masks(int (*callback)(uint8_t mask, uint8_t index), PageAllocatorIteratorDirection direction)
```
meriac: @niklas-arm can you please use a globally defined type for the callback function?

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -329,52 +349,68 @@ void vmpu_switch(uint8_t src_box, uint8_t dst_box)
...
+    region = box->region;
+    dst_count = box->count;
+
+    /* Only write stack and context ACL for secure boxes. */
     if (dst_box)
     {
-        /* handle target box next */
-        box = &g_mpu_box[dst_box];
-        region = box->region;
+        assert(dst_count);
```
meriac: @niklas-arm moving the assert outside of the if-block (above) - as it's used also in other blocks.

core/system/src/mpu/vmpu_freescale_k64.c
```diff
@@ -277,31 +249,29 @@ void vmpu_switch(uint8_t src_box, uint8_t dst_box)
...
-            return UVISOR_TACL_UWRITE | UVISOR_TACL_UREAD;
-        default:
-            /* translate fault_addr into its physical address if it is in the bit-banding region */
-            if (fault_addr >= VMPU_PERIPH_BITBAND_START && fault_addr <= VMPU_PERIPH_BITBAND_END) {
-                fault_addr = VMPU_PERIPH_BITBAND_ALIAS_TO_ADDR(fault_addr);
-            }
-            else if (fault_addr >= VMPU_SRAM_BITBAND_START && fault_addr <= VMPU_SRAM_BITBAND_END) {
-                fault_addr = VMPU_SRAM_BITBAND_ALIAS_TO_ADDR(fault_addr);
-            }
+    if (fault_addr == (uint32_t) &SCB->SCR) {
```
meriac: @niklas-arm (fixme is still missing!)

core/system/src/mpu/vmpu_freescale_k64_mem.c
AlessandroA: UP. Do we need the rounding here?

PR #295, opened 2016-07-22
Update uVisor RTOS docs publisher

meriac:
- Add Word template for rendering uVisor RTOS docs
- Add changes for publishing to local branch

@AlessandroA


AlessandroA:
:+1: LGTM

PR #307, opened 2016-08-02
Add initial clang-format file

meriac:
This file attempts to describe our coding standard in a way that
clang-format understands. It isn't perfect yet, but it is a nice
starting point.

See <http://clang.llvm.org/docs/ClangFormat.html> for usage
instructions, including how to integrate clang-format with your favorite
text editor.


AlessandroA:
👍

PR #308, opened 2016-08-10
Add initial RPC implementation

Patater:
This adds RPC sending and receiving, and the uVisor secure copying.

@niklas-arm @AlessandroA @meriac

### DIFF
--- a/api/inc/pool_queue_exports.h
+++ b/api/inc/pool_queue_exports.h
@@ -1,0 +1,197 @@
+/*
+ * Copyright (c) 2016, ARM Limited, All Rights Reserved
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef UVISOR_POOL_QUEUE_EXPORTS_H
+#define UVISOR_POOL_QUEUE_EXPORTS_H
+
+#include "api/inc/magic_exports.h"
+#include "api/inc/uvisor_exports.h"
+#include "api/inc/uvisor_semaphore_exports.h"
+#include "api/inc/uvisor_spinlock_exports.h"
+#include <stdint.h>
+#include <stddef.h>
+
+#define UVISOR_POOL_QUEUE_NON_BLOCKING (0)
+#define UVISOR_POOL_QUEUE_BLOCKING (1)
+
+#define UVISOR_POOL_SLOT_INVALID     ((uint8_t) 0xFFU)
+#define UVISOR_POOL_SLOT_IS_DEQUEUED ((uint8_t) 0xFEU)
+#define UVISOR_POOL_SLOT_IS_FREE     ((uint8_t) 0xFDU)
+#define UVISOR_POOL_MAX_VALID        ((uint8_t) 0xFCU)
+
+typedef uint8_t uvisor_pool_slot_t;
+
+typedef struct uvisor_pool_queue_entry {
+    union {
+        struct {
+            /* The next slot in the queue */
+            uvisor_pool_slot_t next;
+
+            /* The previous slot in the queue */
+            uvisor_pool_slot_t prev;
+        } queued;
+        struct {
+            /* If the slot is free, the next available slot in the free list */
+            uvisor_pool_slot_t next;
+
+            /* If the slot is free or dequeued */
+            uvisor_pool_slot_t state;
+        } dequeued;
+    };
+} uvisor_pool_queue_entry_t;
+
+typedef struct uvisor_pool {
+    /* Magic that identifies this as a uvisor_pool type. */
+    uint32_t magic;
+
+    /* The array holds slots of data. */
+    void const * array;
+
+    /* The distance between elements in the array. */
+    size_t stride;
+
+    /* The maximum number of elements that could be in the array. */
+    uvisor_pool_slot_t num;
+
+    /* Whether or not the queue can block callers who want to allocate slots
+     * from the pool. If non-zero, when no slots is available in the pool,
+     * callers will be blocked up to their timeout amount of time before giving
+     * up. */
+    int blocking;
+
+    /* The number of items currently allocated from the pool. For testing and
+     * debug purposes only. */
+    uvisor_pool_slot_t num_allocated;
+
+    /* The first free slot. */
+    uvisor_pool_slot_t first_free;
+
+    /* The semaphore is used to block allocations when the pool is full. */
+    UvisorSemaphore semaphore;
+
+    /* The spinlock serializes updates to the management array. */
+    UvisorSpinlock spinlock;
+
+    /* This must be at the end so we can allocate memory for pools by
+     * allocating enough room for the size of the pool appended by an array of
+     * entries. */
+    uvisor_pool_queue_entry_t management_array[];
+} uvisor_pool_t;
+
+/* These are assumed to only be statically allocated, so the management array
+ * in in-place. */
+typedef struct uvisor_pool_queue {
+    /* Magic that identifies this as a uvisor_pool_queue type. */
+    uint32_t magic;
+
+    /* The first allocated slot */
+    uvisor_pool_slot_t head;
+
+    /* The last allocated slot */
+    uvisor_pool_slot_t tail;
+
+    uvisor_pool_t * pool;
+} uvisor_pool_queue_t;
+
+/* Intialize a pool.
+ * Return 0 on success, non-zero otherwise. */
+UVISOR_EXTERN int uvisor_pool_init(uvisor_pool_t * pool, void * array, size_t stride, size_t num, int blocking);
+
+/* Initialize a pool queue.
+ * Return 0 on success, non-zero otherwise. */
+UVISOR_EXTERN int uvisor_pool_queue_init(uvisor_pool_queue_t * pool_queue, uvisor_pool_t * pool, void * array, size_t stride, size_t num, int blocking);
+
+/* Allocate a slot from the pool. If the pool has no more slots available,
+ * block up to the specified length of time in milliseconds. No blocking will
+ * occur if the timeout is zero or the pool was initialized as non-blocking.
+ * This doesn't put anything in the slot for you. It's up to you to do that.
+ * Return the index of the allocated slot, or UVISOR_POOL_SLOT_INVALID if
+ * timed out waiting for an available slot. This function will spin until the
+ * spin lock serializing access to the pool can be taken. */
+UVISOR_EXTERN uvisor_pool_slot_t uvisor_pool_allocate(uvisor_pool_t * pool, uint32_t timeout_ms);
+/* Attempt to allocate a slot. This function will fail if the spin lock
+ * serializing access to the pool can not be taken. */
+UVISOR_EXTERN uvisor_pool_slot_t uvisor_pool_try_allocate(uvisor_pool_t * pool);
+
+/* Enqueue the specified slot into the queue. */
+UVISOR_EXTERN void uvisor_pool_queue_enqueue(uvisor_pool_queue_t * pool_queue, uvisor_pool_slot_t slot);
+UVISOR_EXTERN int uvisor_pool_queue_try_enqueue(uvisor_pool_queue_t * pool_queue, uvisor_pool_slot_t slot);
+
+/* Free the specified slot back into the pool. Invalid slots are ignored.
+ * Return the slot that was freed, or UVISOR_POOL_SLOT_IS_FREE if the slot was
+ * already freed, or UVISOR_POOL_SLOT_INVALID if the slot being requested to
+ * free is outside the range of the queue. */
+UVISOR_EXTERN uvisor_pool_slot_t uvisor_pool_free(uvisor_pool_t * pool, uvisor_pool_slot_t slot);
+UVISOR_EXTERN uvisor_pool_slot_t uvisor_pool_try_free(uvisor_pool_t * pool, uvisor_pool_slot_t slot);
+
+/* Remove the specified slot from the queue. This function does not free the
+ * specified slot back into the pool. Return the slot that was dequeued, or
+ * UVISOR_POOL_SLOT_IS_DEQUEUED if the slot was already dequeued, or
+ * UVISOR_POOL_SLOT_INVALID if the slot being requested to dequeue is outside
+ * the range of the queue. */
+UVISOR_EXTERN uvisor_pool_slot_t uvisor_pool_queue_dequeue(uvisor_pool_queue_t * pool_queue, uvisor_pool_slot_t slot);
+
+/* Remove the first slot from the queue. This function does not free the
+ * specified slot back into the pool. Return the slot that was dequeued or
+ * UVISOR_POOL_SLOT_INVALID if the slot being requested to dequeue is outside
+ * the range of the queue. */
+UVISOR_EXTERN uvisor_pool_slot_t uvisor_pool_queue_dequeue_first(uvisor_pool_queue_t * pool_queue);
+UVISOR_EXTERN uvisor_pool_slot_t uvisor_pool_queue_try_dequeue_first(uvisor_pool_queue_t * pool_queue);
+
+/* Find the first (in queue order) slot that the supplied query function
+ * returns non-zero for. The query function is provided with `context` on every
+ * invocation. This allows query functions to access additional data without
+ * having to use global variables. `uvisor_pool_queue_find_first` is reentrant. */
+typedef int (*TQueryFN_Ptr)(uvisor_pool_slot_t slot, void * context);
+UVISOR_EXTERN uvisor_pool_slot_t uvisor_pool_queue_find_first(uvisor_pool_queue_t * pool_queue,
+                                                              TQueryFN_Ptr query_fn, void * context);
+
+/* Inline helper function to make allocating slots for pool queues easier and
+ * better encapsulated (clients don't need to pull the pool out of the pool
+ * queue, or even realize pool_queue is implemented with a pool) */
+static inline uvisor_pool_slot_t uvisor_pool_queue_allocate(uvisor_pool_queue_t * pool_queue, uint32_t timeout_ms)
+{
+    return uvisor_pool_allocate(pool_queue->pool, timeout_ms);
+}
+
+static inline uvisor_pool_slot_t uvisor_pool_queue_try_allocate(uvisor_pool_queue_t * pool_queue)
+{
+    return uvisor_pool_try_allocate(pool_queue->pool);
+}
+
+/* Inline helper function to make freeing slots for pool queues easier and
+ * better encapsulated (clients don't need to pull the pool out of the pool
+ * queue, or even realize pool_queue is implemented with a pool) */
+static inline uvisor_pool_slot_t uvisor_pool_queue_free(uvisor_pool_queue_t * pool_queue, uvisor_pool_slot_t slot)
+{
+    return uvisor_pool_free(pool_queue->pool, slot);
+}
+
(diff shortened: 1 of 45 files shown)


api/inc/rpc_exports.h
```diff
@@ -0,0 +1,132 @@
...
+    /* The semaphore to wait on for this function group */
+    UvisorSemaphore semaphore;
+} uvisor_rpc_fn_group_t;
+
+#define UVISOR_RPC_OUTGOING_MESSAGE_SLOTS (8)
+
+#define UVISOR_RPC_INCOMING_MESSAGE_SLOTS (8)
+#define UVISOR_RPC_OUTGOING_RESULT_SLOTS (8)
+
+#define UVISOR_RPC_FN_GROUP_SLOTS (8)
```
niklas-arm: Should these macros be overridable by passing this macro definition via the compiler?

api/inc/rpc_exports.h
```diff
@@ -0,0 +1,132 @@
...
+#include "api/inc/pool_queue_exports.h"
+#include "api/inc/uvisor_semaphore_exports.h"
+
+typedef uint32_t (*TFN_Ptr)(uint32_t, uint32_t, uint32_t, uint32_t);
+
+#define UVISOR_RESULT_SLOT_BITS 10
+#define UVISOR_RESULT_SLOT_MASK ((1 << UVISOR_RESULT_SLOT_BITS) - 1)
+
+#define UVISOR_RESULT_COUNTER_MASK (0xFFFFFFFFUL << UVISOR_RESULT_SLOT_BITS)
+#define UVISOR_RESULT_COUNTER_INCREMENT (2 << UVISOR_RESULT_SLOT_BITS)
```
niklas-arm: Maybe add documentation on why we choose `2` and not `1`, ie. on the invalidation mechanism of the cookie, that preserves the slot.

api/inc/rpc.h
```diff
@@ -29,10 +30,6 @@
 /* FIXME This is a dummy implementation. */
 #define UVISOR_BOX_RPC_MAX_INCOMING(max_num_incoming_rpc)
```
niklas-arm: Exposing this dummy API seems wrong to me, because

1. The max incoming PRC message count is fixed for all boxes at the moment.
2. This API is unbalanced, I'd expect to be able to define the `MAX_OUTGOING` message as well.

For the sake of future compatibility I'd like to add a `UVISOR_BOX_RPC_MAX_OUTGOING` macro as well.

api/inc/vmpu_exports.h
```diff
@@ -189,6 +219,11 @@ typedef struct {
      * This is set to `NULL` by uVisor, signalling to the user lib that the
      * box heap needs to be initialized before use! */
     void * active_heap;
+
+    /* Box IDs */
+    int box_id_self;
+    int box_id_caller;
```
niklas-arm: This isn't thread-safe!

If I have two threads doing `rpc_fncall_waitfor` and they get scheduled during an RPC call, but before `uvisor_box_id_caller()` has been called, the `box_id_caller` may be overwritten by the RPC call serviced by the other thread. This variable would have to be moved into thread local storage.

api/src/rpc.c
```diff
@@ -0,0 +1,367 @@
...
+    return (uvisor_rpc_fn_group_t *) fn_group_pool()->array;
+}
+
+/* Place a message into the outgoing queue. `timeout_ms` is how long to wait
+ * for a slot in the outgoing queue before giving up. `msg_slot` is set to the
+ * slot of the message that was allocated. Returns non-zero on failure. */
+static int send_outgoing_rpc(uint32_t p0, uint32_t p1, uint32_t p2, uint32_t p3, const TFN_Ptr fn, uint32_t timeout_ms,
+                             uvisor_rpc_result_t * cookie)
+{
+    static uint32_t shared_result_counter = 0;
```
niklas-arm: This shared counter is not in protected memory. Any box can write to it and influence the counter value that another box receives in the cookie (like writing `0` in a `while(1)` loop).

core/system/src/pool_queue.c
```diff
@@ -0,0 +1,353 @@
...
+ * limitations under the License.
+ */
+#include "api/inc/pool_queue_exports.h"
+#include "semaphore.h"
+#include "spinlock.h"
+#include <string.h>
+
+int uvisor_pool_init(uvisor_pool_t * pool, void * array, size_t stride, size_t num, int blocking)
+{
+    uvisor_pool_slot_t i;
```
niklas-arm:
```c
if (num >= UVISOR_POOL_MAX_VALID) {
    return -1;
}
```

Patater: Fixed

api/src/rpc.c
```diff
@@ -0,0 +1,367 @@
...
+    msg->p3 = p3;
+    msg->function = fn;
+    msg->cookie = uvisor_result_build(counter, slot);
+
+    /* Put the slot into the queue. */
+    uvisor_pool_queue_enqueue(outgoing_message_queue(), slot);
+
+    /* Notify the caller of this function of the slot that was allocated for
+     * sending this RPC message. */
+    *cookie = msg->cookie;
```
Patater: C is for Cookie
That's good enough for me

api/inc/vmpu_exports.h
Patater: Fixed

api/inc/rpc_exports.h
Patater: Not for now

api/inc/rpc.h
Patater: The macro has been deleted.

api/src/uvisor-lib.c
```diff
@@ -19,21 +19,14 @@
 
 int uvisor_lib_init(void)
 {
-    /* Supported */
-    /* Defined in uvisor-input.S */
-    extern uint32_t uvisor_config;
-    extern uint32_t uvisor_export_table_size;
+    TUvisorExportTable * export_table = uvisor_export_table();
```
meriac: @Patater use of const as we only read from the table? Consider returning a const from uvisor_export_table?

api/src/uvisor-lib.c
```diff
@@ -19,21 +19,14 @@
...
-    if (uvisor_export_table->magic != UVISOR_EXPORT_MAGIC) {
+    if (export_table->magic != UVISOR_EXPORT_MAGIC) {
         /* We couldn't find the magic. */
         return -1;
     }
 
-    if (uvisor_export_table->version != UVISOR_EXPORT_VERSION) {
+    if (export_table->version != UVISOR_EXPORT_VERSION) {
         /* The version we understand is not the version we found. */
         return -1;
```
meriac: @Patater Back return codes by generic #defines ?

api/rtx/src/uvisor_semaphore.c
```diff
@@ -0,0 +1,43 @@
...
+int __uvisor_semaphore_init(UvisorSemaphore * s, int32_t count)
+{
+    uvisor_semaphore_internal_t * semaphore = (uvisor_semaphore_internal_t *) s;
+
+    memset(semaphore->data, 0, sizeof(semaphore->data));
+    semaphore->def.semaphore = semaphore->data;
+    semaphore->id = osSemaphoreCreate(&semaphore->def, count);
+
+    /* Error when semaphore->id is NULL */
+    return -(semaphore->id == NULL);
```
meriac: @Patater please use global error codes.

api/rtx/src/uvisor_semaphore.c
```diff
@@ -0,0 +1,43 @@
...
+
+int __uvisor_semaphore_pend(UvisorSemaphore * s, uint32_t timeout_ms)
+{
+    uvisor_semaphore_internal_t * semaphore = (uvisor_semaphore_internal_t *) s;
+
+    int32_t num_available_tokens = osSemaphoreWait(semaphore->id, timeout_ms);
+
+    if (num_available_tokens == -1 || num_available_tokens == 0) {
+        /* Invalid parameters or no tokens available */
+        return -1;
```
meriac: (error code)

api/src/box_id.c
```diff
@@ -0,0 +1,31 @@
...
+
+int uvisor_box_id_self(void)
+{
+    return __uvisor_ps->box_id_self;
+}
+
+int uvisor_box_id_caller(void)
+{
+    /* XXX Not implemented. */
+    return -1;
```
meriac: @Patater if a global not_implemented error code doesn't exist, please define one.

core/system/src/export_table.c
```diff
@@ -145,5 +145,15 @@ const TUvisorExportTable __uvisor_export_table = {
         .thread_destroy = thread_destroy,
         .thread_switch = thread_switch,
     },
+    .pool_queue = {
+        .pool_init = uvisor_pool_init,
```
meriac: @Patater : in the pool_queue - istn't it over the top to use pool_* prefixes for each member ?

api/src/uvisor-lib.c
Patater: Fixed

api/src/uvisor-lib.c
Patater: Fixed

api/rtx/src/uvisor_semaphore.c
Patater: Fixed.

api/rtx/src/uvisor_semaphore.c
Patater: Fixed

api/src/box_id.c
Patater: Fixed

Patater:
Addressed most of the XXX comments.

api/inc/rpc_exports.h
Patater: Fixed

core/system/src/export_table.c
```diff
@@ -83,16 +95,446 @@ static void thread_destroy(void * c)
...
+
+    /* The callee box ID associated with the gateway must be valid. Gateways to
+     * box 0 are not possible. */
+    int box_id = callee_box_id(gateway);
+    if (box_id <= 0 || box_id >= g_vmpu_box_count) {
+        return 0;
+    }
+
+    /* Gateway needs to have good magic. */
+    if (!(gateway->magic == UVISOR_RPC_GATEWAY_MAGIC_ASYNC || gateway->magic == UVISOR_RPC_GATEWAY_MAGIC_SYNC)) {
```
meriac: @Patater Make sure to verify fist the magic before you verify members - line 184 with callee_box_id.

core/system/src/export_table.c
```diff
@@ -83,16 +95,446 @@ static void thread_destroy(void * c)
...
+
+        slot = fn_group_queue->pool->management_array[slot].queued.next;
+    }
+
+    return num_posted;
+}
+
+static int callee_box_id(const TRPCGateway * gateway)
+{
+    return (uint32_t *)gateway->box_ptr - __uvisor_config.cfgtbl_ptr_start;
```
meriac: @Patater Make sure that you only return valid box ID's within start/end of the box-id pointer list - also ensure alignment of 4. Make sure to return a invalid ID if the pointer is outside of the box list.

core/system/src/export_table.c
```diff
@@ -83,16 +95,446 @@ static void thread_destroy(void * c)
...
+/* Return true if and only if the queue is entirely within the box specified by
+ * the provided box_id. */
+static int is_valid_queue(uvisor_pool_queue_t * queue, int box_id)
+{
+    uint32_t bss_start = g_context_current_states[box_id].bss;
+    uint32_t bss_end = bss_start + g_context_current_states[box_id].bss_size;
+
+    uint32_t queue_start = (uint32_t) queue;
+    uint32_t queue_end = queue_start + sizeof(*queue);
+    int queue_is_valid = queue_start >= bss_start && queue_end <= bss_end;
```
meriac: @Patater consider adding a magic value check as additional precaution.

meriac: Isn't the offset within the context deterministic ? If yes, please compare for the exact offset depending on the box id.

core/system/src/export_table.c
```diff
@@ -83,16 +95,446 @@ static void thread_destroy(void * c)
...
+    int pool_is_valid = pool_start >= bss_start && pool_end <= bss_end;
+    if (!pool_is_valid) {
+        return 0;
+    }
+
+    /* XXX A malicious box could lie about its own pool size to make it pretend to
+     * fit within box bss. */
+    uint32_t man_array_start = (uint32_t) queue->pool->management_array;
+    uint32_t man_array_end = man_array_start + sizeof(*queue->pool->management_array) * queue->pool->num;
+    int man_array_is_valid = man_array_start >= bss_start && man_array_end <= bss_end;
```
meriac: @Patater Please add brackets for readability.

Patater: Done

core/system/src/export_table.c
Patater: Done

api/inc/pool_queue_exports.h
```diff
@@ -0,0 +1,201 @@
...
+
+/* Enqueue the specified slot into the queue. */
+UVISOR_EXTERN void uvisor_pool_queue_enqueue(uvisor_pool_queue_t * pool_queue, uvisor_pool_slot_t slot);
+UVISOR_EXTERN int uvisor_pool_queue_try_enqueue(uvisor_pool_queue_t * pool_queue, uvisor_pool_slot_t slot);
+
+/* Free the specified slot back into the pool. Invalid slots are ignored.
+ * Return the slot that was freed, or UVISOR_POOL_SLOT_IS_FREE if the slot was
+ * already freed, or UVISOR_POOL_SLOT_INVALID if the slot being requested to
+ * free is outside the range of the queue. */
+UVISOR_EXTERN uvisor_pool_slot_t uvisor_pool_free(uvisor_pool_t * pool, uvisor_pool_slot_t slot);
```
meriac: @Jaeden - conceptually: if there's only one queue now for each box context: do we need to pass the `pool` parameter in all of these functions, or can it be derived from context instead?

Patater: There are multiple queues per box context: incoming RPC message queue and outgoing result queue.

api/src/pool_queue.c
```diff
@@ -0,0 +1,59 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "api/inc/export_table_exports.h"
+#include "api/inc/pool_queue_exports.h"
+
+int uvisor_pool_init(uvisor_pool_t * pool, void * array, size_t stride, size_t num, int blocking)
+{
+    return uvisor_export_table()->pool_queue.pool_init(pool, array, stride, num, blocking);
```
meriac: @jaeden - consider storing `uvisor_export_table()->pool_queue` into a user mode pointer that is used directly and initialised during user mode init.

api/src/rpc.c
```diff
@@ -0,0 +1,369 @@
...
+} query_for_fn_group_context_t;
+
+static int query_for_fn_group(uvisor_pool_slot_t slot, void * context)
+{
+    query_for_fn_group_context_t * query_context = context;
+    uvisor_rpc_message_t * msg = &incoming_message_array()[slot];
+    size_t i;
+
+    /* See if the target is for a function we can handle. */
+    for (i = 0; i < query_context->fn_count; ++i) {
```
meriac: @Patater : Please add a FIXME with request for optimisation of this scan (hash table etc.).

Patater: Done

api/src/rpc.c
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    msg = &incoming_message_array()[msg_slot];
+
+    /* Save the calling box ID before the RPC target function is called, so that the RPC target function
+     * can read that variable to find out what box called it. */
+    if (box_id_caller) {
+        *box_id_caller = msg->other_box_id;
+    }
+    TFN_Ptr fn = (TFN_Ptr) msg->gateway->target;
```
meriac: @Patater pure paranoia for detecting bugs rather than hacks / memory corruption: run a sanity check whether the gateway is valid (normal gw check including magic etc.)

core/system/src/export_table.c
meriac: Also got through all your remaining XXX'es - either turn them into FIXME's or remove them/fix them.

core/system/src/export_table.c
```diff
@@ -83,16 +95,446 @@ static void thread_destroy(void * c)
...
+    if (!man_array_is_valid) {
+        return 0;
+    }
+
+    /* XXX A malicious box could lie about its own pool size to make it pretend to
+     * fit within box bss. */
+    uint32_t array_start = (uint32_t) queue->pool->array;
+    uint32_t array_end = array_start + queue->pool->stride * queue->pool->num;
```
meriac: Please add brackets for feadability @Patater

core/system/src/export_table.c
```diff
@@ -83,16 +95,446 @@ static void thread_destroy(void * c)
...
+    int array_is_valid = array_start >= bss_start && array_end <= bss_end;
```
meriac: ...same here

core/system/src/export_table.c
Patater: I don't think the magic is really necessary. We can use the queue directly from the box config table instead of from the box context pointer. The box cfg magic is already checked. We can avoid these sorts of errors from that box cfg magic check. We also don't have to trust the box context pointer to the queue, although we still need to validate the queue itself.

core/system/src/mpu/vmpu.c
```diff
@@ -189,30 +189,38 @@ static void vmpu_sanity_check_box_namespace(int box_id, const char *const box_na
...
-    void * box_bss;
+    uint8_t * box_bss;
     UvisorBoxIndex * index;
     uint32_t heap_size = config->heap_size;
+    int i;
 
     if (box_id == 0) {
         /* Box 0 still uses the main heap to be backwards compatible. */
-        heap_size = ((void *) __uvisor_config.heap_end - (void *) __uvisor_config.heap_start) - config->index_size;
+        heap_size = ((uint8_t *) __uvisor_config.heap_end - (uint8_t *) __uvisor_config.heap_start) - config->index_size;
```
meriac: @Patater cast to uint32_t instead as the result is uint32_t, too ?

core/system/src/export_table.c
Patater: Nevermind. The offset within the context is not the same between different versions of uvisor-lib. The RtxBoxIndex (which is located before the RPC queues), could be of any size. The RPC queues themselves may also be application-defined in size in the future. So, the offset of the RtxBoxIndex would be variable if we moved the RPC queues to before the RtxBoxIndex.

I didn't add any additional check here, because I don't currently see a programmatic way to determine the offset. I recommend we create an issue (to fix later) if we want an additional check here.

meriac:
@Patater @Alessandro @niklas-arm Other than the comments above I am fine with the API. A general comment for your own reviews:

Please make sure to verify in top level user-facing functions, whether the provided structure pointers are valid (magic value etc?).

api/src/pool_queue.c
Patater: What is the value of storing the pointer in another place?

I don't see the benefit, but I might be missing something. The export table is already storing the pointer and the pointer can be read from user mode.

core/system/src/export_table.c
Patater: Removed `pool` prefix from the pool table

core/system/src/export_table.c
Patater: Removed all XXX

api/src/rpc.c
Patater: Added a check

core/system/src/export_table.c
Patater: Done

core/system/src/export_table.c
Patater: Done

core/system/src/mpu/vmpu.c
Patater: Either way is fine. Changed to uint32_t.

Patater:
Feedback is all addressed

sherrellbc:
@Patater 

Can this API be used from the public box? The semantics of `UVISOR_BOX_RPC_GATEWAY_SYNC` require the _box_name_. Does the public/main box have a name?

Patater:
@sherrellbc No, this API can't be used to RPC into the public box. Gateways into the public box can't be created.

PR #311, opened 2016-08-23
Add uvisor support to Beetle

fvincenzo:
This PR contains two patches and adds the support for Beetle to uvisor.


core/debug/src/debug.c
```diff
@@ -30,6 +30,41 @@ uint8_t g_buffer[DEBUG_MAX_BUFFER];
 int g_buffer_pos;
 
 #ifndef CHANNEL_DEBUG
```
meriac: @fvincenzo please move all the beetle-specific code to a dedicated debug_beetle.c file.

core/debug/src/debug.c
```diff
@@ -30,6 +30,41 @@ uint8_t g_buffer[DEBUG_MAX_BUFFER];
...
+}
+
+void default_putc(uint8_t data)
+{
+    if(beetle_uart1_initialized == 0)
+        beetle_uart1_initialize();
+
+    char c = (char) data;
+
+    while(beetle_uart1_writable());
```
meriac: @fvincenzo name looks wrong: I assume you spin while *not* writable.

core/debug/src/debug.c
```diff
@@ -30,6 +30,41 @@ uint8_t g_buffer[DEBUG_MAX_BUFFER];
...
 #ifndef CHANNEL_DEBUG
+#ifdef CORE_BEETLE
+#define UART1       0x40005000
+#define UART1DATA   (UART1)
+#define UART1STATE  (UART1 + 0x4)
+#define UART1CTRL   (UART1 + 0x8)
+#define UART1BDIV   (UART1 + 0x10)
+
+#define beetle_uart1_readl(reg) *(volatile uint32_t *)reg
+#define beetle_uart1_writel(reg, val)  *(uint32_t *)reg = val;
```
meriac: @fvincenzo no good reason to use macros here: can you please use static inlines instead?

meriac:
@AlessandroA please review, too.

core/debug/src/debug.c
fvincenzo: ok

core/debug/src/debug.c
fvincenzo: Done :smile:

core/debug/src/debug_beetle.c
```diff
@@ -0,0 +1,67 @@
...
+#define UART1BDIV   (UART1 + 0x10)
+
+static inline uint32_t beetle_uart1_readl(uint32_t reg)
+{
+    return *(volatile uint32_t *)reg;
+}
+
+static inline void beetle_uart1_writel(uint32_t reg, uint32_t val)
+{
+    *(uint32_t *)reg = val;
```
meriac: @fvincenzo "volatile" missing here?

meriac: @fvincenzo form the looks of it you would be better off simply declaring a struct and casting it to the absolute offset. Each struct member would be declared with _IO _R _W as seen in the CMSIS headers - which would solve the "volatile"- problems.

meriac: @fvincenzo (a UART struct, casted to be at UART1 position)

platform/beetle/inc/config.h
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef __CONFIG_H__
+#define __CONFIG_H__
+
+/* uVisor stack configuration */
+#define STACK_SIZE       2048
```
AlessandroA: @fvincenzo These 2 lines could be removed. Could you please verify that it builds fine without them? (They should all have a default value).

AlessandroA:
@fvincenzo For the `debug_beetle.c` file what I suggest to do is to put it into `platform/beetle/src/debug_beetle.c`. If I remember correctly the build system si already able to detect platform-specific code (let me know if it doesn't).

Once the code is in the Beetle-specific folder, you remove the use of the `CORE_BEETLE` symbol from the whole code-base (in `debug.c`, in the configuration file, and in the `debug_beetle.c` file).

fvincenzo:
@AlessandroA 
I modified the PR based on your suggestions and rebased on the latest main. Unfortunately if I build in this condition uvisor stops working.
```
uVisor mode: 2
uvisor_ram : @0x20000140 (7872 bytes) [config]
             @0x20000140 (7872 bytes) [linker]
bss_boxes  : @0x20002000 (8320 bytes) [config]

box stack segment start=0x20002000 end=0x20002000 (length=8320)
MPU.REGIONS=8
MPU.ALIGNMENT=0x0000001F
MPU.ALIGNMENT_BITS=5
HALT_ERROR(./core/system/src/mpu/vmpu_armv7m_mpu.c#178): inferred ACL's (0x003F) don't match exptected ACL's (0x0037)
```

This issue seems not depending from my code.

AlessandroA:
@fvincenzo I can confirm this issue was already observed by @niklas-arm in his port to the EFM32GG_STK3700 module. I'll raise an issue here and verify the ACLs independently from this PR. Thanks for the update!

Issue #315, opened 2016-08-25
uVisor prevents LwIP/Hal from properly using ENET hardware when enabled


sherrellbc:
**EDIT:** I changed the issue's name to more appropriately track the purpose of this thread after relevant information has been uncovered surrounding the original problem. 

---

I have been trying to determine the appropriate uVisor ACL such that I can use the Ethernet hardware on the FRDM K64F development board. After a few hours of debugging and tracing the problem I managed to determine two points of failure when using EthernetInterface.cpp and (a dependency thereof) the LWIP implementation; the root of the corresponding source tree can be found [here](https://github.com/ARMmbed/mbed-os/tree/master/features/net/FEATURE_IPV4/lwip-interface).

The first problem I found was that the LWIP implementation was trying to disable the MPU, which I thought might be problematic for uVisor.

`MPU->CESR &= ~MPU_CESR_VLD_MASK;`

https://github.com/ARMmbed/mbed-os/blob/master/features/net/FEATURE_IPV4/lwip-interface/lwip-eth/arch/TARGET_Freescale/hardware_init_MK64F12.c#L41

I have defined the MPU as an enabled memory range in my ACL, but that does not prevent the system from stopping on that instruction. 

```
const UvisorBoxAclItem g_main_acl[] = {     
         {SIM,    sizeof(*SIM),    UVISOR_TACLDEF_PERIPH}, 
         {OSC,    sizeof(*OSC),    UVISOR_TACLDEF_PERIPH}, 
         {MCG,    sizeof(*MCG),    UVISOR_TACLDEF_PERIPH}, 
         {PORTA,  sizeof(*PORTA),  UVISOR_TACLDEF_PERIPH}, 
         {PORTB,  sizeof(*PORTB),  UVISOR_TACLDEF_PERIPH}, 
         {PORTC,  sizeof(*PORTC),  UVISOR_TACLDEF_PERIPH}, 
         {PORTD,  sizeof(*PORTD),  UVISOR_TACLDEF_PERIPH}, 
         {PORTE,  sizeof(*PORTE),  UVISOR_TACLDEF_PERIPH}, 
         {RTC,    sizeof(*RTC),    UVISOR_TACLDEF_PERIPH}, 
         {LPTMR0, sizeof(*LPTMR0), UVISOR_TACLDEF_PERIPH}, 
         {PIT,    sizeof(*PIT),    UVISOR_TACLDEF_PERIPH}, 
[...]

PR #320, opened 2016-09-02
Allow hardware specific region-ACL extensions

meriac:
- Enables hardware specific region-ACL extensions 
- Move subregion configuration into HW-specific ACLs:
  - For K64F DMA Busmaster overrides are possible
  - For ARMv7-M Subregions can be defined- Enables access to box 0 memories by the Ethernet peripheral
- Background region privileges reduced to Core and Debugger

@AlessandroA


core/system/src/mpu/vmpu_freescale_k64_mem.c
```diff
@@ -63,21 +63,27 @@ void vmpu_mem_switch(uint8_t src_box, uint8_t dst_box)
...
+     * - allow execution
+     * - give read access to ENET DMA bus master */
     vmpu_mpu_set_static_acl(1, FLASH_ORIGIN, (uint32_t) __uvisor_config.secure_end - FLASH_ORIGIN,
         UVISOR_TACL_SREAD |
         UVISOR_TACL_SEXECUTE |
         UVISOR_TACL_UREAD |
         UVISOR_TACL_UEXECUTE |
-        UVISOR_TACL_USER);
+        UVISOR_TACL_USER,
+        0x4UL<<18);
```
AlessandroA: Could you add spaces in `0x4UL << 18)`? (also in the other occurrence below)

BTW, do we want to provide an interface/macro for selecting a certain bus master or to set the K64F-specific ACLs? The hardcoded numbers are not very user-friendly. If we don't want to do it now we can add and issue/story.

meriac:
@AlessandroA I fixed the spacing problem. I would like to avoid though macros for now for the hw-specific ACLs, as they are currently only used in this two lines plus I don't want to make this "as thing".

core/system/src/mpu/vmpu_freescale_k64_mpu.c
```diff
@@ -70,11 +70,17 @@ uint32_t vmpu_round_up_region(uint32_t addr, uint32_t size)
...
-static uint32_t vmpu_map_acl(UvisorBoxAcl acl)
+static uint32_t vmpu_map_acl(UvisorBoxAcl acl, uint32_t acl_hw_spec)
 {
     /* handle user permissions */
     uint32_t perm = (acl & UVISOR_TACL_USER) ? (acl & UVISOR_TACL_UACL) : 0;
 
+    /* Allow uVisor-internal code to extend permissions for certain
+     * ACLs by hardware-specific ACLs like DMA busmaster access.
+     * Ensure that core ACLs are specified trough the generic ACL
+     * method b masking it out. */
```
AlessandroA:
```diff
-    * method b masking it out. */
+    * method by masking it out. */
```

AlessandroA:
LGTM 👍

sherrellbc:
DMA confirmed to work (see https://github.com/ARMmbed/uvisor/issues/315) to/from box 0 using LwIP as a test case (the original motivation for this patch). Other problems related to managing mutually exclusive memory (mutex) at run-time have now manifested as either a side-effect of this patch or as something previously unknown and un-tested. See the above quoted git issue for more information.

PR #334, opened 2016-09-20
Virtualize NVIC_SystemReset

AlessandroA:
This PR virtualizes the `NVIC_SystemReset` API.

> **Note**: The API can be used only by the debug box.

Upon merge, we need to update mbed OS and the version of CMSIS 5 it contains (there is a PR ready for that).

I'm also proposing the following change: The `halt_error` handler in a debug box should not trigger a reset automatically, as it is counterintuitive. Instead, uVisor will halt upon return.

The debug box can still call `NVIC_SystemReset` if it wants to.

@meriac @Patater @niklas-arm

### DIFF
--- a/api/inc/cmsis_nvic_virtual.h
+++ b/api/inc/cmsis_nvic_virtual.h
@@ -18,6 +18,7 @@
 #define __UVISOR_API_NVIC_VIRTUAL_H__
 
 #include "api/inc/interrupts.h"
+#include "api/inc/unvic_exports.h"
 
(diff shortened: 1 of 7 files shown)


Patater:
Since halt_error is going back to the former behavior of not returning, we could also restore the `__attribute__ ((__noreturn__))` to its declaration that was removed when we added the debug box feature.

Patater:
I don't see the reason for adding `vIRQ: Add reason to vIRQ_SystemReset` right now. Consider dropping the commit.

Issue #326, opened 2016-09-15
Interrupts cause memFault


asensaoui:
Sometimes, interrupts cause a Memmanag fault, not everytime, so i create a small program with only one interruption which is a button (sw1), here is the link to the source files : https://github.com/asensaoui/uvisor_interrupts 
( I ported my own board, a Cortex M3, it's not officially supported, and also I changed all the NVIC to __NVIC and made some modification to core_cm3.h (adding the CMSIS_NVIC_VIRTUAL ...) I've also changed this to bypass the bug within the initialization : 
`    /* Unlock the upper SRAM subregion only. */
    vmpu_mpu_set_static_acl(
        1,
        (uint32_t) __uvisor_config.sram_start,
        total_size,
        UVISOR_TACLDEF_DATA | UVISOR_TACL_EXECUTE,
        subregions_disable_mask
    );`
)

Here is the output of the debugger : 
396,635 @"***********************************************************\n"
396,648 @"                    MEMMANAGE FAULT\n"
396,680 @"***********************************************************\n"
396,693 @"\n"
396,726 @"* CFSR  : 0x00000001\n"
396,739 @"\r\n"
396,752 @"\r* MMFAR : 0xE000ED34\n"
396,766 @"\r\n"
396,779 @"\r* No MPU violation found\n"
396,792 @"\r\n"
396,805 @"\r* MEMORY MAP\n"
396,818 @"\r  Address:           0xE000ED34\n"
396,850 @"\r  Region/Peripheral: [not available]\n"
396,864 @"\r    Base address:    0xE000ED34\n"
396,896 @"\r    End address:     0xE000ED34\n"
396,909 @"\r    Before bitband:  [invalid]\n"
396,922 @"\r\n"
396,954 @"\r* EXCEPTION STACK FRAME\n"
396,967 @"  Exception from unprivileged code\n"
397,000 @"    psp:     0x2000A2A8\n"
397,033 @"\r    lr:      0xFFFFFFFD\n"
397,046 @"\r  Exception stack frame:\n"
397,059 @"    psp[08]: 0x2000A6BC | align\n"
397,072 @"    psp[07]: 0x00000000 | xPSR\n"
397,104 @"    psp[06]: 0x00000000 | pc\n"
397,117 @"    psp[05]: 0x1000BFCD | lr\n"
397,149 @"    psp[04]: 0x10001E9C | r12\n"
397,162 @"    psp[03]: 0x10009327 | r3\n"
[...]

PR #335, opened 2016-09-20
Fix page heap alignment issues and ISR stack pointer issues

niklas-arm:
This fixes the alignment issues on the page heap MPU region with the introduction of an additional shift.

This also fixes one issue with the ISR stack pointer on v7-M MPU on my experimental EFM32GG port, as described in #326.
However, I still get Hardfaults for an unstack error within 10mins, but only if IRQ2 (used by `us_ticker`) is registered. Other IRQs including GPIO interrupts do not trigger this behavior.

### DIFF
--- a/core/system/src/mpu/vmpu_armv7m.c
+++ b/core/system/src/mpu/vmpu_armv7m.c
@@ -229,7 +229,7 @@ static int vmpu_mem_push_page_acl_iterator(uint8_t mask, uint8_t index)
     uint32_t size = g_page_size * 8;
     vmpu_region_translate_acl(
         &region,
-        ((uint32_t) __uvisor_config.page_end - size * (index + 1)),
+        (g_page_head_end_rounded - size * (index + 1)),
         size,
         UVISOR_TACLDEF_DATA | UVISOR_TACL_EXECUTE,
         ~mask
@@ -478,4 +478,28 @@ void vmpu_arch_init_hw(void)
         UVISOR_TACLDEF_DATA | UVISOR_TACL_EXECUTE,
         subregions_disable_mask
     );
+
+    /* On page heap alignments:
+     *
+     * Individual pages in the page heap are protected by subregions.
+     * A page of size 2^N must have its start address aligned to 2^N.
+     * However, for page sizes > 1/8th region size, the start address is
+     * not guaranteed to be aligned to 2^N.
+     *
+     * Example: 2^N = page size, 2^(N-1) = 1/8th SRAM (32kB page size in a 128kB SRAM).
+     *
+     * |           |
+     * +-----------+ <-- uvisor_config.page_end: 0x30000 = 3/8th * 128kB is aligned to 16kB.
+     * | 32kB page |
+     * +-----------+ <-- page start address: 0x30000 - 32kB = 0x10000 is not aligned to 32kB!!
+     * |           |
+     *
+     * Due to these contradicting alignment requirements, it is not possible
+     * to have a page size larger than 1/8th region size.
+     */
+    if (subregions_size < *__uvisor_config.page_size) {
+        HALT_ERROR(SANITY_CHECK_FAILED,
+                   "The page size (%ukB) must not be larger than 1/8th of SRAM (%ukB).",
+                   *__uvisor_config.page_size / 1024, subregions_size / 1024);
+    }
 }
--- a/core/system/src/mpu/vmpu_armv7m.c
+++ b/core/system/src/mpu/vmpu_armv7m.c
@@ -359,8 +359,10 @@ void vmpu_acl_stack(uint8_t box_id, uint32_t bss_size, uint32_t stack_size)
 
     /* allocate context pointer */
     g_context_current_states[box_id].bss = slots_ctx ? box_mem_pos : (uint32_t) NULL;
-    /* ensure stack band on top for stack underflow dectection */
-    g_context_current_states[box_id].sp = (box_mem_pos + size - UVISOR_STACK_BAND_SIZE);
+    /* `(box_mem_pos + size)` is already outside the memory protected by the
+     * MPU region, so a pointer 8B below stack top is chosen (8B due to stack
+     * alignment requirements). */
+    g_context_current_states[box_id].sp = (box_mem_pos + size) - 8;
 
     /* Reset uninitialized secured box context. */
     if (slots_ctx) {
--- a/core/debug/src/debug.c
+++ b/core/debug/src/debug.c
@@ -65,7 +65,7 @@ void debug_exception_frame(uint32_t lr, uint32_t sp)
     int i;
     int mode = lr & 0x4;
 
-    char exc_sf_verbose[CONTEXT_SWITCH_EXC_SF_SIZE + 1][6] = {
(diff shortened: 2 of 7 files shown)


asensaoui:
I gave it a try, the interrupt issue is resolved, but as @Patater said it breaks IRQ nesting.

AlessandroA:
@niklas-arm WRT "Align page heap MPU region to correct alignment" could you please specify the changes needed in the EFM32GG ld and in the porting guide so that I can fix them right after the PR?

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -478,4 +478,28 @@ void vmpu_arch_init_hw(void)
...
+     *
+     * Example: 2^N = page size, 2^(N-1) = 1/8th SRAM (32kB page size in a 128kB SRAM).
+     *
+     * |           |
+     * +-----------+ <-- uvisor_config.page_end: 0x30000 = 3/8th * 128kB is aligned to 16kB.
+     * | 32kB page |
+     * +-----------+ <-- page start address: 0x30000 - 32kB = 0x10000 is not aligned to 32kB!!
+     * |           |
+     *
+     * Due to these contradicting alignment requirements, it is not possible
```
AlessandroA: Can you turn this into a FIXME or a GH issue? We can probably mark the issue as enhancement.

niklas-arm:
> specify the changes needed in the EFM32GG ld

I don't think there are any changes needed in mbed-os, the fix is required inside uVisor only.

niklas-arm:
For reference, this is [the port I am using](https://github.com/niklas-arm/mbed-os/commit/882e1db122ed3cc91689584a28984524dcb4a981#diff-d9334c46f62467d3f40f1aaba61bf3c3R185).

core/system/src/mpu/vmpu_armv7m.c
```diff
@@ -360,7 +360,7 @@ void vmpu_acl_stack(uint8_t box_id, uint32_t bss_size, uint32_t stack_size)
     /* allocate context pointer */
     g_context_current_states[box_id].bss = slots_ctx ? box_mem_pos : (uint32_t) NULL;
     /* ensure stack band on top for stack underflow dectection */
-    g_context_current_states[box_id].sp = (box_mem_pos + size - UVISOR_STACK_BAND_SIZE);
+    g_context_current_states[box_id].sp = (box_mem_pos + size) - 8;
```
niklas-arm: I don't think the `UVISOR_STACK_BAND_SIZE` [is really thought through](https://github.com/ARMmbed/uvisor/search?utf8=✓&q=UVISOR_STACK_BAND_SIZE).
It's not implementable on v7-M with out another MPU region, so what's the point of this abstraction?

core/system/src/mpu/vmpu_armv7m.c
niklas-arm: This will be difficult to solve with `1 page = 1 subregions`, due to subregion alignment issues. Larger pages sizes require `1 page = 2^N subregions`. I'll open a GH issue.

core/system/src/mpu/vmpu_armv7m.c
niklas-arm: I'm opening an issue for this, since I believe the intention for the stack band and the implementation diverged, especially now that the box stack isn't even used as the main stack anymore.

core/system/src/mpu/vmpu_armv7m.c
niklas-arm: https://github.com/ARMmbed/uvisor/issues/339

core/system/src/mpu/vmpu_armv7m.c
niklas-arm: https://github.com/ARMmbed/uvisor/issues/340

asensaoui:
Hi @niklas-arm ,

I'm getting a hardfault, after pushing a button several time( like 30 times in a row, and sometimes less, after one or two pushes ).
I think there is a problem with the alignment of the stack within the unvic_gateway_context_switch_in:

    /* Source box: Get the current stack pointer. */
    /* Note: We must use the current PSP as the SVCall can only give us the MSP
     * register, since it was triggered from privileged mode. */
    src_sp = context_validate_exc_sf(__get_PSP());

This started to occur more often when I started using RPC API.

Patater:
Hello @asensaoui,

I have a [PR open to fix a hard fault resulting from unstacking faults](https://github.com/ARMmbed/uvisor/pull/356). Maybe you'd like to test it to see if it helps address your problem.

Regarding stack alignment within `unvic_gateway_context_switch_in`, what led you to this conclusion? This may be deserving of filing another issue.

Cheers,
Patater

asensaoui:
@Patater , Thanks for your reply. Already did, still getting the hardfault. 
What led me to that conclusion is the pc and lr registers in the stack frame : 

@"***********************************************************\n"
608,804 @"                    HARD FAULT\n"
608,817 @"***********************************************************\n"
608,849 @"\n"
608,862 @"* HFSR  : 0x40000000\n"
608,875 @"\r\n"
608,888 @"\r* CFSR  : 0x00000082\n"
608,920 @"\r\n"
608,933 @"\r* DFSR  : 0x00000001\n"
608,946 @"\r\n"
608,959 @"\r* BFAR  : 0x200048F0\n"
608,991 @"\r\n"
609,023 @"\r* MMFAR : 0x200048F0\n"
609,036 @"\r\n"
609,068 @"\r* EXCEPTION STACK FRAME\n"
609,081 @"  Exception from privileged code\n"
609,094 @"    msp:     0x20001EA8\n"
609,126 @"\r    lr:      0xFFFFFFF1\n"
609,139 @"\r  Exception stack frame:\n"
609,152 @"    msp[08]: 0x100052D9 | align\n"
609,165 @"    msp[07]: 0x4100020B | xPSR\n"
609,178 @"    msp[06]: 0x100008F6 | pc\n"
609,191 @"    msp[05]: 0x1000389D | lr\n"
609,223 @"    msp[04]: 0x10002DA0 | r12\n"
609,255 @"    msp[03]: 0x200048F0 | r3\n"
609,268 @"    msp[02]: 0x00000000 | r2\n"
609,281 @"    msp[01]: 0x10002E98 | r1\n"
609,294 @"    msp[00]: 0x200048F0 | r0\n"
609,307 @"\n"

a snippet of the the Disassembly at the address in the lr register :

10003894:   mov     r3, r4
477           src_sp = context_validate_exc_sf(__get_PSP());
10003896:   mov     r0, r3
10003898:   bl      0x100008dc <context_validate_exc_sf>
1000389c:   str     r0, [r7, #32]
480           xpsr = ipsr & ~0x1FF;

Thank you.

Patater:
Thanks for the update, @asensaoui.

Could you open a new issue and ideally provide source to a minimal application that reproduces your issue? I'll debug from there.

Thanks much!

asensaoui:
@Patater ,

While trying to reproduce the issue in a simple example, I failed, uVisor was working just fine. So I assume that the problem is within my project and not uVisor. 
Sorry for disturbing.

Thanks,
asensaoui.

Patater:
Hi @asensaoui,

Thanks for making an effort to make the simple example. If in the future it does end up looking like an issue in uVisor, please file an issue.

Thanks!

PR #341, opened 2016-09-23
Wrap thread_switch in an atomic section

niklas-arm:
The system hook thread_switch gets called both through the SVC and 
though the PendSV, which can get interrupted by the SVC.
Since calls to the uVisor API are not re-entrant, we need to wrap this
function in an atomic section to prevent faults.

### DIFF
--- a/core/system/src/export_table.c
+++ b/core/system/src/export_table.c
@@ -27,6 +27,16 @@
 #include "halt.h"
 #include "vmpu.h"
 
+/* Wrap a function call into an atomic section with IRQ state restoration. */
+#define atomic_call_wrapper(fn) { \
+        uint32_t primask = __get_PRIMASK(); \
+        __disable_irq(); \
+        fn; \
+        if (!(primask & 0x01)) { \
+            __enable_irq(); \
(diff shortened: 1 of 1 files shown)


core/system/src/export_table.c
```diff
@@ -27,6 +27,16 @@
 #include "halt.h"
 #include "vmpu.h"
 
+/* Wrap a function call into an atomic section with IRQ state restoration. */
+#define atomic_call_wrapper(fn) { \
+        uint32_t primask = __get_PRIMASK(); \
+        __disable_irq(); \
```
AlessandroA: What happens when this is called from an SVC?

core/system/src/export_table.c
```diff
@@ -27,6 +27,16 @@
 #include "halt.h"
 #include "vmpu.h"
 
+/* Wrap a function call into an atomic section with IRQ state restoration. */
+#define atomic_call_wrapper(fn) { \
```
AlessandroA: Nice :)

AlessandroA:
@meriac

PR #356, opened 2016-10-11
Stack fault recovery

Patater:
Recover from stacking and unstacking faults. Reprioritize faults relative to SVC, so we can recover from more faults and also, as a bonus, get better debug messages. (Other miscellaneous fixes are also included.)

Tested on EFM32GG and K64F with a test that fails before this PR's commits are applied and passes after this PR's commits are applied.

### DIFF
--- a/core/debug/src/debug_box.c
+++ b/core/debug/src/debug_box.c
@@ -142,3 +142,31 @@ void debug_register_driver(const TUvisorDebugDriver * const driver)
     g_debug_box.box_id = g_active_box;
     g_debug_box.initialized = 1;
 }
+
+/* FIXME This is a bit platform specific. Consider moving to a platform
+ * specific location. */
+uint32_t debug_box_enter_from_priv(uint32_t lr) {
+    uint32_t shcsr;
+    uint32_t from_priv = !(lr & 0x4);
+
+    /* If we are not handling an exception caused from privileged mode, return
+     * the original lr. */
+    if (!from_priv) {
+        return lr;
+    }
+
+    shcsr = SCB->SHCSR;
+
+    /* Make sure SVC is active. */
+    assert(shcsr & SCB_SHCSR_SVCALLACT_Msk);
+
+    /* We had a fault (from SVC), so clear the SVC fault before returning. SVC
+     * and all other exceptions must be no longer active after the EXC RETURN,
+     * or else we cause usage faults when doing SVCs later (for example, to
+     * reboot via the debug_reboot SVC). */
+    SCB->SHCSR = shcsr & ~SCB_SHCSR_SVCALLACT_Msk;
+
+    /* Return to Thread mode and use the Process Stack for return. The PSP will
+     * have been changed already. */
+    return 0xFFFFFFFD;
+}
--- a/core/debug/src/debug_box.c
+++ b/core/debug/src/debug_box.c
@@ -47,14 +47,14 @@ static void debug_deprivilege_and_return(void * debug_handler, void * return_han
 {
     /* Source box: Get the current stack pointer. */
     /* Note: The source stack pointer is only used to assess the stack
-     *       alignment. */
+     *       alignment and to read the xpsr. */
     uint32_t src_sp = context_validate_exc_sf(__get_PSP());
 
     /* Destination box: The debug box. */
     uint8_t dst_id = g_debug_box.box_id;
 
     /* Copy the xPSR from the source exception stack frame. */
-    uint32_t xpsr = ((uint32_t *) src_sp)[7];
+    uint32_t xpsr = vmpu_unpriv_uint32_read((uint32_t) &((uint32_t *) src_sp)[7]);
 
     /* Destination box: Forge the destination stack frame. */
     /* Note: We manually have to set the 4 parameters on the destination stack,
--- a/core/system/src/unvic.c
+++ b/core/system/src/unvic.c
@@ -579,15 +579,14 @@ void unvic_init(void)
     /* Verify that the priority bits read at runtime are realistic. */
     assert(g_nvic_prio_bits > 0 && g_nvic_prio_bits <= 8);
 
-    /* check that minimum priority is still in the range of possible priority
-     * levels */
+    /* Check that minimum priority is still in the range of possible priority
+     * levels. */
     assert(__UVISOR_NVIC_MIN_PRIORITY < UVISOR_VIRQ_MAX_PRIORITY);
 
-    /* by setting the priority group to 0 we make sure that all priority levels
+    /* By setting the priority group to 0 we make sure that all priority levels
      * are available for pre-emption and that interrupts with the same priority
      * level occurring at the same time are served in the default way, that is,
-     * by IRQ number
-     * for example, IRQ 0 has precedence over IRQ 1 if both have the same
-     * priority level */
+     * by IRQ number. For example, IRQ 0 has precedence over IRQ 1 if both have
+     * the same priority level. */
     NVIC_SetPriorityGrouping(0);
 }
--- a/core/system/src/unvic.c
+++ b/core/system/src/unvic.c
@@ -583,6 +583,21 @@ void unvic_init(void)
      * levels. */
     assert(__UVISOR_NVIC_MIN_PRIORITY < UVISOR_VIRQ_MAX_PRIORITY);
 
+    /* Set the priority of each exception. SVC is lower priority than
+     * MemManage, BusFault, and UsageFault, so that we can recover from
+     * stacking MemManage faults more simply. */
+    static const uint32_t priority_0 = __UVISOR_NVIC_MIN_PRIORITY - 2;
+    static const uint32_t priority_1 = __UVISOR_NVIC_MIN_PRIORITY - 1;
+    assert(priority_0 < __UVISOR_NVIC_MIN_PRIORITY);
+    assert(priority_1 < __UVISOR_NVIC_MIN_PRIORITY);
+    NVIC_SetPriority(MemoryManagement_IRQn, priority_0);
+    NVIC_SetPriority(BusFault_IRQn, priority_0);
+    NVIC_SetPriority(UsageFault_IRQn, priority_0);
+    NVIC_SetPriority(SVCall_IRQn, priority_1);
+    NVIC_SetPriority(DebugMonitor_IRQn, __UVISOR_NVIC_MIN_PRIORITY);
+    NVIC_SetPriority(PendSV_IRQn, UVISOR_VIRQ_MAX_PRIORITY);
+    NVIC_SetPriority(SysTick_IRQn, UVISOR_VIRQ_MAX_PRIORITY);
+
     /* By setting the priority group to 0 we make sure that all priority levels
      * are available for pre-emption and that interrupts with the same priority
      * level occurring at the same time are served in the default way, that is,
--- a/core/vmpu/src/armv7m/vmpu_armv7m.c
+++ b/core/vmpu/src/armv7m/vmpu_armv7m.c
@@ -39,17 +39,17 @@ static const MpuRegion* vmpu_fault_find_region(uint32_t fault_addr)
 {
     const MpuRegion *region;
 
-    /* check current box if not base */
+    /* Check current box if not base. */
     if ((g_active_box) && ((region = vmpu_region_find_for_address(g_active_box, fault_addr)) != NULL)) {
         return region;
     }
 
-    /* check base-box */
+    /* Check base-box. */
     if ((region = vmpu_region_find_for_address(0, fault_addr)) != NULL) {
         return region;
     }
 
-    /* If no region was found. */
+    /* If no region was found */
     return NULL;
 }
 
@@ -57,27 +57,26 @@ uint32_t vmpu_fault_find_acl(uint32_t fault_addr, uint32_t size)
 {
     const MpuRegion *region;
 
-    /* return ACL if available */
+    /* Return ACL if available. */
     /* FIXME: Use SECURE_ACCESS for SCR! */
     if (fault_addr == (uint32_t) &SCB->SCR) {
         return UVISOR_TACL_UWRITE | UVISOR_TACL_UREAD;
     }
 
-    /* translate fault_addr into its physical address if it is in the bit-banding region */
+    /* Translate fault_addr into its physical address if it is in the bit-banding region. */
     if (fault_addr >= VMPU_PERIPH_BITBAND_START && fault_addr <= VMPU_PERIPH_BITBAND_END) {
         fault_addr = VMPU_PERIPH_BITBAND_ALIAS_TO_ADDR(fault_addr);
-    }
-    else if (fault_addr >= VMPU_SRAM_BITBAND_START && fault_addr <= VMPU_SRAM_BITBAND_END) {
+    } else if (fault_addr >= VMPU_SRAM_BITBAND_START && fault_addr <= VMPU_SRAM_BITBAND_END) {
         fault_addr = VMPU_SRAM_BITBAND_ALIAS_TO_ADDR(fault_addr);
     }
 
-    /* search base box and active box ACLs */
+    /* Search base box and active box ACLs. */
     if (!(region = vmpu_fault_find_region(fault_addr))) {
         return 0;
     }
 
-    /* ensure that data fits in selected region */
-    if((fault_addr + size) > region->end) {
+    /* Ensure that data fits in selected region. */
+    if ((fault_addr + size) > region->end) {
         return 0;
     }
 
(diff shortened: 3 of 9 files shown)


core/system/inc/page_allocator_config.h
```diff
@@ -23,18 +23,18 @@
 
 /* We can only protect a small number of pages efficiently, so there should be
  * a relatively low limit to the number of pages.
- * By default a maximum of 16 pages are allowed. This can only be overwritten
- * by the porting engineer for the current platform. */
+ * By default a certain maximum number of pages are allowed. This can only be
+ * overwritten by the porting engineer for the current platform. */
 #ifndef UVISOR_PAGE_MAX_COUNT
-#define UVISOR_PAGE_MAX_COUNT (16UL)
+#define UVISOR_PAGE_MAX_COUNT (128UL)
```
niklas-arm: I believe @meriac wanted this to stay low. This results in  (128 + 8 + 31) / 32) = 5 * `uint32_t` * #boxes (something like 5 * (5 + 1) * 4 = 30 * 4 = 120B).

core/system/inc/page_allocator.h
```diff
@@ -59,6 +59,6 @@ extern uint32_t g_page_size;
 /* Points to the beginning of the page heap. */
 extern const void * g_page_heap_start;
 /* Contains the page usage mapped by owner. */
-extern uint32_t g_page_owner_map[UVISOR_MAX_BOXES][(UVISOR_PAGE_MAX_COUNT + 31) / 32];
+extern uint32_t g_page_owner_map[UVISOR_MAX_BOXES][UVISOR_PAGE_MAP_COUNT];
```
niklas-arm: Good catch!

core/system/inc/page_allocator_config.h
meriac: @Patater It's not acceptable to increase the pagesize for everybody.
- either we make that value configurable
- or create alternative uVisor-binaries for testing
- or add the page eviction

api/inc/box_config.h
```diff
@@ -109,6 +109,9 @@ UVISOR_EXTERN void const * const main_cfg_ptr;
     \
     UVISOR_EXTERN const __attribute__((section(".keep.uvisor.cfgtbl_ptr"), aligned(4))) void * const box_name ## _cfg_ptr = &box_name ## _cfg;
 
+#define UVISOR_BOX_CFG_PTR_DECL(box_name) \
```
AlessandroA: How about we call this `UVISOR_BOX_EXTERN`? We know that internally it declares the box cfg ptr, but from the POV of the user it's just extern-declaring the box.

core/uvisor-config.h
```diff
@@ -27,7 +27,13 @@
  * link failure, so it can be used to track down a rough estimation of the
  * uVisor flash usage upper bound.
  */
-#define UVISOR_FLASH_LENGTH_MAX 0x9C00
+#if NDEBUG
+/* Release builds can be up to this big. */
```
AlessandroA: How does this work when we preprocess the linker script? The options for the preprocessing step are the following:

```make
LINKER_CONFIG:=\
    -D$(CONFIGURATION) \
    -I$(PLATFORM_DIR)/$(PLATFORM)/inc \
    -include $(CORE_DIR)/uvisor-config.h
```

But `NDEBUG` is not set there.

api/inc/box_config.h
```diff
@@ -107,7 +107,10 @@ UVISOR_EXTERN void const * const main_cfg_ptr;
         acl_list_count \
     }; \
     \
-    extern const __attribute__((section(".keep.uvisor.cfgtbl_ptr"), aligned(4))) void * const box_name ## _cfg_ptr = &box_name ## _cfg;
+    UVISOR_EXTERN const __attribute__((section(".keep.uvisor.cfgtbl_ptr"), aligned(4))) void * const box_name ## _cfg_ptr = &box_name ## _cfg;
+
+#define UVISOR_BOX_CFG_PTR_DECL(box_name) \
```
AlessandroA: How about we call this `UVISOR_BOX_EXTERN`? We know that internally it declares the box cfg ptr, but from the POV of the user it's just extern-declaring the box.

core/uvisor-config.h
Patater: Good catch. I had a look into the release `*.linker` files and they are using the large debug limit. I'll have to set NDEBUG in the linker script generation step.

api/inc/box_config.h
Patater: OK

api/inc/box_config.h
Patater: OK

core/uvisor-config.h
Patater: Fixed

mazimkhan:
retest

mazimkhan:
retest uvisor

mazimkhan:
build uvisor

Patater:
Rebased on latest master

core/debug/src/debug_box.c
```diff
@@ -142,3 +142,46 @@ void debug_register_driver(const TUvisorDebugDriver * const driver)
     g_debug_box.box_id = g_active_box;
     g_debug_box.initialized = 1;
 }
+
+/* XXX This is a bit platform specific. Consider moving to an platform specific
```
AlessandroA: If this is meant to stay, please turn into a FIXME.

core/vmpu/src/armv7m/vmpu_armv7m.c
```diff
@@ -91,7 +91,7 @@ static int vmpu_fault_recovery_mpu(uint32_t pc, uint32_t sp, uint32_t fault_addr
     uint8_t mask, index, page;
 
     /* No recovery is possible if the MPU syndrome register is not valid. */
-    if (fault_status != 0x82) {
+    if (!((fault_status == 0x82) || (fault_status & 0x18))) {
```
AlessandroA: Could you add a comment specifying what the magic numbers mean? I actually always wanted to use the CMSIS symbols for these checks, but it would make them much longer. What do you think?

core/vmpu/src/armv7m/vmpu_armv7m.c
```diff
@@ -127,29 +127,45 @@ void vmpu_sys_mux_handler(uint32_t lr, uint32_t msp)
...
-                DEBUG_FAULT(FAULT_MEMMANAGE, lr, msp);
-                HALT_ERROR(FAULT_MEMMANAGE, "Cannot recover from privileged MemManage fault");
+                return;
             }
+
+            /* If recovery was not successful, throw an error and halt. */
+            DEBUG_FAULT(FAULT_MEMMANAGE, lr, lr & 0x4 ? psp : msp);
+            VMPU_SCB_MMFSR = fault_status;
+            HALT_ERROR(PERMISSION_DENIED, "Access to restricted resource denied");
+            debug_box_enter_from_priv(ipsr, lr);
```
AlessandroA: Does this line imply that this path is only reachable by a non-recoverable from-privileged MemManage fault? If yes, please specify it as a comment in line 164.

core/debug/src/debug_box.c
Patater: Yes, will move to a FIXME. We'll refactor when we have another architecture to add support for.

core/vmpu/src/armv7m/vmpu_armv7m.c
AlessandroA: Also, I don't understand the `debug_box_enter_from_priv` very well. The debug box is a "hidden feature", meaning that it doesn't have visibility from the uVisor core. Instead, halt or print functions internally revert to the debug box if present. What is this function going to do here? I would expect that this function is called from some debug-box-specific function before returning here.

core/vmpu/src/armv7m/vmpu_armv7m.c
```diff
@@ -193,6 +206,7 @@ void vmpu_sys_mux_handler(uint32_t lr, uint32_t msp)
         case HardFault_IRQn:
             DEBUG_FAULT(FAULT_HARD, lr, lr & 0x4 ? psp : msp);
             HALT_ERROR(FAULT_HARD, "Cannot recover from a hard fault.");
+            debug_box_enter_from_priv(ipsr, lr);
```
AlessandroA: See comments above.

core/vmpu/src/armv7m/vmpu_armv7m.c
Patater: No, `debug_box_enter_from_priv` (if that's the line you mean) is reachable from a non-privileged MemManage fault as well. `debug_box_enter_from_priv` does nothing when called from privileged mode, letting the ordinary `EXC_RETURN` (instead of a forged one) enter the debug box.

core/vmpu/src/armv7m/vmpu_armv7m.c
Patater: k

core/vmpu/src/kinetis/vmpu_kinetis.c
```diff
@@ -82,51 +82,67 @@ void vmpu_sys_mux_handler(uint32_t lr, uint32_t msp)
...
-                DEBUG_FAULT(FAULT_BUS, lr, msp);
-                HALT_ERROR(FAULT_BUS, "Cannot recover from privileged bus fault");
+                return;
             }
+
+            /* If recovery was not successful, throw an error and halt. */
+            DEBUG_FAULT(FAULT_BUS, lr, lr & 0x4 ? psp : msp);
+            VMPU_SCB_BFSR = fault_status;
+            HALT_ERROR(PERMISSION_DENIED, "Access to restricted resource denied");
+            debug_box_enter_from_priv(ipsr, lr);
```
AlessandroA: See comments above.

core/system/src/unvic.c
```diff
@@ -583,6 +583,18 @@ void unvic_init(void)
      * levels. */
     assert(__UVISOR_NVIC_MIN_PRIORITY < UVISOR_VIRQ_MAX_PRIORITY);
 
+    /* Set the priority of each exception. SVC is lower priority than
+     * MemManage, BusFault, and UsageFault, so that we can recovery from
+     * stacking MemManage faults more simply. Set DebugMonitor, PendSV, SYSTICK
+     * to the same priority as SVC. */
+    NVIC_SetPriority(MemoryManagement_IRQn, 0);
```
AlessandroA: Not sure if there's an elegant way to do this, but it seems to me that these hard-coded priority values are too decoupled from the `__UVISOR_NVIC_MIN_PRIORITY ` symbol. For example, at least we should check that the highest priority number used here (`1`) is lower than the `__UVISOR_NVIC_MIN_PRIORITY` value.

core/vmpu/src/armv7m/vmpu_armv7m.c
Patater: `debug_box_enter_from_priv` never returns if from called with an `lr` that indicates an exception from privileged mode. It makes a fake EXC_RETURN, to use the `psp` that `debug_deprivilege_and_return` made, and executes the EXC_RETURN. When called with an `lr` that indicates an exception from non-privileged mode, the real EXC_RETURN is used just as before.

core/system/src/unvic.c
Patater: Done

core/vmpu/src/armv7m/vmpu_armv7m.c
Patater: Added a commit to switch magic numbers to CMSIS symbols for these fault status bits.

Patater:
Responded to Alessandro's comments

Patater:
Tests pass on EFM32

Patater:
Tests pass on K64F

core/vmpu/src/armv7m/vmpu_armv7m.c
Patater: I didn't like it either. I justified it to myself my thinking, "We are halting anyway. We don't care if we leave anything on the stack."

Your suggestion is an improvement in the sense that the stack is cleaned up. I can see the importance of leaving the uVisor stack clean when we support halting individual boxes.

Patater:
Responded to Alessandro's comments and re-ran tests. Tests are passing on EFM32 and K64F.

core/debug/src/debug_box.c
```diff
@@ -47,14 +47,14 @@ static void debug_deprivilege_and_return(void * debug_handler, void * return_han
...
-     *       alignment. */
+     *       alignment and to read the xpsr. */
     uint32_t src_sp = context_validate_exc_sf(__get_PSP());
 
     /* Destination box: The debug box. */
     uint8_t dst_id = g_debug_box.box_id;
 
     /* Copy the xPSR from the source exception stack frame. */
-    uint32_t xpsr = ((uint32_t *) src_sp)[7];
+    uint32_t xpsr = vmpu_unpriv_uint32_read((uint32_t) &((uint32_t *) src_sp)[7]);
```
AlessandroA: Isn't this an overkill? The `context_validate_exc_sf` function does exactly the same check (reads with the `t` instruction).

core/system/src/unvic.c
```diff
@@ -583,6 +583,22 @@ void unvic_init(void)
      * levels. */
     assert(__UVISOR_NVIC_MIN_PRIORITY < UVISOR_VIRQ_MAX_PRIORITY);
 
+    /* Set the priority of each exception. SVC is lower priority than
+     * MemManage, BusFault, and UsageFault, so that we can recovery from
+     * stacking MemManage faults more simply. Set DebugMonitor, PendSV, SYSTICK
+     * to the same priority as SVC. */
```
AlessandroA: The comment `Set DebugMonitor, PendSV, SYSTICK to the same priority as SVC.` is outdated now.

core/debug/src/debug_box.c
```diff
@@ -142,3 +142,37 @@ void debug_register_driver(const TUvisorDebugDriver * const driver)
...
+     * the original lr. */
+    if (!from_priv) {
+        return lr;
+    }
+
+    shcsr = SCB->SHCSR;
+    from_svc = shcsr & SCB_SHCSR_SVCALLACT_Msk;
+
+    /* Make sure SVC is active. */
+    if (!from_svc) {
```
AlessandroA: Can this happen? If not, maybe it's better to turn it into a debug-only runtime assertion.

core/debug/src/debug_box.c
```diff
@@ -142,3 +142,37 @@ void debug_register_driver(const TUvisorDebugDriver * const driver)
...
+
+    /* We had a fault (from SVC), so clear the SVC fault before returning. SVC
+     * and all other exceptions must be no longer active after the EXC RETURN,
+     * or else we cause usage faults when doing SVCs later (for example, to
+     * reboot via the debug_reboot SVC). */
+    SCB->SHCSR = shcsr & ~SCB_SHCSR_SVCALLACT_Msk;
+
+    /* Return to Thread mode and use the Process Stack for return. The PSP will
+     * have been changed already in debug_deprivilege_and_return via
+     * context_switch_in. */
```
AlessandroA: Just to make sure that this comment does not become outdated when we change the code, you could turn it into something more general, as "It's the responsibility of the caller to se the PSP to a valid value.".

core/vmpu/src/armv7m/vmpu_armv7m.c
```diff
@@ -127,29 +130,45 @@ void vmpu_sys_mux_handler(uint32_t lr, uint32_t msp)
...
     switch (ipsr) {
         case MemoryManagement_IRQn:
-            /* Currently we only support recovery from unprivileged mode. */
-            if (lr & 0x4) {
+            fault_status = VMPU_SCB_MMFSR;
+
+            /* If we are having an unstacking fault, we can't read the pc
+             * at fault. */
+            if (fault_status & (SCB_CFSR_MSTKERR_Msk | SCB_CFSR_MUNSTKERR_Msk)) {
+                /* fake pc */
```
AlessandroA: s/fake/Fake

core/debug/src/debug_box.c
Patater: Yes, the commit description spells it out as not really necessary, as a non-recoverable fault would happen in `context_validate_exc_sf`, preventing further progress. As currently written, the unprivileged access instruction is not technically necessary.

However, from the defensive coding perspective, it makes sense to use the unprivileged access instruction. Maybe the `context_validate_exc_sf` goes away on accident or has a bug introduced on accident. It's also nice to consistently have all places where uVisor accesses unprivileged memory using unprivileged access instructions.

core/debug/src/debug_box.c
Patater: OK. Debug makes sense because it'd be a stupid uVisor programmer's fault if this happened.

core/debug/src/debug_box.c
Patater: Updated

core/vmpu/src/armv7m/vmpu_armv7m.c
```diff
@@ -127,29 +130,45 @@ void vmpu_sys_mux_handler(uint32_t lr, uint32_t msp)
...
-                DEBUG_FAULT(FAULT_MEMMANAGE, lr, msp);
-                HALT_ERROR(FAULT_MEMMANAGE, "Cannot recover from privileged MemManage fault");
+                return lr;
             }
+
+            /* If recovery was not successful, throw an error and halt. */
+            DEBUG_FAULT(FAULT_MEMMANAGE, lr, lr & 0x4 ? psp : msp);
+            VMPU_SCB_MMFSR = fault_status;
+            HALT_ERROR(PERMISSION_DENIED, "Access to restricted resource denied");
+            lr = debug_box_enter_from_priv(lr);
```
AlessandroA: Is this tested in release mode as well?

core/debug/src/debug_box.c
AlessandroA: 👍

core/vmpu/src/armv7m/vmpu_armv7m.c
AlessandroA: 👍

PR #371, opened 2016-11-04
Add script to test a range of commits

Patater:
This is handy to see if all the commits you have in your branch, back to
`origin/master`, each build individually.

@AlessandroA @meriac

### DIFF
--- a/tools/test_commit_range.sh
+++ b/tools/test_commit_range.sh
@@ -1,0 +1,68 @@
+#!/bin/sh
+###########################################################################
+#
+#  Copyright 2016, ARM Limited, All Rights Reserved
+#  SPDX-License-Identifier: Apache-2.0
+#
+#  Licensed under the Apache License, Version 2.0 (the "License"); you may
+#  not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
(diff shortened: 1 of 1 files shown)


Patater:
I have one idea to improve this script actually. Please allow me to implement before merging.

Patater:
I have one idea to improve this script actually. Please allow me to implement before merging.

Patater:
OK. I made the script return to the initial branch when done. Without this change, the script would always leave you with a detached head.

tools/test_commit_range.sh
```diff
@@ -0,0 +1,40 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+###########################################################################
+
+# Check if all commits since `origin/master` build
+
+COMMITS=$(git rev-list --reverse origin/master..HEAD)
```
AlessandroA: Could we pass `remote/reference` as a cmd line arg? It should be easy to add. In my case I never have `origin` for example, but the actual remote names, `ARMmbed`, `AlessandroA`, etc. No cmd line arg could default to `origin/master`

tools/test_commit_range.sh
```diff
@@ -0,0 +1,68 @@
...
+# value of ORIGIN.
+PARAM_ORIGIN=$1
+
+if [ ! -z $PARAM_ORIGIN ]; then
+    ORIGIN=$PARAM_ORIGIN
+fi
+
+# Come up with the list of commits between the ORIGIN and HEAD, starting from
+# the oldest commit.
+COMMITS=$(git rev-list --reverse $ORIGIN..HEAD)
```
AlessandroA: As a nice side effect of this change, you can now also use a commit instead of a branch, so if you run the script once, find a bug, fix it, then you can re-start from the just-fixed commit :)

PR #372, opened 2016-11-04
Refactor the vMPU and debug files

AlessandroA:
This PR introduces minor fixes and refactors the debug and vMPU folders.

The idea behind the PR is to make the folder structure in the `core` folder consistent and to avoid some code duplication.

In addition, the Makefile has been simplified so that (i.) you don't need to manually add source files to the compilation, and (ii.) MPU-specific files are conditionally compiled through filter directories:

The new debug functions (v8-M specific) will be added on top of this PR separately.

```
core/vmpu/vmpu.c
          kinetis/vmpu_kinetis.c
          kinetis/...
          ...
```

@meriac @niklas-arm @Patater


core/system/src/main.c
```diff
@@ -16,85 +16,65 @@
...
 #include <uvisor.h>
 #include "debug.h"
+#include "page_allocator.h"
 #include "svc.h"
 #include "unvic.h"
 #include "vmpu.h"
-#include "page_allocator.h"
 
-UVISOR_NOINLINE void uvisor_init_pre(void)
+static UVISOR_NOINLINE void uvisor_init_pre(void)
```
Patater: Why add static, especially if this is supposed to be not inlined?

core/system/src/main.c
```diff
@@ -119,30 +99,26 @@ static void load_priv_sys_hooks(void)
...
-                        __uvisor_config.page_size);
+    /* Initialize the page allocator. */
+    page_allocator_init(__uvisor_config.page_start, __uvisor_config.page_end, __uvisor_config.page_size);
 
-    /* Init SVC call interface. */
+    /* Initialize the SVCall interface. */
     svc_init();
 
-    /* Load privileged system hooks. */
+    /* Load the privileged system hooks. */
```
Patater: With or without 'the' is fine in this case.

AlessandroA:
> The commit "vMPU: Move files to dedicated folder" should not rename "core/debug/src/debug_kinetis.c" → "core/debug/src/debug_freescale_k64.c". Doing so undoes what a previous commit ("Debug: Move debug source code all in same folder") wanted to do.

Fixed.

> Now that all the vmpu_* files are in a vmpu directory, does it still make sense to have vmpu_ in the file names?

I would like to keep `vmpu_` as otherwise the purpose of the files is not very clear.

> The commit "Makefile: Automate source file listings" also renames a bunch of files. Renaming should not be done in this commit which is meant to auto discover source file listings, but in a different commit.

I split the commit into 2 so the changes to the Makefile related to automating the discovery process are cleaner. The new commit still need to rename+change the Makefile, as otherwise the commit itself would not compile.

AlessandroA:
* Commit msg fixed.
* Rebased to latest ARMmbed/uvisor.

PR #373, opened 2016-11-08
Proper syndrome registers debugging

AlessandroA:
This PR makes debugging more user-friendly by explaining in a verbose way the reason for every fault.

@meriac @Patater @niklas-arm

### DIFF
--- a/core/debug/src/debug.c
+++ b/core/debug/src/debug.c
@@ -93,17 +93,56 @@ UVISOR_WEAK void debug_fault_bus_hw(void) {}
 
 static void debug_fault_hard(void)
 {
-    dprintf("* HFSR  : 0x%08X\r\n\r\n", SCB->HFSR);
-    dprintf("* CFSR  : 0x%08X\r\n\r\n", SCB->CFSR);
-    dprintf("* DFSR  : 0x%08X\r\n\r\n", SCB->DFSR);
-    dprintf("* BFAR  : 0x%08X\r\n\r\n", SCB->BFAR);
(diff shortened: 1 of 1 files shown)


core/debug/src/debug.c
```diff
@@ -93,34 +93,126 @@ UVISOR_WEAK void debug_fault_bus_hw(void) {}
...
 {
-    dprintf("* HFSR  : 0x%08X\r\n\r\n", SCB->HFSR);
-    dprintf("* CFSR  : 0x%08X\r\n\r\n", SCB->CFSR);
-    dprintf("* DFSR  : 0x%08X\r\n\r\n", SCB->DFSR);
-    dprintf("* BFAR  : 0x%08X\r\n\r\n", SCB->BFAR);
-    dprintf("* MMFAR : 0x%08X\r\n\r\n", SCB->MMFAR);
+    dprintf("* FAULT SYNDROME REGISTERS\r\n");
+    dprintf("\r\n");
+
+    uint32_t hfsr = SCB->CFSR;
```
Patater: Should this be HFSR and not CFSR?

AlessandroA: Good catch! Fixed.

AlessandroA:
@Patater Added whole register as well. Now you always get the full register view (HFSR, CFSR or DFSR) and then also the detailed bit-by-bit explanation. On a given fault, only bits specific to that fault are spelled out (UFSR, BFSR, MMFSR views in the CFSR).

Patater:
Excellent. Just what the doctor ordered. Thanks!

PR #375, opened 2016-11-08
svc: Make svc_exports.h great again

Patater:
Removing svc_exports.h broke the register-level gateway. Re-create
svc_exports.h with just the minimum needed to make the register-level
gateway work again.

Fixes: f51198ca378a ("Expose all uVisor API via a function table")


api/inc/register_gateway.h
```diff
@@ -19,13 +19,17 @@
...
+#include "api/inc/svc_exports.h"
 #include <stdint.h>
 
 /** Get the offset of a struct member.
  * @internal
  */
 #define __UVISOR_OFFSETOF(type, member) ((uint32_t) (&(((type *)(0))->member)))
 
+/** Generate the SVCall opcode from the SVC ID. */
+#define UVISOR_SVC_OPCODE(id) ((uint16_t) 0xDF00 | (uint8_t) ((id) & 0xFF))
```
meriac: @Patater the (uint8_t) caste in combination and the 0xFF seems to be too much - either one or the other?

Patater: This matches what we had before. I'm just moving it between files now.

Patater:
Tests still pass on EFM32.

PR #376, opened 2016-11-14
Add MPU invalidation API and semaphore timeouts

Patater:
This is useful for testing as we can use it to force memory faults to
occur at controlled locations.

Also add semaphore timeout reporting, so that `rpc_fncall_waitfor` can report timeouts properly.

### DIFF
--- a/api/inc/api.h
+++ b/api/inc/api.h
@@ -60,6 +60,7 @@ typedef struct {
 
     void (*debug_init)(const TUvisorDebugDriver * const driver);
     void (*error)(THaltUserError reason);
+    void (*vmpu_mem_invalidate)(void);
 
     int                (*pool_init)(uvisor_pool_t *, void *, size_t, size_t, int);
     int                (*pool_queue_init)(uvisor_pool_queue_t *, uvisor_pool_t *, void *, size_t, size_t, int);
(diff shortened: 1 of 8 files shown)


api/inc/api.h
```diff
@@ -70,6 +70,8 @@ typedef struct {
     uvisor_pool_slot_t (*pool_queue_dequeue_first)(uvisor_pool_queue_t *);
     uvisor_pool_slot_t (*pool_queue_find_first)(uvisor_pool_queue_t *, TQueryFN_Ptr, void *);
 
+    void (*vmpu_mem_invalidate)(void);
```
AlessandroA: Isn't the offset of the function here different from the one in https://github.com/ARMmbed/uvisor/pull/376/commits/ecb88f98dc439ab10d070a2203b5feae189ec263#diff-a763365abbc3765a4810f65eec790b53R48 ?

AlessandroA:
@Patater IIUC this PR is exposing an API that uVisor already had internally (to invalidate the MPU regions) as a way for a user to invalidate all of its current regions. If that's correct, could you please update the commit message?

Patater:
Updated commit message

meriac:
LGTM

PR #379, opened 2016-11-30
Docker: Update GCC to 5.4.1 2016q3

AlessandroA:
@meriac @Patater @niklas-arm


AlessandroA:
Superseded by https://github.com/ARMmbed/uvisor/pull/382, which is now updated to GCC 6.1. We should keep an eye on the updated version of the Fedora packages though, to get the latest version of arm-none-eabi-gcc which supports the `-march=v8-m.main` and `-mcmse` options.

Issue #369, opened 2016-11-01
Out-by-one error in vmpu_mpu_push()


nchong:
This function pushes a slot/region into the v7M MPU. There is an out-by-one error in the check to see if the slot is in range or not. Consider the case when:

```
ARMv7M_MPU_REGIONS_MAX = 8
g_mpu_slot = 7
```

Then the check for overflow `(++g_mpu_slot > ARMv7M_MPU_REGIONS_MAX)` will fail (since `8 > 8` is false).

I think the fix is to change the strict inequality to `(++g_mpu_slot >= ARMv7M_MPU_REGIONS_MAX)`.

Found using ArchEx.

ciarmcom:
ARM Internal Ref: IOTSEC-170

AlessandroA:
Thanks for reporting this @nchong! Yes, it definitely looks like at bug, but I would ask @niklas-arm for confirmation.

AlessandroA:
Fixed by #388.

PR #388, opened 2016-12-20
Fix out-by-one error in vmpu_mpu_push

nchong:
See issue #369.

This function pushes a slot/region into the v7M MPU. If `g_mpu_slot` is equal to `ARMv7M_MPU_REGIONS_MAX` (say 8) then the overflow check

```
  (++g_mpu_slot > ARMv7M_MPU_REGIONS_MAX)
    ==
  (8 > 8)
    ==
  false
```

does not fire (when it should) and the function will access the array `g_mpu_priority[8]` out-of-bounds and write to an unsupported region number.


Patater:
LGTM

Issue #380, opened 2016-12-05
v7M MPU is enabled too early


nchong:
During the initialisation of the vmpu in the procedure `vmpu_init_post`:

1. The MPU is enabled by `vmpu_arch_init()` which calls `vmpu_mpu_lock()`
2. Box configurations are loaded and regions are programmed for box 0 by `vmpu_load_box()`

This means that the MPU is enabled at (1) and *subsequently* has regions programmed at (2). At (1), only static regions will have been programmed. In particular, higher-numbered regions will still have their reset value (UNKNOWN); hence, this is architecturally UNPREDICTABLE.

A fix is to move the call to `vmpu_mpu_lock()` after `vmpu_load_box()`. This guarantees that all RBAR/RASR registers have been written-to (by `vmpu_mpu_invalidate()`) before the MPU is enabled.

Found using ArchEx.

ciarmcom:
ARM Internal Ref: IOTSEC-213

PR #389, opened 2016-12-20
Fix enabling v7M MPU too early

nchong:
See issue #380.

Move call to` vmpu_mpu_lock()` after `vmpu_load_box()`. This guarantees that all RBAR/RASR registers have been written-to (by `vmpu_mpu_invalidate()`) before the MPU is enabled.

My only concern is whether this change is compatible with the Kinetis MPU.


Patater:
Please rebase on latest uVisor.

PR #390, opened 2017-01-05
Enable K64F DMA

Patater:

### DIFF
--- a/core/vmpu/src/kinetis/vmpu_kinetis_mem.c
+++ b/core/vmpu/src/kinetis/vmpu_kinetis_mem.c
@@ -76,7 +76,7 @@ void vmpu_mem_init(void)
 
     /* rest of SRAM, accessible to mbed
      * - non-executable for uvisor
-     * - five read/write access to ENET DMA bus master */
+     * - give read/write access to ENET DMA bus master */
     vmpu_mpu_set_static_acl(2, (uint32_t) __uvisor_config.page_end,
         (uint32_t) __uvisor_config.sram_end - (uint32_t) __uvisor_config.page_end,
         UVISOR_TACL_SREAD |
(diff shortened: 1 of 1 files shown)


core/vmpu/src/kinetis/vmpu_kinetis_mem.c
```diff
@@ -63,20 +63,30 @@ void vmpu_mem_switch(uint8_t src_box, uint8_t dst_box)
 
 void vmpu_mem_init(void)
 {
+    /* UART DMA is bus master 2 (configured through M2UM). ENET DMA is bus
+     * master 3 (configured through M3UM). */
+    static const uint32_t bmum_read_bit = (1 << 2);
+    static const uint32_t bmum_write_bit = (1 << 1);
+#define M3UM 18
```
meriac: @Patater The mixture of #define and static const is not acceptable. Either both need to be #defines (namespaced and at the top of the file) or both static consts. ALso the abbreviated name M3UM is not useful.

Alternatively please use a single RW define/const (feel free to add a RO and WO define, too) for readability.

My preference though is to keep the hardcoded hex number, in combination with the comment above - unless you are ready to add proper macros and all defines for this register. It's more readable when verifying that against the K64F spec.

Patater: I'd rather not use hardcoded hex numbers, but names from the datasheet (easy to search, easy to know what one is looking at).

M3UM is useful in that it matches the register name in the data sheet. Also, there isn't a need to add all defines for all these registers, just the ones we need.

I'm not sure what style we like better regarding defines or static const. Where possible, I prefer static const.

I'll revise this a bit and you can see if you like it better.

Patater: Revised to use only static consts named with the register name, concatenated with the bitfield name, concatenated with the bit name from the datasheet.

PR #395, opened 2017-01-12
Update QUICKSTART.md

AnotherButler:
-Remove idioms
-Change passive to active voice
-Make other minor grammar changes

### DIFF
--- a/docs/api/QUICKSTART.md
+++ b/docs/api/QUICKSTART.md
@@ -1,28 +1,28 @@
 # Quick-Start Guide for uVisor on mbed OS
 
-This guide will help you get started with uVisor on mbed OS by walking you through creating a sample application for the NXP FRDM-K64F board.
+This guide will help you get started with uVisor on mbed OS by showing you how to create a sample application for the NXP FRDM-K64F board.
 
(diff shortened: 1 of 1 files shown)


docs/api/QUICKSTART.md
```diff
@@ -49,12 +49,12 @@ int main(void)
 {
     while (true) {
         led = !led;
-        Thread::wait(500);
```
meriac: @AlessandroA @AnotherButler before accepting this pull request we must verify whether the proposed change to floating-point is resolved in the pre-compiler and not during runtime. uVisor currently doesn't support floating point numbers during runtime.

AlessandroA:
I'm going to test the updated code and report back here when finished. Meanwhile, thanks a lot for the patch @AnotherButler.

Once this PR is merged, I will propose another couple commits to fix some outdated semantics and will then export the doc to mbed OS via a separate PR.

AlessandroA:
Update: The example works but because of some previous changes in the `wait()` API implementation we need to add a new ACL to the list provided in the doc for the guide to work. I will merge this PR and then create a new one containing the updated example.

Issue #392, opened 2017-01-06
A quick change in debugging documentation


asensaoui:
Box config called twice and we don't set up the heap_size : 
[debugging.md](https://github.com/ARMmbed/uvisor/blob/07dde17d610cca51f24135de9c65b15938f89f97/docs/api/DEBUGGING.md#the-debug-box)
```
/* Configure the debug box. */
UVISOR_BOX_NAMESPACE(NULL);
UVISOR_BOX_CONFIG(box_debug, UVISOR_BOX_STACK_SIZE); // box config called twice, and the heap size is not set up
UVISOR_BOX_MAIN(box_debug_main, osPriorityNormal, UVISOR_BOX_STACK_SIZE);
UVISOR_BOX_CONFIG(box_debug, acl, UVISOR_BOX_STACK_SIZE, box_context);
```

I have also some questions about the debug_box: 
Does the system reboot only in release mode and not in the debug one? If so, can I reboot it after halting the system in debug mode? 

Could we change the behavior of the system after catching an exception, like for example restarting just the box leading to the conflict?

Best regards,

adbridge:
ARM Internal Ref: IOTSEC-247

AlessandroA:
Hi @asensaoui thanks for reporting. #400 should fix the bug in the docs. As for your questions:

> Does the system reboot only in release mode and not in the debug one? If so, can I reboot it after halting the system in debug mode?

Since uVisor 0.26.0 the debug box does not reboot automatically anymore. Upon return from the `halt_error` handler, the system halts.

You can still reboot the system if you want, by calling the `NVIC_SystemReset()` API. This is only possible from the debug box. It works both in debug and release mode:

```C
static void halt_error(int reason) {
    printf("We halted with reason %i\r\n", reason);

    /* If we don't do anything, the system will halt upon return. */
    /* A debug box handler like this one can also decide to reboot the whole
     * system. This is only allowed from the debug box. */
    NVIC_SystemReset();
}
```

> Could we change the behavior of the system after catching an exception, like for example restarting just the box leading to the conflict?

We do not support the reboot of a single box yet, but it's definitely something we would like to look into in the future.

AlessandroA:
Fixed in #400.

PR #400, opened 2017-01-19
Docs: Fix sample code for debug box

AlessandroA:
Fixes:
* #392: "A quick change in debugging documentation"

@niklas-arm @Patater 

Thanks @asensaoui @janjongboom for also reporting this.

### DIFF
--- a/core/debug/src/debug_box.c
+++ b/core/debug/src/debug_box.c
@@ -92,9 +92,7 @@ void debug_halt_error(THaltError reason)
     if (!g_debug_box.initialized || debugged_once_before) {
         while (1);
     } else {
-        /* Remember that debug_deprivilege_and_return() has been called once.
-         * We'll reboot after the debug handler is run, so this will go back to
-         * zero after the reboot. */
+        /* Remember that debug_deprivilege_and_return() has been called once. */
         debugged_once_before = 1;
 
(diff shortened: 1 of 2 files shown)


AlessandroA:
@Patater added clarification regarding reboot from a debug box handler.

core/debug/src/debug_box.c
```diff
@@ -92,9 +92,7 @@ void debug_halt_error(THaltError reason)
     if (!g_debug_box.initialized || debugged_once_before) {
         while (1);
     } else {
-        /* Remember that debug_deprivilege_and_return() has been called once.
-         * We'll reboot after the debug handler is run, so this will go back to
-         * zero after the reboot. */
+        /* Remember that debug_deprivilege_and_return() has been called once. */
```
Patater: Perhaps note that the debug box can only be entered once, to avoid the debug box calling itself recursively. This is probably important enough to go into the documentation (not just comments) as well, as this limits what you can do after the debug box runs before you have to reboot. For instance, you can't catch a fault and then restart that box and catch a subsequent fault; you get one debug box entry per boot. This means you pretty much have to reboot.

AlessandroA: Updated with improved docs before the code snippet (the comment has not been changed).

PR #403, opened 2017-01-26
First box to call NVIC owns the interrupt

niklas-arm:
This also copies the initial VTOR table from flash to ram and sets correct priority. That way, statically allocated interrupts are preserved and need only be enabled.

cc @AlessandroA @Patater

### DIFF
--- a/core/system/src/unvic.c
+++ b/core/system/src/unvic.c
@@ -32,7 +32,7 @@ uint8_t g_nvic_prio_bits;
  * @internal
  *
  * If multiple disable-/re-enable-all functions are called in the same box, a
- * counter is respectively incremented/drecremented so that it never happens
+ * counter is respectively incremented/decremented so that it never happens
  * that a nested function re-enables IRQs for the caller. */
 uint32_t g_irq_disable_all_counter[UVISOR_MAX_BOXES];
 
@@ -68,7 +68,7 @@ void unvic_acl_add(uint8_t box_id, void *function, uint32_t irqn)
     uv = &g_unvic_vector[irqn];
 
     /* check if IRQ entry is populated */
-    if(uv->id || uv->hdlr)
+    if(uv->id != UVISOR_BOX_ID_INVALID)
     {
         HALT_ERROR(PERMISSION_DENIED,
                    "Permission denied: IRQ %d is owned by box %d\n\r", irqn,
@@ -80,9 +80,12 @@ void unvic_acl_add(uint8_t box_id, void *function, uint32_t irqn)
     uv->hdlr = function;
 }
 
+#define UNVIC_ISR_OWNER_OTHER 0
+#define UNVIC_ISR_OWNER_NONE  1
+#define UNVIC_ISR_OWNER_SELF  2
+
 static int unvic_acl_check(int irqn)
 {
-    int is_irqn_registered;
     TIsrUVector *uv;
 
     /* don't allow to modify uVisor-owned IRQs */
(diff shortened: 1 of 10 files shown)


core/system/src/unvic.c
```diff
@@ -91,16 +93,11 @@ static int unvic_acl_check(int irqn)
     /* get vector entry */
     uv = &g_unvic_vector[irqn];
 
-    /* an IRQn slot is considered as registered if the ISR entry in the unprivileged
-     * table is not 0 */
-    is_irqn_registered = uv->hdlr ? 1 : 0;
+    /* an IRQn slot is considered as registered if the vector id is not unused. */
+    is_irqn_registered = (uv->id != UNVIC_VECTOR_UNUSED);
```
Patater: We wouldn't need this check if, instead of `UNVIC_VECTOR_UNUSED`, we use an invalid box ID define (which we may need to create). Simply checking `uv->id != g_active_box` would be good enough.

core/system/src/unvic.c
```diff
@@ -604,4 +561,19 @@ void unvic_init(void)
      * by IRQ number. For example, IRQ 0 has precedence over IRQ 1 if both have
      * the same priority level. */
     NVIC_SetPriorityGrouping(0);
+
+    /* Make sure the UNVIC_VECTOR_UNUSED is definitely NOT a valid box id. */
+    if (vmpu_is_box_id_valid(UNVIC_VECTOR_UNUSED)) {
```
Patater: We wouldn't have to do this check if `UNVIC_VECTOR_UNUSED` were replaced with an invalid box ID define.

core/system/src/unvic.c
```diff
@@ -604,4 +561,19 @@ void unvic_init(void)
...
+            "UNVIC_VECTOR_UNUSED (%u) must not be a valid box id!\n",
+            UNVIC_VECTOR_UNUSED);
+    }
+
+    /* Copy the user interrupt table into the handlers. */
+    for (uint32_t ii=0; ii < NVIC_VECTORS; ii++) {
+        g_unvic_vector[ii].id = UNVIC_VECTOR_UNUSED;
+        g_unvic_vector[ii].hdlr = (TIsrVector) user_vtor[ii + NVIC_OFFSET];
+        /* set default priority (SVC must always be higher) */
+        NVIC_SetPriority(ii, __UVISOR_NVIC_MIN_PRIORITY);
```
Patater: Is registering the user interrupt table interrupts with the public box really necessary? It sounds like a nice thing to do at first glance, but given that interrupts must be enabled before use, isn't just having enable (and the other `vIRQ` functions) register for the interrupt good enough? Is there some case I'm missing where an interrupt in the user interrupt table can be used without calling one of the `NVIC` or `vIRQ` functions?

If it is possible for a user to use an interrupt without calling any NVIC (or vIRQ) functions, how easy is it for a user to change the user interrupt table? Does a user need to change mbed OS code?

On the surface, I agree with this change. Interrupts in the user interrupt table are "legacy" interrupts and should be removed in favor of using the "non-legacy" vIRQ API. This fits with the uVisor story well of: the public box is for backwards compatibility; to do things with secure boxes you need to use uVisor APIs. However, if it is too hard to remove some interrupt from the user interrupt table (and I'd argue that having to change a local copy of mbed OS is too hard), then I think usability goes out the window if there is no other way to unregister the interrupt from the public box. We may need to rethink including the ability for a box to unregister interrupts.

core/system/src/unvic.c
```diff
@@ -68,7 +70,7 @@ void unvic_acl_add(uint8_t box_id, void *function, uint32_t irqn)
     uv = &g_unvic_vector[irqn];
 
     /* check if IRQ entry is populated */
-    if(uv->id || uv->hdlr)
+    if(uv->id != UNVIC_VECTOR_UNUSED)
     {
         HALT_ERROR(PERMISSION_DENIED,
```
Patater: Instead of halting, which shuts down the entire system even on release builds, consider debug asserting so we can catch programmer mistakes, but silently not fulfill the request in release builds. This will allow us to write tests that make sure we handle interrupt ownership contention. Note that this will require changes to all functions that call `unvic_acl_check`, as the function can now return in the case where the IRQ is registered, but not to the caller. All functions using `unvic_acl_check` will have to deal with: IRQ unclaimed, IRQ claimed by current box, IRQ claimed by other box.

core/system/src/unvic.c
Patater: Nevermind. You are only setting the priority, not claiming the IRQ. Good job.

core/system/src/unvic.c
niklas-arm: Yup, this same mechanism is used in page allocator. I'll make it generic then.

niklas-arm:
Updated for review.

core/vmpu/inc/vmpu.h
```diff
@@ -205,6 +205,9 @@ extern int vmpu_is_region_size_valid(uint32_t size);
  */
 extern uint32_t vmpu_round_up_region(uint32_t addr, uint32_t size);
 
+/** Invalid box id for use in marking objects with invalid ownership. */
+#define UVISOR_BOX_ID_INVALID ((uint8_t) -1)
```
Patater: If this were in `vmpu_exports.h`, we wouldn't have to redefine it as -1 in the unsupported files.

core/system/src/unvic.c
```diff
@@ -296,77 +259,73 @@ void unvic_irq_enable_all(void)
...
-        DPRINTF("IRQ %d pending status cleared\n\r", irqn);
-        NVIC_ClearPendingIRQ(irqn);
+    if (!unvic_acl_check(irqn)) {
         return;
     }
-    else {
-        HALT_ERROR(NOT_ALLOWED, "IRQ %d is unregistered; state cannot be changed", irqn);
-    }
+    g_unvic_vector[irqn].id = g_active_box;
+    DPRINTF("IRQ %d registered to box %d\n\r", irqn, g_active_box);
```
Patater: The registration of an irqn to a box is something we do often in these functions. Consider placing this into an itty bitty static function, so we don't have to repeat the code and DPRINTF string so much.

Heck, even the call and check of `unvic_acl_check` is done very often. Maybe a new function like `unvic_irq_register` would be good. You'd use it like so.

```C
/* Attempt to register for the IRQ. */
if (!unvic_irq_register(irqn)) {
    /* We couldn't claim the interrupt. */
    return 0;
}
```

This is just a suggestion. Do what you think makes sense.

core/system/src/unvic.c
```diff
@@ -604,4 +563,12 @@ void unvic_init(void)
      * by IRQ number. For example, IRQ 0 has precedence over IRQ 1 if both have
      * the same priority level. */
     NVIC_SetPriorityGrouping(0);
+
+    /* Copy the user interrupt table into the handlers. */
+    for (uint32_t ii=0; ii < NVIC_VECTORS; ii++) {
+        g_unvic_vector[ii].id = UVISOR_BOX_ID_INVALID;
+        g_unvic_vector[ii].hdlr = (TIsrVector) user_vtor[ii + NVIC_OFFSET];
+        /* set default priority (SVC must always be higher) */
```
Patater: Use complete sentences (with capital letter and correct punctuation) for new comments.

core/vmpu/inc/vmpu.h
niklas-arm: Good point.

core/system/src/unvic.c
```diff
@@ -91,113 +97,86 @@ static int unvic_acl_check(int irqn)
...
-     *       if it is currently un-registered (that is, if the registered handler
-     *       is NULL) */
-    if(uv->id != g_active_box && (uv->id || is_irqn_registered))
+/* @retval 0 you do not own this interrupt
+ * @retval 1 you own this interrupt
+ */
+static int unvic_isr_register(uint32_t irqn)
+{
+    int isr_status = unvic_acl_check(irqn);
+    switch(isr_status)
```
Patater: Add space before `(`

core/system/src/unvic.c
Patater: OK

core/system/src/unvic.c
```diff
@@ -68,7 +68,7 @@ void unvic_acl_add(uint8_t box_id, void *function, uint32_t irqn)
     uv = &g_unvic_vector[irqn];
 
     /* check if IRQ entry is populated */
-    if(uv->id || uv->hdlr)
+    if(uv->id != UVISOR_BOX_ID_INVALID)
```
Patater: Space before `(`

core/system/src/unvic.c
```diff
@@ -91,113 +97,86 @@ static int unvic_acl_check(int irqn)
...
-     * note: when an IRQ is released (de-registered) the corresponding IRQn is
-     *       disabled, to ensure that no spurious interrupts are served */
-    if (vector) {
-        g_unvic_vector[irqn].id = g_active_box;
-    }
-    else {
-        NVIC_DisableIRQ(irqn);
-        g_unvic_vector[irqn].id = 0;
+    if (!unvic_isr_register(irqn)) {
+        return;
```
AlessandroA: Here and in other similar instances `return` should be replaced with an `HALT_ERROR`, so the user gets a notification, the NVIC API does not silently fail, and the system halts exactly at that point.

The backward compatibility introduced by this PR still allows legacy code to work as expected.

As @Patater points out this is a source of DoS, and it will be solved when we will support individually-killable boxes.

core/system/src/unvic.c
```diff
@@ -604,4 +569,12 @@ void unvic_init(void)
      * by IRQ number. For example, IRQ 0 has precedence over IRQ 1 if both have
      * the same priority level. */
     NVIC_SetPriorityGrouping(0);
+
+    /* Copy the user interrupt table into the handlers. */
+    for (uint32_t ii=0; ii < NVIC_VECTORS; ii++) {
```
AlessandroA:
```diff
-uint32_t ii=0;
+uint32_t ii = 0;
```

The `ii` is very MATLAB-y 😄

AlessandroA:
@niklas-arm As part of this PR you should also quickly update the docs. Please scan the docs folder for occurrences of the NVIC APIs. While the APIs haven't changed, the underlying behavior has.

This includes at least:

* https://github.com/ARMmbed/uvisor/blob/master/docs/api/API.md
* https://github.com/ARMmbed/uvisor/blob/master/docs/api/QUICKSTART.md

Please specify the new behavior (no need to set an ISR first; first box touching an IRQ gets the IRQ; you cannot release ownership of an IRQ; if you don't set an IRQ you get the box 0 default one).

core/system/src/unvic.c
```diff
@@ -604,4 +569,12 @@ void unvic_init(void)
      * by IRQ number. For example, IRQ 0 has precedence over IRQ 1 if both have
      * the same priority level. */
     NVIC_SetPriorityGrouping(0);
+
+    /* Copy the user interrupt table into the handlers. */
```
AlessandroA: Could you add a comment here specifying the reason why this is needed and the consequences? In particular, it's important to specify that if a secure box code enables an IRQ before setting its ISR, the IRQ might be immediately served by the static ISR, which is out of the secure box control (because it's usually set by box 0/the host OS).

Related to this: https://github.com/ARMmbed/uvisor/pull/403#issuecomment-276931004.

core/system/src/unvic.c
niklas-arm: I "fixed" this by halting in [the `unvic_acl_check` function](https://github.com/ARMmbed/uvisor/pull/403/commits/0cea83820e3009fa8a13898b912852b5ceb264e7#diff-8079bf2b7ed17a83b95fadc1b10f7625R106). This prevents duplicate code and restores the original behavior.

AlessandroA: Yup! The `return` gives the wrong expectation, though. Even if we use hardcoded IRQ ACLs, the API will never be meant to return if the IRQ is not registered to the calling box.

Consider replacing occurrences of:

```C
if (!unvic_isr_register(irqn)) {
        return;
}
```

with:

```C
/* This function halts if the IRQ is owned by another box or by uVisor. */
unvic_isr_register(irqn)
```

AlessandroA: (If instead you want to keep the semantics by which `unvic_isr_register(irqn)` returns the ownership status of `irqn`, you will have to halt from every function that calls that function. I like this semantics so I'd rather duplicate the `HALT_ERROR`. @Patater ?)

niklas-arm: -1 for duplication, I'll call the function without if.

core/system/src/unvic.c
```diff
@@ -91,113 +97,78 @@ static int unvic_acl_check(int irqn)
...
-    /* an IRQn slot is considered as registered if the ISR entry in the unprivileged
-     * table is not 0 */
-    is_irqn_registered = uv->hdlr ? 1 : 0;
+    if (uv->id == g_active_box) {
+        return UNVIC_ISR_OWNER_SELF;
+    }
+    if (uv->id == UVISOR_BOX_ID_INVALID) {
+        return UNVIC_ISR_OWNER_NONE;
+    }
+    HALT_ERROR(PERMISSION_DENIED, "Permission denied: IRQ %d is owned by box %d\n\r", irqn, uv->id)
```
AlessandroA: Now the return values from both `unvic_acl_check` and `unvic_isr_register` do not make sense anymore. I'd rather have `unvic_acl_check` return the ownership status and `unvic_isr_register` be a `void` function that either registers the IRQ or halts.

niklas-arm:
Updated.

core/system/src/unvic.c
```diff
@@ -91,113 +94,73 @@ static int unvic_acl_check(int irqn)
...
+            g_unvic_vector[irqn].id = g_active_box;
+            DPRINTF("IRQ %d registered to box %d\n\r", irqn, g_active_box);
+        case UNVIC_ISR_OWNER_SELF:
+            return;
+        default:
+            break;
     }
-
-    return is_irqn_registered;
+    HALT_ERROR(PERMISSION_DENIED, "Permission denied: IRQ %d is owned by another box!\n\r", irqn);
```
AlessandroA: `\n\r` ---> `\r\n`

core/system/src/unvic.c
```diff
@@ -296,77 +259,53 @@ void unvic_irq_enable_all(void)
...
-    /* enable IRQ */
-    if (is_irqn_registered) {
-        DPRINTF("IRQ %d pending status cleared\n\r", irqn);
-        NVIC_ClearPendingIRQ(irqn);
-        return;
-    }
-    else {
-        HALT_ERROR(NOT_ALLOWED, "IRQ %d is unregistered; state cannot be changed", irqn);
-    }
+    /* clear pending IRQ */
```
AlessandroA: `/* Clear pending IRQ. */`

core/system/src/unvic.c
```diff
@@ -296,77 +259,53 @@ void unvic_irq_enable_all(void)
...
-    /* enable IRQ */
-    if (is_irqn_registered) {
-        DPRINTF("IRQ %d pending status set (will be served as soon as possible)\n\r", irqn);
-        NVIC_SetPendingIRQ(irqn);
-        return;
-    }
-    else {
-        HALT_ERROR(NOT_ALLOWED, "IRQ %d is unregistered; state cannot be changed", irqn);
-    }
+    /* set pending IRQ */
```
AlessandroA: `/* Set pending IRQ. */`

core/system/src/unvic.c
```diff
@@ -91,113 +94,73 @@ static int unvic_acl_check(int irqn)
...
+    unvic_isr_register(irqn);
 
-    /* set default priority (SVC must always be higher) */
-    NVIC_SetPriority(irqn, __UVISOR_NVIC_MIN_PRIORITY);
-
-    DPRINTF("IRQ %d %s box %d\n\r",
-            irqn,
-            vector ? "registered to" : "released by",
-            g_active_box);
+    /* save unprivileged handler */
```
AlessandroA: `/* Save the unprivileged handler. */`

AlessandroA:
retest uvisor

AlessandroA:
retest uvisor

AlessandroA:
retest uvisor

AlessandroA:
retest uvisor

PR #402, opened 2017-01-26
PIT peripheral is required on K64F

janjongboom:
@AlessandroA 

FYI, I'm adding RPC guide and how to use flash storage to the quickstart guide. Will come tomorrow or so.

PR #404, opened 2017-01-27
RPC and non-volatile storage from a secure box

janjongboom:
Depends on #402.

Add documentation on:

* How to set up RPC to call functions in a secure box from the main box.
* How to use non-volatile storage from a secure box.

Plus a nit on how to netcat to devices controlled by pyOCD or OpenOCD (as that took me a while to figure out).

@AlessandroA @iriark01


docs/api/QUICKSTART.md
```diff
@@ -193,14 +192,14 @@ Create a new source file, `~/code/uvisor-example/source/secure_box.cpp`. We will
...
 #include "mbed.h"
-#include "rtos.h"
 #include "uvisor-lib/uvisor-lib.h"
 
 /* Private static memory for the secure box */
 typedef struct {
     uint32_t * buffer;          /* Static private memory, pointing to dynamically allocated private memory */
     uint32_t counter;           /* Static private memory */
     int index;                  /* Static private memory */
+    RawSerial * pc;             /* Static private memory, pointing to dynamically allocated private memory */
```
janjongboom: I added this because calling `printf` from the RPC thread crashed the board with an RTX error. Using uvisor_ctx->pc->printf is fine.

docs/api/QUICKSTART.md
```diff
@@ -249,21 +248,27 @@ static void private_button_on_press(void)
         uvisor_ctx->index = 0;
 
         /* For debug purposes: Print the content of the buffer. */
-        printf("Thread count between button presses: ");
+        uvisor_ctx->pc->printf("Thread count between button presses: ");
```
janjongboom: This runs in an ISR, so we shouldn't do this.

I tried adding mbed-events from a uVisor box, but all hell breaks loose. Random crashes etc.

janjongboom:
@AlessandroA ping

docs/api/manual/Flash.md
```diff
@@ -0,0 +1,136 @@
...
+
+```cpp
+    {(void*)0xff000,   SECTOR_SIZE,    UVISOR_TACL_ACCESS | UVISOR_TACL_SHARED},
+```
+
+When you now re-compile the program runs to completion, and you can both read and write to flash from your secure box.
+
+## Conclusion
+
+uVisor cannot just be used for isolating memory between boxes, or to block usage of certain peripherals. You can also use uVisor to block off reading parts of the flash storage from unauthorized boxes. Unfortunately, because writing to flash goes through the FTFE peripheral, you do not have the same fine-grained control over writing to flash. You can solve this by adding a specific secure box which owns the FTFE peripheral, and use a [public secure entry point](../QUICKSTART.md#Expose-public-secure-entry-points-to-the-secure-box) to control which boxes can write to (specific parts of) flash.
```
iriark01: "cannot just be used" means "one does not simply walk into Mordor"
You want "can be used not just to"

janjongboom:
@AlessandroA @meriac ping

janjongboom:
See #407 and #408.

PR #407, opened 2017-02-10
Add 'Expose public secure entry points to a secure box' section to the uVisor documentation

janjongboom:
@AlessandroA  @iriark01 

Follow up from #404.


janjongboom:
@AlessandroA ping

PR #409, opened 2017-02-14
docs: Add initial uVisor Onboarding Guide

Patater:
Preview available at https://github.com/Patater/uvisor/blob/onboard/docs/core/ONBOARD.md

@AlessandroA @niklas-arm


AlessandroA:
As this document will likely grow a lot in the future, I think we should stay away from colloquial sentences and stick to concise bullet points.

As for the structure, a long time ago we moved away from 80-col files for markdown because of a requirement from the docs team (apparently it was not rendered well or something like that). Now all of our docs are unwrapped. I prefer the 80-col limit much more, but it would now break consistency. What do you think? @Patater

Patater:
Consise bullet points are boring. We've got to have some style in our voice. The onboarding guide isn't really meant to be overall technical, and certainly the introduction needs to be informal and fun. There are technical portions, and where it is technical, we already do use bullet points where it makes sense to do so.

This wasn't written to be processed by the docs team, but for easy reading and editing among ourselves. If we do want it to be processed by the docs team at some point, we should file an issue to fix the 80 columns problem. It's so much easier to edit with the 80 column limit.

AlessandroA:
retest uvisor

AnotherButler:
@Patater Is this an internal document for our team to use, or is this technical documentation for contributors?

Patater:
@AnotherButler This technical documentation for all contributors. We keep internal documents internal.

PR #412, opened 2017-02-15
vMPU refactoring

AlessandroA:
This PR refactors the vMPU code that enumerates boxes.

The refactoring brings the following changes:

* Parts of `vmpu_enumerate_boxes` have been moved to other functions, which makes it easier to read the code.
* The code has been updated to reflect the migration from simple "per-box context + stack" model to "per-box BSS sections + stack", where the BSS sections contain multiple items (index, context, heap, RPC queues).
    * There is now a unified way to read and reason about the BSS sections, by using the `UvisorBssSections` struct.
* Added an optimization for ARMv7-M:
    * Boxes are enumerated based on the order that minimizes the SRAM usage.
    * If the total SRAM usage exceeds the available allocated space, then the error message which is printed before halting suggests the space to allocate to remove the memory overflow.

**Note**: This refactoring breaks backwards compatibility, in that it removes the magic pointer `uvisor_ctx`.

A `__uvisor_ctx` pointer is now introduced, which still points to the user context for each box — but it's now a `void *` pointer. This means that box developers will have to cast it to their own context structure.

Although this is a UX change, I believe it makes it more explicit (from a semantics perspective it's a bit confusing that `uvisor_ctx`, never explicitly declared, means something different in each box). This point is open to discussion, but keep in mind that in the future we will likely expose the whole `UvisorBoxIndex` instead of just the context pointer. Developers will need to be aware that the index is a box-agnostic structure of pointers that uVisor switches in and out when switching processes.

@Patater @niklas-arm

### DIFF
--- a/api/inc/rpc_exports.h
+++ b/api/inc/rpc_exports.h
@@ -20,6 +20,7 @@
 #include "api/inc/pool_queue_exports.h"
 #include "api/inc/uvisor_semaphore_exports.h"
 #include "api/inc/rpc_gateway_exports.h"
+#include "api/inc/vmpu_exports.h"
 
 typedef uint32_t (*TFN_Ptr)(uint32_t, uint32_t, uint32_t, uint32_t);
 
@@ -136,5 +137,24 @@ typedef struct uvisor_rpc_fn_group {
 typedef UVISOR_RPC_OUTGOING_MESSAGE_TYPE(UVISOR_RPC_OUTGOING_MESSAGE_SLOTS) uvisor_rpc_outgoing_message_queue_t;
 typedef UVISOR_RPC_INCOMING_MESSAGE_TYPE(UVISOR_RPC_INCOMING_MESSAGE_SLOTS) uvisor_rpc_incoming_message_queue_t;
 typedef UVISOR_RPC_FN_GROUP_TYPE(UVISOR_RPC_FN_GROUP_SLOTS) uvisor_rpc_fn_group_queue_t;
+
+typedef struct uvisor_rpc_t {
+    /* Outgoing message queue */
+    uvisor_rpc_outgoing_message_queue_t outgoing_message_queue;
+
+    /* Incoming message queue */
+    uvisor_rpc_incoming_message_queue_t incoming_message_queue;
+
+    /* Function group queue */
+    uvisor_rpc_fn_group_queue_t fn_group_queue;
+
+    /* Counter to avoid waiting on the same RPC result twice by accident. */
+    uint32_t result_counter;
+} uvisor_rpc_t;
+
+static inline uvisor_rpc_t * uvisor_rpc(UvisorBoxIndex * const index)
+{
+    return (uvisor_rpc_t *) index->bss.address_of.rpc;
+}
 
 #endif
--- a/api/inc/vmpu_exports.h
+++ b/api/inc/vmpu_exports.h
@@ -153,59 +153,65 @@ typedef struct {
     UvisorBoxAcl acl;
 } UVISOR_PACKED UvisorBoxAclItem;
 
+/* This struct contains all the BSS sections that uVisor allocates for a secure
+ * box. It can be used to keep the sizes of the sections or their pointers. */
+typedef struct uvisor_bss_sections_t {
+    uint32_t index;
+    uint32_t context;
+    uint32_t rpc_outgoing_message_queue;
+    uint32_t rpc_incoming_message_queue;
+    uint32_t rpc_fn_group_queue;
+    uint32_t heap;
+} UVISOR_PACKED UvisorBssSections;
+
+/* The number of per-box BSS sections. */
+#define UVISOR_BSS_SECTIONS_COUNT (sizeof(UvisorBssSections) / sizeof(uint32_t))
+
+/* Compile-time per-box configuration table
+ * Each box has one of this table in flash. Every other data structure that this
+ * table might point to must be in flash as well. The uVisor core must check the
+ * sanity of the table before trusting its fields. */
 typedef struct {
-    /* Contains user provided size of box context without guards of buffers. */
-    uint32_t context_size;
-    /* Contains total memory used by the RPC queues (incl. management and pool). */
-    uint32_t rpc_outgoing_message_size;
-    uint32_t rpc_incoming_message_size;
-    uint32_t rpc_fn_group_size;
-} UVISOR_PACKED uvisor_sizes_t;
+    const uint32_t magic;
+    const uint32_t version;
 
-/* The number of additional bss sections per box bss.
- * The size of each section is stored in the box config, and uVisor core will
- * iterate over the box bss, split it into sections as defined by the size table
- * and assign a pointer to beginning of that section into the box index pointer table.
- */
-#define UVISOR_BOX_INDEX_SIZE_COUNT (sizeof(uvisor_sizes_t) / sizeof(uint32_t))
-
-typedef struct {
-    uint32_t magic;
-    uint32_t version;
-
-    /* Box stack size includes stack guards and rounding buffer. */
-    uint32_t stack_size;
-    /* Contains user provided size of box heap without guards of buffers. */
-    uint32_t heap_size;
-    /* Contains the size of the index (must be at least sizeof(UvisorBoxIndex)). */
-    uint32_t index_size;
-
+    /* The UvisorBssSections struct is union-ed with a size_t array to allow for
+     * loops to scan the sizes of all the BSS sections and allocate the
+     * necessary space for each of them. */
     union {
-        uint32_t bss_size[UVISOR_BOX_INDEX_SIZE_COUNT];
-        uvisor_sizes_t sizes;
-    };
+        size_t sizes[UVISOR_BSS_SECTIONS_COUNT];
+        UvisorBssSections size_of;
+    } const bss;
+
+    /* Contains the size of the secure box static stack. */
+    /* Note: This does not include guards. */
+    /* Note: It is kept separately from the BSS sections as it's implementation
+     *       specific where the stack sits with respect to the BSS. */
+    const uint32_t stack_size;
 
     /* Opaque-to-uVisor data that potentially contains uvisor-lib-specific or
      * OS-specific per-box configuration */
     const void * const lib_config;
 
-    const char * box_namespace;
+    const char * const box_namespace;
     const UvisorBoxAclItem * const acl_list;
-    uint32_t acl_count;
+    const uint32_t acl_count;
 } UVISOR_PACKED UvisorBoxConfig;
 
+/* Enumeration-time per-box index table
+ * Each box has one of this table in SRAM. The index tables are initialized at
+ * box enumeration time and are then managed by the secure boxes themselves. */
+/* Note: Each box is able to read and write its own version of this table. Do
+ *       not trust these pointers in the uVisor core. */
 typedef struct {
+    /* The UvisorSramPointers struct is union-ed with a void * array to allow
+     * for loops to scan the pointers to all the SRAM sections and access them
+     * individually. */
     union {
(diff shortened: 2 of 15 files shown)


core/vmpu/src/armv7m/vmpu_armv7m.c
```diff
@@ -300,79 +300,97 @@ void vmpu_load_box(uint8_t box_id)
...
-    /* Check for correct context address alignment. Alignment needs to be a
-     * muiltiple of the size. */
-    if ((box_mem_pos & (size - 1)) != 0) {
-        box_mem_pos = (box_mem_pos & ~(size - 1)) + size;
-    }
+    DPRINTF("\tbox[%i] stack=%i bss=%i rounded=%i\n\r", box_id, stack_size, bss_size, region_size);
 
     /* Check if we have enough memory left. */
-    if ((box_mem_pos + size) > ((uint32_t) __uvisor_config.bss_boxes_end)) {
+    /* Note: This should never happen as memory overflow is checked at boxes
```
Patater: s/boxes/box

core/vmpu/src/armv7m/vmpu_armv7m.c
```diff
@@ -300,79 +300,97 @@ void vmpu_load_box(uint8_t box_id)
 
 extern int vmpu_region_bits(uint32_t size);
 
+/* Compute the MPU region size for the given BSS sections and stack sizes.
+ * The function also updates the region_start parameter to meet the alignment
+ * requirements of the MPU. */
+static uint32_t vmpu_acl_stack_region_size(uint32_t * const region_start, uint32_t const bss_size,
```
Patater: Consider removing `const` from `uint32_t * const region_start`, to make explicit that `region_start` will be modified. As written, it requires a careful eye to understand where the `const` applies to, as it isn't in a form that coders expect and can sight-read easily. And `const` for value types isn't really that important to state explicitly, anyway. It's worth the readability improvement to remove 'const' here: be obvious and not subtle.

core/vmpu/src/armv7m/vmpu_armv7m.c
```diff
@@ -518,3 +513,76 @@ void vmpu_arch_init_hw(void)
...
+        swap_boxes(&box_order[start], &box_order[i]);
+        min_size = __vmpu_order_boxes(box_order, best_order, start + 1, end, min_size);
+        swap_boxes(&box_order[start], &box_order[i]);
+    }
+    return min_size;
+}
+
+void vmpu_order_boxes(int * const best_order, int box_count)
+{
+    /* Start with a the same order of the configuration table pointers. */
```
Patater: s/a the same order of the configuration/the same order of configuration

core/vmpu/inc/vmpu.h
```diff
@@ -161,8 +161,8 @@ extern int vmpu_fault_recovery_bus(uint32_t pc, uint32_t sp, uint32_t fault_addr
 
 uint32_t vmpu_fault_find_acl(uint32_t fault_addr, uint32_t size);
 
-extern void vmpu_acl_stack(uint8_t box_id, uint32_t bss_size, uint32_t stack_size, uint32_t * bss_start,
-                           uint32_t * stack_pointer);
+extern void vmpu_acl_sram(uint8_t box_id, uint32_t bss_size, uint32_t stack_size, uint32_t * bss_start,
```
Patater: Prefer function names that are verbs.

Was this stacking an ACL region before? Or was it adding the stack to the ACL?

Consider using `vmpu_acl_sram_push`, if there is some stacking semantics. Otherwise, `vmpu_acl_sram_add` would be good.

core/vmpu/src/vmpu.c
```diff
@@ -215,6 +215,43 @@ static void vmpu_sanity_checks_box_namespace(int box_id, const char *const box_n
     } while (box_namespace[length]);
 }
 
+static void vmpu_sanity_checks_box_cfgtbl(uint8_t box_id, UvisorBoxConfig const * box_cfgtbl)
```
AlessandroA: Done

core/vmpu/src/vmpu.c
```diff
@@ -176,7 +176,7 @@ static int vmpu_sanity_checks(void)
     }
 }
 
-static void vmpu_sanity_check_box_namespace(int box_id, const char *const box_namespace)
+static void vmpu_sanity_checks_box_namespace(int box_id, const char *const box_namespace)
```
AlessandroA: Done

api/inc/vmpu_exports.h
```diff
@@ -153,59 +153,65 @@ typedef struct {
     UvisorBoxAcl acl;
 } UVISOR_PACKED UvisorBoxAclItem;
 
+/* This struct contains all the BSS sections that uVisor allocates for a secure
+ * box. It can be used to keep the sizes of the sections or their pointers. */
+typedef struct uvisor_bss_sections_t {
```
AlessandroA: This structure is used both to hold sizes:

```C
union {
    size_t sizes[UVISOR_BSS_SECTIONS_COUNT];
    UvisorBssSections size_of;
} const bss;
```

and pointers:

union {
    void * pointers[UVISOR_BSS_SECTIONS_COUNT];
    UvisorBssSections address_of;
} const bss;
```

so I used `uint32_t` as base type.

If you are referring to the RPC casting that would be needed anyway, even with a `void *`, but I think there are better ways to fix the RPC specific problem. The reason I didn't fix it now is because we will anyway need to change that code. What other casting were you thinking about?

File: api/inc/vmpu_exports.h [hunk 8]
```diff
@@ -153,59 +153,65 @@ typedef struct {
...
-        uvisor_sizes_t sizes;
-    };
+        size_t sizes[UVISOR_BSS_SECTIONS_COUNT];
+        UvisorBssSections size_of;
+    } const bss;
+
+    /* Contains the size of the secure box static stack. */
+    /* Note: This does not include guards. */
+    /* Note: This is kept separately from the BSS sections as it's
+     *       implementation specific where the stack sits wrt the BSS. */
```
AlessandroA: Done

core/vmpu/inc/vmpu.h
```diff
@@ -161,7 +161,8 @@ extern int vmpu_fault_recovery_bus(uint32_t pc, uint32_t sp, uint32_t fault_addr
 
 uint32_t vmpu_fault_find_acl(uint32_t fault_addr, uint32_t size);
 
-extern void vmpu_acl_stack(uint8_t box_id, uint32_t bss_size, uint32_t stack_size);
+extern void vmpu_acl_stack(uint8_t box_id, uint32_t bss_size, uint32_t stack_size, uint32_t * bss_start,
+                           uint32_t * stack_pointer);
```
AlessandroA: I don't think this is necessary as the state is now set exactly where the code has been moved to:

```C
    uint32_t bss_start = 0;
    uint32_t stack_pointer = 0;
    if (box_id == 0) {
        /* The public box does not need an explicit ACL. Instead, it uses the
         * default access permissions given by the background regions. */
        /* Note: This relies on the assumption that all supported MPU
         *       architectures allow us to setup a background region somehow. */
        bss_start = (uint32_t) __uvisor_config.heap_start;       <----------------- Previously using the state
        stack_pointer = __get_PSP();
    } else {
        uint32_t stack_size = box_cfgtbl->stack_size;
        vmpu_acl_stack(box_id, bss_size, stack_size, &bss_start, &stack_pointer);
    }

    /* Set the box state for the SRAM sections. */
    g_context_current_states[box_id].bss = bss_start;            <----------------- State now set here
    g_context_current_states[box_id].bss_size = bss_size;
    g_context_current_states[box_id].sp = stack_pointer;
```

so *technically* it is using the state (because the same local variable will be used to setup the state few lines later) (see `stack_pointer` for example).

core/vmpu/src/vmpu.c
```diff
@@ -248,6 +248,18 @@ static void vmpu_sanity_checks_box_cfgtbl(uint8_t box_id, UvisorBoxConfig const
...
+    if (box_id == 0) {
+        if (box_cfgtbl->stack_size != 0) {
+            HALT_ERROR(SANITY_CHECK_FAILED, "Box %i @0x%08X: The public box stack size must be set to 0 (found %uB).\r\n",
+                       box_id, (uint32_t) box_cfgtbl, box_cfgtbl->stack_size);
+        }
+        if (box_cfgtbl->bss.size_of.heap != 0) {
+            HALT_ERROR(SANITY_CHECK_FAILED, "Box %i @0x%08X: The public box heap size must be set to 0 (found %uB).\r\n",
+                       box_id, (uint32_t) box_cfgtbl, box_cfgtbl->bss.size_of.heap);
+        }
+    }
```
AlessandroA: Added a comment that should clarify it. See also the related comments in `vmpu_box_index_init`:

```C
        /* FIXME: At the moment we cannot just reliably use the size of the heap
         *        as provided by the configuration table, because the public box
         *        has a variable heap size that is not identified at compile
         *        time. We should remove this dependency. */
```

```C
        /* Set the heap pointer for the public box.
         * This value must be set manually because the previous loop will have
         * caught a zero heap size and thus set the pointer to NULL. */
```

core/vmpu/src/armv7m/vmpu_armv7m.c
AlessandroA: I would have preferred to keep `uint32_t * const region_start` (*region_start is a constant pointer to uint32_t*), but I see your point on readability.

Fixed.

core/vmpu/src/armv7m/vmpu_armv7m.c
AlessandroA: Fixed.

core/vmpu/src/armv7m/vmpu_armv7m.c
AlessandroA: Done.

core/vmpu/inc/vmpu.h
AlessandroA: This was actually just creating an ACL for the box private stack. Historically we only had a stack, then a small context was added, then the whole BSS sections. The name was mistakenly kept (so nothing to do with *stacking* stuff).

I agree with the "functions as verb" approach, but I would like to not change this specific one now. This would require fixing other functions in the file and in other vMPU drivers. I'd like to fix them when we introduce the v8-M driver.

api/inc/vmpu_exports.h
Patater: Can we use the real type of each element instead of `void *` or `uint32_t`?

core/vmpu/inc/vmpu.h
Patater: OK

core/vmpu/inc/vmpu.h
Patater: OK. Please file an issue. We'll prioritize and fix as desired.

api/inc/rpc_exports.h
```diff
@@ -137,4 +138,23 @@ typedef UVISOR_RPC_OUTGOING_MESSAGE_TYPE(UVISOR_RPC_OUTGOING_MESSAGE_SLOTS) uvis
...
+
+    /* Incoming message queue */
+    uvisor_rpc_incoming_message_queue_t incoming_message_queue;
+
+    /* Function group queue */
+    uvisor_rpc_fn_group_queue_t fn_group_queue;
+
+    /* Counter to avoid waiting on the same RPC result twice by accident. */
+    uint32_t result_counter;
+} uvisor_rpc_t;
```
Patater: Could these be part of the UvisorBoxIndex instead of their own thing?

To make such a change would involve a bit more work that I'd like us to do now, but potentially worth filing an issue about. The code would be cleaner with this change.

AlessandroA: Yes I agree, I will create an issue.

PR #419, opened 2017-03-21
Adding API page

iriark01:
It's not in the YML, meaning it's not built by the docs system


AlessandroA:
retest uvisor

iriark01:
Sorry, what?

AlessandroA:
@iriark01 It's an automated message to let the CI to retest uVisor :) The tests should all be green — We're currently looking into why they're not.

PR #421, opened 2017-04-07
Makefile: Build uVisor for M3, even on M4s

Patater:
Currently, uVisor builds with `-mcpu=cortex-m4`, even on Cortex M3 CPUs.
This can lead to using unavailable instructions. Build uVisor using the
least common denominator architecture. This avoids using unavailable
instructions when targeting Cortex M3 CPUs.

@AlessandroA @niklas-arm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,10 +140,10 @@ ifeq ("$(PROGRAM_VERSION)","")
 PROGRAM_VERSION:='unknown'
 endif
 
-FLAGS_CM4:=-mcpu=cortex-m4 -march=armv7e-m -mthumb
+FLAGS_CORE:=-mcpu=cortex-m3 -march=armv7-m -mthumb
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -140,7 +140,7 @@ ifeq ("$(PROGRAM_VERSION)","")
 PROGRAM_VERSION:='unknown'
 endif
 
-FLAGS_CM4:=-mcpu=cortex-m4 -march=armv7e-m -mthumb
+FLAGS_CM4:=-mcpu=cortex-m3 -march=armv7-m -mthumb
```
AlessandroA: Could you rename `FLAGS_CM4` to `FLAGS_CORE`? Other than that, LGTM.

Patater: Renamed

PR #429, opened 2017-05-30
Update for CMSIS-RTOS2 (RTX5) and ARMv8-M

Patater:
This first few commits get uVisor to work with RTX5. Then, the rest of the commits bring over all our ARMv8-M work, plus various refactoring and bug fixes that also affect ARMv7-M.

@AlessandroA @niklas-arm


Patater:
Each commit builds individually. Tests still run.

PR #431, opened 2017-06-05
Add box id for namespace api

niklas-arm:
Adds a cross-privilege memcmp function used for comparing the secure namespace with the user provided one.

cc @AlessandroA @Patater


Patater:
retest uvisor

Patater:
retest uvisor

PR #432, opened 2017-06-08
WIP RPC that uses IPC

Patater:
WIP for RPC (no gateway yet)

@AlessandroA @niklas-arm


Patater:
retest uvisor

Patater:
@p03189 No. The current API interprets a timeout of 0 as "let me know if it is done yet, without any blocking". Any other timeout value will be interpreted as "block forever until it is done".

p03189:
Oh, thank you.

PR #436, opened 2017-06-14
Remove definition of g_debug_box from core/debug/inc/debug.h

mikisch81:
`g_debug_box` will now be declared in _core/debug/inc/debug.h_ and defined in _core/debug/src/debug_box.c_.
@Patater @alzix


mikisch81:
retest uvisor

PR #439, opened 2017-06-16
Don't merge: setting up for the new docs engine

iriark01:
Please don't merge this; we're setting up for the new docs engine


iriark01:
Got it; thanks

PR #441, opened 2017-06-27
Point at latest uvisor-tests

Patater:
This enables uvisor and uvisor-tests to stay synchronized.


Patater:
Merged as part of #444

PR #445, opened 2017-07-04
add eclipse project helper

alzix:
useful for instrumenting example projects with Eclipse+PyOCD debug environment
this script should be called from from `mbed-os/features/FEATURE_UVISOR/importer/TARGET_IGNORE/uvisor` within example project

### DIFF
--- a/tools/eclipse_prj_helper/Makefile.template
+++ b/tools/eclipse_prj_helper/Makefile.template
@@ -1,0 +1,47 @@
+###########################################################################
+#
+#  Copyright (c) 2013-2017, ARM Limited, All Rights Reserved
+#  SPDX-License-Identifier: Apache-2.0
+#
+#  Licensed under the Apache License, Version 2.0 (the "License"); you may
+#  not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+###########################################################################
+
+###########################################################################
+#            DO NOT EDIT - This is an auto-generated file                 #
+###########################################################################
+# generated from:
+#   $template_file
+
+
+WORKSPACE_DIR := $workspace_dir
+UVISOR_ROOT_DIR := $uvisor_dir
+IMPORTER_DIR := $importer_dir
+
+PLATFORM := $target
+TOOLCHAIN := $toolchain
+
+
+
+.PHONY: all clean
+all:
+	@echo Build uVisor
+	@test -e $(UVISOR_ROOT_DIR) && $(MAKE) -C $(UVISOR_ROOT_DIR)
+	@echo Publish uVisor build artifacts
+	@test -e $(UVISOR_ROOT_DIR) && $(MAKE) -C $(IMPORTER_DIR) publish
+	@echo Build Workspace
+	@cd $(WORKSPACE_DIR) && mbed compile --profile debug -m $(PLATFORM) -t $(TOOLCHAIN)
+clean:
+	@echo Cleaning Workspace
+	@$(MAKE) -C $(UVISOR_ROOT_DIR) clean
+	@rm -rf $(WORKSPACE_DIR)/BUILD
--- a/tools/eclipse_prj_helper/generate_prj.py
+++ b/tools/eclipse_prj_helper/generate_prj.py
@@ -1,0 +1,157 @@
+#!/usr/bin/env python
+###########################################################################
+#
+#  Copyright (c) 2013-2017, ARM Limited, All Rights Reserved
+#  SPDX-License-Identifier: Apache-2.0
+#
+#  Licensed under the Apache License, Version 2.0 (the "License"); you may
+#  not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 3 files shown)


alzix:
retest uvisor

tools/eclipse_prj_helper/Makefile.template
```diff
@@ -0,0 +1,47 @@
...
+
+
+.PHONY: all clean
+all:
+	@echo Build uVisor
+	@test -e $(UVISOR_ROOT_DIR) && $(MAKE) -C $(UVISOR_ROOT_DIR)
+	@echo Publish uVisor build artifacts
+	@test -e $(UVISOR_ROOT_DIR) && $(MAKE) -C $(IMPORTER_DIR) publish
+	@echo Build Workspace
+	@cd $(WORKSPACE_DIR) && mbed compile --profile debug -m $(PLATFORM) -t $(TOOLCHAIN)
```
Patater: If the uVisor build fails, can refrain from building the application?

theotherjimmy:
Hello,

You could modify the exporter to check for 
```python
"UVISOR" in targets.features
```
and then use a different template.
But I don't like having anything special for features, anywhere.

@alzix @Patater Could you let me know what would be required to make an mbed-os exporter (I'm guessing make-gcc-arm and therefore eclipse-gcc-arm) Uvisor aware?

tools/eclipse_prj_helper/generate_prj.py
```diff
@@ -0,0 +1,157 @@
...
+        fh.write(
+            makefile_template.safe_substitute(
+                workspace_dir=workspace,
+                uvisor_dir=uvisor_dir,
+                importer_dir=importer_dir,
+                target=target,
+                toolchain=toolchain,
+                template_file=template_file
+            )
+        )
```
theotherjimmy: I would have written this with Jinja 2, but that's just because I wrote many exporters using Jinja 2.
It has a few advantages:
 * it does not use the same syntax as Make
 * loops
 * Exceptions when it tries to use an undefined variable

tools/eclipse_prj_helper/Makefile.template
theotherjimmy: You could rewrite the make rule to use multiple rules. For example:
```make
UVISOR-SOURCES := $(wildcard $(UVISOR_ROOT_DIR)/*)
uvisor-executable: $(UVISOR-SOURCES)
	@echo Build uVisor
	@test -e $(UVISOR_ROOT_DIR) && $(MAKE) -C $(UVISOR_ROOT_DIR)
    @touch uvisor-executable

uvisor-publish: uvisor-executable
	@echo Publish uVisor build artifacts
	@test -e $(UVISOR_ROOT_DIR) && $(MAKE) -C $(IMPORTER_DIR) publish
    @touch uvisor-publish

workspace-sources=$(wildcard $(WORKSPACE_DIR)/*)
all: uvisor-publish $(workspace-sources)
	@echo Build Workspace
	@cd $(WORKSPACE_DIR) && mbed compile --profile debug -m $(PLATFORM) -t $(TOOLCHAIN)

theotherjimmy: Aside: Calling make in make is generally a sign that something has gone horribly wrong with your makefile.

theotherjimmy:
Looking at the changes here, you call mbed CLI. That sort of thing is not allowed in any exporter.

tools/eclipse_prj_helper/Makefile.template
alzix: @Patater - the make file will fail - in case any of it is recipes will fail. Thus application wont be built in case uVisor build was not successful.
@theotherjimmy - calling make from make is requirement for rebuilding uVisor sources - which happens to have make driven build system.
@theotherjimmy - regarding the purposed makefile change. The purposed changes has following flaws comparing to original ones:
1. the purposed change tries to mimic mbed-cli dependencies - mbed-cli is fully capable to manage its own build dependencies, which i tend to believe happen to more complicated than a simple glob. 
2. in case mbed-cli does rebuild the binary on every build regardless to its dependencies - the fix should be applied to mbed-cli rather than usage scripts.
3. the purposed change introduces dependencies which will not be honored in parralel build use case (-j) as there is no guarantee that `uvisor-publish` will be executed after `uvisor-executable`

tools/eclipse_prj_helper/generate_prj.py
alzix: @theotherjimmy - there is always more than one way for doing stuff :)
Personally I do prefer the "right tool for the right job" approach. Jinja seems to be an overkill for such a simple task - since this one is a simple, straight forward string substitution. There is no complex logic involved requiring conditional statements and/or loops.
BTW, Template.safe_substitute does rais KeyError exception in case placeholders are missing.

alzix:
@theotherjimmy - i'm familiar with general approach in mbed-cli export tools, where generated workspaces do not rely on the mbed-cli. But in current case i had following justifications:

- uVisor binary produced by exported Makefile is somehow different and is not working correctly for debug build - we will look in to this later. We have dedicated Ticket for that.
- the makefile is static - meaning that adding new sources/headers will require workspace rebuild. While when using mbed-cli - it dynamically picks up new files. Since this workspace is intended to be used for  development purposes this is a significant benefit.
- having two build systems is error prone approach - things working locally may/may not work on CI builds. I think the best example for this would be the old Eclipse driven Android Studio, where builds where driven by Eclipse or Ant for CI builds. This solution was pretty fast replaced by IntelliJ in favor of unified build system (Gradle).
- mbed-cli availability is on developers machines is proven by the fact `mbed export` and `mbed deploy` commands should be executed before the build.

alzix:
@Patater - commit message reworded

theotherjimmy:
> mbed-cli availability is on developers machines is proven by the fact mbed export and mbed deploy commands should be executed before the build.

This is simply not true. Anyone can export from the developer website without installing any tools. That's why you can't rely on mbed CLI.

PR #447, opened 2017-07-11
[WIP] do not carsh on BKPT during semihosting in case debugger is not connected

alzix:
as it will result in HardFault 

this change will allow IoT application developers to test their application against debug version of mbed-os without forcing them to connect the debugger with semihosting

references: 

- http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.faqs/ka14536.html
- 14.2 Breakpoint feature (The Definitive Guide to Cortex-M3 and Cortex-M4 Processors)

Tested - debug build is running on K64F without the debugger being attached


Patater:
retest uvisor

niklas-arm:
Fixed by https://github.com/ARMmbed/uvisor/pull/466 instead, please close.

PR #449, opened 2017-07-13
pool_queue.c: Check return value of pool_init

thomasfangyi:
Pass on the error value to uvisor_pool_queue_init if the pool fails to init.


core/system/src/pool_queue.c
```diff
@@ -51,7 +51,9 @@ int uvisor_pool_init(uvisor_pool_t * pool, void * array, size_t stride, size_t n
 
 int uvisor_pool_queue_init(uvisor_pool_queue_t * pool_queue, uvisor_pool_t * pool, void * array, size_t stride, size_t num)
 {
-    uvisor_pool_init(pool, array, stride, num);
+    int pool_init_ret = uvisor_pool_init(pool, array, stride, num);
+    if (pool_init_ret)
```
Patater: Please add braces.

PR #450, opened 2017-07-17
IPC: Check whether wait_tokens are allocated before wait

thomasfangyi:
When wait_tokens are unallocated, it is possible that wait never returns
or it depends on a future ipc call that has yet to be created. I believe
this validation on input parameter should be necessary.

### DIFF
--- a/api/src/ipc.c
+++ b/api/src/ipc.c
@@ -111,8 +111,13 @@ static int ipc_waitfor(int (*cond)(uint32_t have, uint32_t expect), uint32_t wai
     uint32_t completed_tokens;
     bool condition_met;
     for (;;) {
-        /* Read the current tokens and clear any we were waiting on. */
         uvisor_spin_lock(ipc_tokens_lock());
+        /* Check we are not waiting for some unallocated tokens */
+        if ((wait_tokens & *ipc_allocated_tokens()) != wait_tokens) {
+            uvisor_spin_unlock(ipc_tokens_lock());
+            return UVISOR_ERROR_INVALID_PARAMETERS;
+        }
+        /* Read the current tokens and clear any we were waiting on. */
         completed_tokens = *ipc_completed_tokens() & wait_tokens;
         condition_met = cond(completed_tokens, wait_tokens);
         if (condition_met) {


api/src/ipc.c
```diff
@@ -108,6 +108,10 @@ static void ipc_free_allocated_completed_tokens(uint32_t tokens)
 
 static int ipc_waitfor(int (*cond)(uint32_t have, uint32_t expect), uint32_t wait_tokens, uint32_t * done_tokens, uint32_t timeout_ms)
 {
+    /* Check we are not waiting for some unallocated tokens */
+    if ((wait_tokens & *ipc_allocated_tokens()) != wait_tokens) {
```
Patater: Reading of allocated tokens is not atomic and must be synchronized with the allocation of tokens. Consider using the tokens lock while reading the allocated tokens.

api/src/ipc.c
```diff
@@ -108,6 +108,13 @@ static void ipc_free_allocated_completed_tokens(uint32_t tokens)
 
 static int ipc_waitfor(int (*cond)(uint32_t have, uint32_t expect), uint32_t wait_tokens, uint32_t * done_tokens, uint32_t timeout_ms)
 {
+    uvisor_spin_lock(ipc_tokens_lock());
+    /* Check we are not waiting for some unallocated tokens */
+    if ((wait_tokens & *ipc_allocated_tokens()) != wait_tokens) {
+        uvisor_spin_unlock(ipc_tokens_lock());
+        return UVISOR_ERROR_INVALID_PARAMETERS;
+    }
+    uvisor_spin_unlock(ipc_tokens_lock());
```
Patater: A token might get freed or allocated between here and when we read the completed tokens. This block of code doesn't prevent waiting on tokens being allocated in all cases due to this race condition. The way to avoid this would be to hold the lock the entire time one is spinning, but that would prevent any change to the completed tokens and make the waitfor useless.

api/src/ipc.c
thomasfangyi: I doubt whether reading allocated tokens need synchronisation. Shouldn't allocated tokens be a thread-local data structure? Allocation and freeing is limited to the thread. Completed tokens are shared between multiple threads.

thomasfangyi: (Based on the assumption that token allocation and free should be in static linkage and not an exposed api)

Patater: We don't have thread-local storage, unfortunately.

thomasfangyi: I see. So each box can have multiple threads so it's necessary to lock the tokens.

api/src/ipc.c
Patater: Holding the lock while reading both allocated and completed tokens is sufficient.

PR #451, opened 2017-07-17
IPC: Make token allocation/free static

thomasfangyi:
The two functions should be used for internal linkage, as no external
usage for those could be found and they are internal to IPC call APIs.
Exposing these functions for external use might also cause
synchronisation issues.


thomasfangyi:
#452

PR #452, opened 2017-07-17
IPC: Minors fixes

thomasfangyi:
Update comments and fix styles.

### DIFF
--- a/api/src/ipc.c
+++ b/api/src/ipc.c
@@ -92,9 +92,9 @@ uint32_t ipc_allocate_token(uint32_t * tokens)
 }
 
 /** Free tokens
+ *  The token lock must have been already acquired
  * @param    tokens[in,out] free the specified tokens_to_free from tokens, atomically modifying tokens
- * @param    tokens_to_free[in] the tokens to free (represented by 1 bits)
- * @return   the token, or 0 if no token available */
+ * @param    tokens_to_free[in] the tokens to free (represented by bits) */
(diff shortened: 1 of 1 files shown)


Patater:
We can't "Make token allocation/free static" because then our tests can't run. Our tests depend on being able to use the token allocator (with their own token memory, not the real box ones) in order to unit test the token allocator.

Could you please drop the commit "Make token allocation/free static"?

Thanks

thomasfangyi:
I've removed the commit.

In a new PR #457 I introduced a new token allocation method which probably breaks some existing tests. Any suggestions on how to resolve this?

api/src/ipc.c
```diff
@@ -92,10 +92,10 @@ uint32_t ipc_allocate_token(uint32_t * tokens)
 }
 
 /** Free tokens
+ *  The token lock must have been already acquired
  * @param    tokens[in,out] free the specified tokens_to_free from tokens, atomically modifying tokens
- * @param    tokens_to_free[in] the tokens to free (represented by 1 bits)
- * @return   the token, or 0 if no token available */
-void ipc_free_tokens(uint32_t * tokens, uint32_t tokens_to_free)
+ * @param    tokens_to_free[in] the tokens to free (represented by bits) */
+static void ipc_free_tokens(uint32_t * tokens, uint32_t tokens_to_free)
```
Patater: This can't be static, due to being used directly by tests.

thomasfangyi: that was a miss. I'm fixing it.

PR #462, opened 2017-07-26
Handle 128th character dropping issue

orenc17A:


niklas-arm:
One could probably also implement this by checking length against `(DEBUG_MAX_BUFFER - 2)` and then not overwriting the data. But this works fine too. cc @Patater.

Patater:
Yeah, I'd find that easier to read than the current implementation.

orenc17A:
retest uvisor

orenc17A:
@niklas-arm  @Patater 
could you please review the new implementation

orenc17A:
Done

PR #466, opened 2017-07-31
Enable debug print control

theamirocohen:
fix issue: K64F debug version is not working without debugger being attached

### DIFF
--- a/api/inc/api.h
+++ b/api/inc/api.h
@@ -33,6 +33,11 @@
 UVISOR_EXTERN_C_BEGIN
 
 extern void uvisor_init(void);
+/* This method is an api method which initializes the g_semihosting_magic variable.
+ * This function will only work if called in privileged mode or from secure code,
+ * it is only intended to be called by debugger after reset.
+ */
+void debug_semihosting_enable(void);
 
 typedef struct {
     uint32_t magic;
@@ -84,6 +89,8 @@ typedef struct {
     bool (*spin_trylock)(UvisorSpinlock * spinlock);
     void (*spin_lock)(UvisorSpinlock * spinlock);
     void (*spin_unlock)(UvisorSpinlock * spinlock);
+
+    void (*debug_semihosting_enable)(void);
 
     OsEventObserver os_event_observer;
 } UVISOR_PACKED UvisorApi;
--- a/core/debug/src/debug.c
+++ b/core/debug/src/debug.c
@@ -37,27 +37,39 @@
  * overwritten to provide a different mechanism for printing messages (e.g. the
  * UART port instead of semihosting). */
 #define DEBUG_MAX_BUFFER 128
-uint8_t g_buffer[DEBUG_MAX_BUFFER];
-int g_buffer_pos = 0;
+#define DEBUG_SEMIHOSTING_MAGIC 0xDEADD00D
+
+static uint8_t g_buffer[DEBUG_MAX_BUFFER];
+static uint32_t g_buffer_pos = 0;
+
+__attribute__((section(".uninitialized"))) static uint32_t g_semihosting_magic;
+
+void debug_semihosting_enable(void)
+{
+    g_semihosting_magic = DEBUG_SEMIHOSTING_MAGIC;
+}
 
 UVISOR_WEAK void default_putc(uint8_t data)
 {
-    g_buffer[g_buffer_pos++] = data;
-    if (g_buffer_pos == (DEBUG_MAX_BUFFER - 1)) {
-        data = '\n';
-    }
+    if (DEBUG_SEMIHOSTING_MAGIC == g_semihosting_magic) {
 
-    if (data == '\n') {
-        g_buffer[g_buffer_pos] = 0;
-        asm volatile(
-            "mov r0, #4\n"
-            "mov r1, %[data]\n"
-            "bkpt #0xAB\n"
-            :
-            : [data] "r" (&g_buffer)
-            : "r0", "r1"
-        );
-        g_buffer_pos = 0;
+        g_buffer[g_buffer_pos++] = data;
+        if (g_buffer_pos == (DEBUG_MAX_BUFFER - 1)) {
+            data = '\n';
+        }
+
+        if (data == '\n') {
+            g_buffer[g_buffer_pos] = 0;
+            asm volatile(
+                "mov r0, #4\n"
+                "mov r1, %[data]\n"
+                "bkpt #0xAB\n"
+                :
+                : [data] "r" (&g_buffer)
+                : "r0", "r1"
+            );
+            g_buffer_pos = 0;
+        }
     }
 }
 
--- a/core/linker/default.h
+++ b/core/linker/default.h
@@ -155,4 +155,14 @@ SECTIONS
         . += STACK_GUARD_BAND;
         __uvisor_stack_end__ = .;
     } > STACK_S
+
+    .uninitialized (NOLOAD):
+    {
+        . = ALIGN(32);
+        __uninitialized_start = .;
+        *(.uninitialized)
+        KEEP(*(.keep.uninitialized))
+        . = ALIGN(32);
+        __uninitialized_end = .;
+    } > RAM_S
 }
--- a/core/system/src/api.c
+++ b/core/system/src/api.c
@@ -171,4 +171,6 @@ const UvisorApi __uvisor_api = {
         .thread_destroy = thread_destroy_transition,
         .thread_switch = thread_switch_transition,
     },
+
+    .debug_semihosting_enable = debug_semihosting_enable,
 };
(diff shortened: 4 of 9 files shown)


core/debug/src/debug.c
```diff
@@ -37,27 +37,40 @@
...
+static uint32_t g_buffer_pos = 0;
+
+__attribute__((section(".uninitialized"))) static uint32_t g_semihosting_magic;
+
+/* This method is an api method which initialize the g_semihosting_magic variable,
+ * when debugging this api should be called through gdb
+ */
+void debug_semihosting_enable(void)
+{
+	g_semihosting_magic = 0xDEADD00D;
```
niklas-arm: Put the magic value into a define for less duplication.

docs/lib/DEBUGGING.md
```diff
@@ -73,7 +73,9 @@ The following is the minimum set of commands you need to send to the device to f
 (gdb) monitor flash device = ${device_name}
 (gdb) load BUILD/${target}/${your_app}.elf
 (gdb) file BUILD/${target}/${your_app}.elf
+(gdb) call uvisor_api.debug_semihosting_enable()
 ```
+the call uvisor_api.debug_semihosting_enable command is necessary in order to enable semihosting debug printing.
```
niklas-arm:
```diff
- the call uvisor_api.debug_semihosting_enable command is necessary in order to enable
+ The call to `uvisor_api.debug_semihosting_enable` is required to enable
```

docs/lib/DEBUGGING.md
```diff
@@ -197,4 +199,4 @@ This board provides both an on-board proprietary debugging port (ST-LINK) and a
 
 If you are using ST-LINK, please refer to the [STMicroelectronics website](http://www.st.com/web/catalog/tools/FM146/CL1984/SC724/SS1677/PF251168?sc=internet/evalboard/product/251168.jsp) for information about the tools and drivers you need. Please note we have not tested this debugger with uVisor.
 
-If instead you want to connect your debugger to the JTAG port, you must wire the needed pins to your connector. This [guide](https://www.segger.com/admin/uploads/evalBoardDocs/AN00015_ConnectingJLinkToSTM32F429Discovery.pdf) explains how to do that in detail. The guide is specific to the J-Link connectors, but you can apply it to other connectors.
\ No newline at end of file
+If instead you want to connect your debugger to the JTAG port, you must wire the needed pins to your connector. This [guide](https://www.segger.com/admin/uploads/evalBoardDocs/AN00015_ConnectingJLinkToSTM32F429Discovery.pdf) explains how to do that in detail. The guide is specific to the J-Link connectors, but you can apply it to other connectors.
```
niklas-arm: I dont' see a difference here, is it a different line ending?

core/debug/src/debug.c
```diff
@@ -37,27 +37,40 @@
...
-int g_buffer_pos = 0;
+static uint8_t g_buffer[DEBUG_MAX_BUFFER];
+static uint32_t g_buffer_pos = 0;
+
+__attribute__((section(".uninitialized"))) static uint32_t g_semihosting_magic;
+
+/* This method is an api method which initialize the g_semihosting_magic variable,
+ * when debugging this api should be called through gdb
+ */
+void debug_semihosting_enable(void)
```
niklas-arm: `extern` declare this function in a header file to avoid forward declaring it manually later on.

core/system/src/api.c
```diff
@@ -111,6 +111,8 @@ static uint32_t get_api_version(uint32_t version)
     return UVISOR_API_VERSION;
 }
 
+void debug_semihosting_enable(void);
```
niklas-arm: Move to header file.

core/system/src/api.c
```diff
@@ -171,4 +173,6 @@ const UvisorApi __uvisor_api = {
         .thread_destroy = thread_destroy_transition,
         .thread_switch = thread_switch_transition,
     },
+
+    .debug_semihosting_enable = debug_semihosting_enable,
```
niklas-arm: So, just to clarify to myself: This function will only work if called in privileged mode or from secure code. This is fine, since it is supposed to be called by the debugger after reset. Nice design!

core/system/src/core_armv7m/svc.c
```diff
@@ -37,6 +37,8 @@ void __svc_not_implemented(void)
     HALT_ERROR(NOT_IMPLEMENTED, "function not implemented\n\r");
 }
 
+void debug_semihosting_enable(void);
```
niklas-arm: Move to header file.

core/system/src/api.c
alzix: i think we should add this note to the function declaration in the header file

api/inc/api.h
```diff
@@ -85,6 +85,8 @@ typedef struct {
     void (*spin_lock)(UvisorSpinlock * spinlock);
     void (*spin_unlock)(UvisorSpinlock * spinlock);
 
+    void (*debug_semihosting_enable)(void);
```
alzix: please add comment stating that it is ONLY intended to be called by debugger after reset

docs/lib/DEBUGGING.md
theamirocohen: didn't change anything, might be some indentations

Patater:
Don't mention Jira tickets in the commit message, since that's an internal tracking reference not useful to those outside of ARM.

docs/lib/DEBUGGING.md
Patater: Please configure your text editor to not remove newlines at the ends of files.

core/linker/default.h
```diff
@@ -76,6 +76,7 @@ MEMORY
 
 SECTIONS
 {
+
```
Patater: Don't add newline here.

core/linker/default.h
```diff
@@ -155,4 +156,14 @@ SECTIONS
...
+    .uninitialized (NOLOAD):
+    {
+        . = ALIGN(32);
+        __uninitialized_start = .;
+        *(.uninitialized)
+        KEEP(*(.keep.uninitialized))
+        . = ALIGN(32);
+        __uninitialized_end = .;
+    } > RAM_S
```
Patater: Is it intended that the uninitialized section will not be modifiable by unprivileged code (via MPU protection)?

theamirocohen: yep, it is called by the debugger after reset in privileged mode or from secure code only.

Patater:
Please squash commits.

core/linker/default.h
Patater: Have you verified that a test or example application can't set this value?

theamirocohen: The symbol is not exposed to the application so cannot be set, but if, for some reason, the application might guess the value's address I've verified that the address cannot be addressed (Bus Fault triggered).

Patater: Yup, I was concerned about the address "guessing" part. Good that we get an access denied fault.

Patater:
Your commit message explicitly mentions K64F, but doesn't this work for all platforms and not just K64F?

theamirocohen:
We know for sure that in K64F there is a problem probing the debug bit, it is set with or without semihosting. 
This fix will automatically apply to all platforms.

Patater:
Please update the commit message to remove K64F, as the commit updates all targets. Please also explain the motivation for this change in the commit message.

https://docs.mbed.com/docs/mbed-os-handbook/en/latest/cont/code_style/#guidelines-for-github-pull-requests

Patater:
`6. Wrap the body at 72 characters.`
-- https://docs.mbed.com/docs/mbed-os-handbook/en/latest/cont/code_style/#guidelines-for-github-pull-requests

api/inc/api.h
```diff
@@ -33,6 +33,11 @@
 UVISOR_EXTERN_C_BEGIN
 
 extern void uvisor_init(void);
+/*This method is an api method which initialize the g_semihosting_magic variable,
```
Patater: Add a space after `*`

Patater:
```diff
- initialize
+ initializes
```

api/inc/api.h
```diff
@@ -33,6 +33,11 @@
...
+ * This function will only work if called in privileged mode or from secure code,
+ * it is only intended to be called by debugger after reset
```
Patater: Add a period after `reset`.

```diff
- reset
+ reset.
```

api/inc/api.h
```diff
@@ -33,6 +33,11 @@
 UVISOR_EXTERN_C_BEGIN
 
 extern void uvisor_init(void);
+/* This method is an api method which initializes the g_semihosting_magic variable,
```
Patater:
```diff
- This method is an api method which initializes the g_semihosting_magic variable,
+ This method is an api method which initializes the g_semihosting_magic variable.
```

PR #468, opened 2017-07-31
vmpu: Check start_addr is no greater than end_addr first

thomasfangyi:
If the check is not done first, it can be avoided at
`vmpu_buffer_access_is_ok_static` where there is not check for start is
smaller than end. Line 247 only checks range.

### DIFF
--- a/core/vmpu/src/mpu_armv7m/vmpu_armv7m_mpu.c
+++ b/core/vmpu/src/mpu_armv7m/vmpu_armv7m_mpu.c
@@ -380,6 +380,12 @@ bool vmpu_buffer_access_is_ok(int box_id, const void * addr, size_t size)
     uint32_t start_addr = (uint32_t) addr;
     uint32_t end_addr = start_addr + size - 1;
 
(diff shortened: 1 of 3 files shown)


core/vmpu/src/mpu_armv7m/vmpu_armv7m_mpu.c
```diff
@@ -397,11 +402,6 @@ bool vmpu_buffer_access_is_ok(int box_id, const void * addr, size_t size)
     }
 
     assert(start_addr <= end_addr);
```
Patater: Move this assert as well, before the if, for all implementations.

thomasfangyi:
I think the test failure was due to service outage of GitHub at the time of testing.

alzix:
retest uvisor

alzix:
this is a magic word for triggering CI to retest the PR

Patater:
It looks like the failure was with the K64F without uVisor test run timing out.

PR #469, opened 2017-07-31
pool_queue: Force termination of find_first

thomasfangyi:
When the pool is corrupted to contain a cycle, it would be bad to keep
the loop infinite because this creates a denial of service. The counter
will force termination of look_first to a finite number of iterations.

### DIFF
--- a/core/system/src/pool_queue.c
+++ b/core/system/src/pool_queue.c
@@ -361,10 +361,11 @@ static uvisor_pool_slot_t find_first(uvisor_pool_queue_t * pool_queue,
 {
     uvisor_pool_slot_t slot;
     uvisor_pool_t * pool = UVISOR_AUTO_NS_ALIAS(pool_queue->pool);
+    uvisor_pool_slot_t iterated = 0;
 
     /* Walk the queue, looking for the first slot that matches the query. */
     slot = pool_queue->head;
-    while (slot != UVISOR_POOL_SLOT_INVALID)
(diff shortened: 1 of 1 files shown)


core/system/src/pool_queue.c
```diff
@@ -361,10 +361,11 @@ static uvisor_pool_slot_t find_first(uvisor_pool_queue_t * pool_queue,
 {
     uvisor_pool_slot_t slot;
     uvisor_pool_t * pool = UVISOR_AUTO_NS_ALIAS(pool_queue->pool);
+    uvisor_pool_slot_t iterated = 0;
 
     /* Walk the queue, looking for the first slot that matches the query. */
     slot = pool_queue->head;
-    while (slot != UVISOR_POOL_SLOT_INVALID)
+    while (slot != UVISOR_POOL_SLOT_INVALID && iterated <= pool->num)
```
Patater: `pool->num` is user modifiable, so this doesn't really prevent DoS very well. Consider using UVISOR_POOL_MAX_VALID instead.

thomasfangyi: It wouldn't make much of a different because uint8_MAX is 255 whereas MAX_VALID is 252, but I'll do it anyways

Patater: Turns out `pool->num` is not a `uint32_t` or `size_t` as I had misremembered, so switching to `UVISOR_POOL_MAX_VALID` doesn't make too much of a difference.

PR #472, opened 2017-07-31
Remove RTX specific types

c1728p9:
Replace use of osRtxMutex_t with the generic mbed_rtos_storage_mutex_t.

### DIFF
--- a/api/rtx/inc/rtx_box_index.h
+++ b/api/rtx/inc/rtx_box_index.h
@@ -18,7 +18,7 @@
 #define __RTX_BOX_INDEX_H__
 
 #include "cmsis_os2.h"
-#include "rtx_os.h"
+#include "mbed_rtos_storage.h"
 
 #ifdef __cplusplus
 extern "C" {
@@ -36,7 +36,7 @@ typedef struct
     osMutexAttr_t mutex_attr;
 
     /* Internal data of the mutex */
-    osRtxMutex_t mutex_data;
+    mbed_rtos_storage_mutex_t mutex_data;
 } RtxBoxIndex;
 
 #ifdef __cplusplus


api/rtx/inc/rtx_box_index.h
```diff
@@ -36,7 +36,7 @@ typedef struct
     osMutexAttr_t mutex_attr;
 
     /* Internal data of the mutex */
-    osRtxMutex_t mutex_data;
+    mbed_rtos_storage_thread_t mutex_data;
```
Patater: Is there an `mbed_rtos_storage_mutex_t` we should be using?

Patater:
Please explain the motivation for the change in the commit message. I'm not yet sure there is a strong enough motivation to use the mbed OS type over the RTX type.

c1728p9:
Hi @Patater, I updated the PR to have a more detailed commit message. It might be that it doesn't make sense to have other back ends to uvisior. If this is the case I can close out this PR and update https://github.com/ARMmbed/mbed-os/pull/4776 instead. Just let me know.

Patater:
Seems weird that CMSIS doesn't already deal with this problem. It feels like "mbed_rtos_storage.h" should be "cmsis_rtos_storage.h" and part of upstream CMSIS.

c1728p9:
@bulislaw did you propose adding storage abstraction to cmsis-os?

bulislaw:
Yes, there was longer discussion about that with Robi and Reinhard and maybe Jonatan. They said that from RTOS/RTX point of view that's wasn't necessary as far as I remember.

Patater:
This depends on https://github.com/ARMmbed/uvisor/pull/471

Patater:
retest uvisor

Patater:
Thanks for the PR, but we're going to reject this until the time when we are ready to integrate with a non-RTX mbed OS backend.

PR #473, opened 2017-08-01
pool_queue: Additional sanity checks

fangyi-zhou:
Summary: the free slot in the pool should be completely disjoint with the enqueued slots in the queue. Adding additional checks allows the data structure to be more robust against invalid user operations/corrupted data structure.

- Checks whether a slot in enqueued before freeing
- Checks whether a slot is dequeued before enqueueing
- Checks whether the next slot is in bound before continuing the find
- Refactor dequeue checks


Patater:
Please rebase your changes on latest master (don't merge).

core/system/src/pool_queue.c
```diff
@@ -100,8 +100,9 @@ static int enqueue(uvisor_pool_queue_t * pool_queue, uvisor_pool_slot_t slot)
     }
     uvisor_pool_queue_entry_t * slot_entry = &pool->management_array[slot];
 
-    if (slot_entry->dequeued.state == UVISOR_POOL_SLOT_IS_FREE) {
-        /* Reject if a free slot is being enqueued */
+    if (slot_entry->dequeued.state != UVISOR_POOL_SLOT_IS_DEQUEUED) {
```
Patater: Consider checking for each state explicitly, in case more states are added in the future.

PR #476, opened 2017-08-07
Remove memory map references from uvisor core

### LINKED ISSUE #398: Peripheral is always 'not available' on ACL violation
According to [the docs](https://github.com/ARMmbed/uvisor/blob/master/docs/api/QUICKSTART.md#the-public-box-acls) I should see which peripheral is responsible for the fault, but this does not work anymore. It's always marked as 'not available'...

FYI, this is mbed-os-example-uvisor without SIM ACL on K64F:

```
***********************************************************
                    BUS FAULT
***********************************************************

* FAULT SYNDROME REGISTERS

  CFSR: 0x00008200
  BFAR: 0x40048044
  --> PRECISERR: precise data access.

* No MPU violation found

* MEMORY MAP
  Address:           0x40048044
  Region/Peripheral: [not available]
    Base address:    0x40048044
    End address:     0x40048044
    AIPS slot:       72
```

mikisch81:
Fix #398 

Remove any platform-specific memory-map references from uvisor core.

### DIFF
--- a/core/debug/src/core_armv7m/mpu_kinetis/debug_kinetis.c
+++ b/core/debug/src/core_armv7m/mpu_kinetis/debug_kinetis.c
@@ -20,31 +20,6 @@
 #include "memory_map.h"
 #include "vmpu_kinetis_map.h"
 
-static uint32_t debug_aips_slot_from_addr(uint32_t aips_addr)
-{
-    return ((aips_addr & 0xFFF000) >> 12);
-}
-
-static uint32_t debug_aips_bitn_from_alias(uint32_t alias, uint32_t aips_addr)
-{
-    uint32_t bitn;
-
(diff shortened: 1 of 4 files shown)


core/debug/src/core_armv7m/mpu_kinetis/debug_kinetis.c
```diff
@@ -95,48 +70,10 @@ static void debug_fault_mpu(void)
...
-{
-    uint32_t aips_addr;
-    const MemMap *map;
-
-    dprintf("* MEMORY MAP\n");
-    dprintf("  Address:           0x%08X\n", address);
-
-    /* find system memory region */
-    if((map = memory_map_name(address)) != NULL)
     {
```
Patater: You've added a `{` after the else. This is good, but should be done in a separate commit so that the changes are easier to read.

PR #478, opened 2017-08-08
Implement a STATIC_ASSERT macro

theamirocohen:
Implement STATIC_ASSERT macro to catch errors at compile time.


api/inc/uvisor_exports.h
```diff
@@ -78,10 +78,15 @@
  *
  * This macro works from both inside and outside function scope.
  *
- * FIXME This is currently not implemented. This issue is tracked at
- * https://github.com/ARMmbed/uvisor/issues/288
+ * The implementations differ due to compilation differences, c++ static_assert
+ * is known from c++11 while mbed-os compiles with c++98
  */
-#define UVISOR_STATIC_ASSERT(cond, msg)
+#ifdef  __cplusplus
```
Patater: We should use a different ifdef here.

Consider checking for C11 (`__STDC_VERSION`) first, and then falling back to other implementations.

api/inc/uvisor_exports.h
```diff
@@ -78,10 +78,15 @@
...
+#define UVISOR_STATIC_ASSERT(cond, msg) typedef char STATIC_ASSERT_##msg[(cond)?1:-1]
+#else
+#define UVISOR_STATIC_ASSERT(cond, msg) _Static_assert(cond, #msg)
+#endif/*__CPP__*/
```
Patater: Remove `/*__CPP__*/` comment. It's annoying to maintain and doesn't help much, since the ifdef is so small.

tools/uvisor-tests.txt
```diff
@@ -1 +1 @@
-0ac4c38729d2602bff96d68983a7790143210912
\ No newline at end of file
+19ac0f1047d9bb7223fb8854f531968a8fe55d7b
```
Patater: Is it necessary to update uvisor-test.txt? What modifications to uvisor-test do you need to support `UVISOR_STATIC_ASSERT`?

Patater:
The fallback implementation is not working when UVISOR_STATIC_ASSERT is used as a global statement, as in https://github.com/ARMmbed/uvisor/blob/bd3ec193ccfd30a6db28e6bcc10e570cc63f0132/api/inc/rpc_gateway.h#L47. Please revise the implementation so that it additionally works in this case.

alzix:
please squash the "fix code review" commit

api/inc/uvisor_exports.h
```diff
@@ -77,11 +77,18 @@
...
+ * is known from C++11 (__cplusplus > 199711L) while mbed-os compiles with c++98,
+ * and C files use _Static_assert. */
+#ifdef __cplusplus
+#if __cplusplus > 199711L
+#define UVISOR_STATIC_ASSERT(cond, msg) static_assert(cond, #msg)
+#else
+#define UVISOR_STATIC_ASSERT(cond, msg) typedef char STATIC_ASSERT_##msg[(cond)?1:-1]
+#endif
+#else
+#define UVISOR_STATIC_ASSERT(cond, msg) _Static_assert(cond, #msg)
```
Patater: Please use elseif to eliminate excessive nesting. Also, check that C is new enough for _Static_assert, and otherwise fallback to the typedef-based implementation.

api/inc/uvisor_exports.h
```diff
@@ -77,11 +77,19 @@
...
  *
  * This macro works from both inside and outside function scope.
- *
- * FIXME This is currently not implemented. This issue is tracked at
- * https://github.com/ARMmbed/uvisor/issues/288
- */
-#define UVISOR_STATIC_ASSERT(cond, msg)
+ * The implementations differ due to compilation differences, C++ static_assert
+ * is known from C++11 (__cplusplus > 199711L) while mbed-os compiles with c++98,
+ * and C _Static_assert is known grom GCC version 4.6.0. */
```
Patater:
```diff
- grom
+ from
```

PR #479, opened 2017-08-09
Update uvisor-tests.txt to latest

orenc17A:


orenc17A:
retest uvisor

Patater:
We should update only when we need to.

PR #485, opened 2017-08-14
Clear magic numbers from code

mikisch81:


Patater:
Thanks. This on the right track.

Please resolve test failures.

core/vmpu/src/vmpu.c
```diff
@@ -616,10 +616,12 @@ int vmpu_init_pre(void)
 
 void vmpu_init_post(void)
 {
+#if !defined(ARCH_MPU_ARMv8M)
```
mikisch81: According to the Cortex M33 user guide, there is no NONBASETHRDENA bit in SCB->CCR (this bit is reserved):
image

core/debug/src/debug.c
```diff
@@ -376,10 +376,12 @@ static void debug_fault_hard(void)
 
 void debug_init(void)
 {
+#if !defined(ARCH_MPU_ARMv8M)
```
mikisch81: According to the Cortex M33 user guide, there is no DISDEFWBUF bit in SCnSCB->ACTLR (this bit is reserved):
image

mikisch81:
@Patater @niklas-arm @alzix 
Can you take another look, seems that there were 2 register sets which are not right for v8m.

core/vmpu/src/vmpu.c
Patater: Woah. This is a big concern. If it turns out we can't do `NONBASETHRDENA` on M33, uVisor has to be reengineered a bit. We might come off lucky if `NONBASETHRDENA` is always effectively on by default, though.

core/debug/src/debug.c
Patater: Without this, I wonder if we still get good debug information from the MMFAR or SFAR.

core/vmpu/src/vmpu.c
Patater: Looks like the NONBASETHRDENA bit is always 1, so we are good to go.

mikisch81: But bit 0 is not NONBASETHRDENA in cortex m33, so how can you have this functionality in v8m?

Patater:
Merged manually. Thanks!

PR #491, opened 2017-08-16
ipc: Check the token of IO descriptor before processing

fangyi-zhou:
When control has been returned to user after IPC request. It is possible
that the user modifies the token. In Line 146/150, the token has been
read from the descriptor without checking. This may lead to token
corruption, e.g. some unallocated tokens get set to complete.


core/system/src/ipc.c
```diff
@@ -188,6 +188,18 @@ static bool ipc_io_is_ok(int box_id, const uvisor_ipc_io_t * io) {
            vmpu_buffer_access_is_ok(box_id, io->msg, io->desc->len);
 }
 
+static bool ipc_io_token_is_ok(uvisor_ipc_t * ipc, const uvisor_ipc_io_t * io) {
+    const uint32_t token = io->desc->token;
+    UvisorSpinlock * lock = &ipc->tokens_lock;
+    bool status = false;
```
thomasfangyi: Maybe add `__builtin_pops(token) == 1` as well?

core/system/src/ipc.c
```diff
@@ -317,6 +335,13 @@ void ipc_drain_queue(void)
             continue;
         }
 
+        if (!ipc_io_token_is_ok(recv_ipc, recv_io)) {
+            /* The token of the IO request is invalid. Ignore it, don't put it
+             * back. This shouldn't happen in a non-malicious box. */
+            put_it_back(send_queue, send_slot);
```
Patater: The comment says "don't put it back", but the code says `put_it_back`.

thomasfangyi: "Don't put it back" is for the receiving request. Will fix the comment at a later time (possibly those similar ones as well)

thomasfangyi: Updated comments

core/system/src/ipc.c
Patater: Yeah, an individual request should only use a single token. This function isn't used to check when waiting for multiple requests to complete.

fangyi-zhou:
retest uvisor

core/system/src/ipc.c
```diff
@@ -190,6 +190,21 @@ static bool ipc_io_is_ok(int box_id, const uvisor_ipc_io_t * io) {
...
 
+static bool ipc_io_token_is_ok(uvisor_ipc_t * ipc, const uvisor_ipc_io_t * io) {
+    const uint32_t token = io->desc->token;
+    if (__builtin_popcount(token) != 1) {
+        return false;
+    }
+    UvisorSpinlock * lock = &ipc->tokens_lock;
+    bool status = false;
+    if (!uvisor_spin_trylock(lock)) {
+        return status;
```
danny4478: Can we return false and declare and use status only when needed?

fangyi-zhou:
Token is taken and used from user space without check. A user can change the token in the descriptor and this change will affect uVisor. This breaks certain functionality.

However, considering the security requirement of IPC is that a box is free to mess up with itself, this fix is optional, as corrupted tokens should not affect other boxes.

core/system/src/ipc.c
```diff
@@ -313,8 +334,17 @@ void ipc_drain_queue(void)
...
+             * and don't put it back. Put the send IO back because it is not
+             * malicious. */
+            put_it_back(send_queue, send_slot);
+            continue;
+        }
+
+        if (!ipc_io_token_is_ok(recv_ipc, recv_io)) {
+            /* The token of the receive IO is invalid. Ignore it, don't put it
+             * back. This shouldn't happen in a non-malicious box. Put the send
+             * IO back because it is not malicious. */
```
danny4478: I still do not understand something. Correct me if I am wrong:
The message is still not delivered, so the allocated_tokens on the receiving side is still not updated, which should cause ipc_io_token_is_ok() to fail here.
Is that correct?

fangyi-zhou: I don't understand what you mean. `allocated_tokens` will be updated by the user side.

fangyi-zhou: The aim of this fix is to prevent user assigning a token *maliciously* to the descriptor, so that the token system might be corrupted due to the malicious token.

fangyi-zhou: Example sequence of events:

Suppose there is a valid receive request 

Non-malicious:
- User: IPC send request
- uVisor API mode: allocate token, enqueue request, return control
- User: read token, possibly call wait functions on the tokens
- uVisor core mode: ipc_drain_queue: deliver message

Malicious:
- User: IPC send request
- uVisor API mode: allocate token, enqueue request, return control
- User: `desc->token = 0xffffffff`
- uVisor core mode: ipc_drain_queue: deliver message, `completed_tokens &= desc->token`
Now `completed_tokens` are corrupted

fangyi-zhou: Again, this will only shoot the malicious box in the foot, so this fix is optional.

Patater:
Will drop the token checking commit, but keep the comments commit

PR #493, opened 2017-08-17
Add return statement in halt flow of register_gateway

mikisch81:
- In case there is a debug box HALT_ERROR doesn't halt but forges the
  debug_box stack, deprivileges and set the thread mode return address
  to the debug-driver's halt callback function.
- So after HALT_ERROR need to return in order to exit handler mode so
  the debug-driver's halt callback gets called.


Patater:
- Is there a test that catches this issue?
- If not, we should update the tests first and then update the test pointer here. Perhaps make a new "acl" test.

Patater:
If the code did continue out of the HALT, I'd expect to be able to test for that. But, that's not what happens. HALT doesn't return, whether in debug mode or not. So, this bug would be hard to see in the wild.

PR #495, opened 2017-08-24
Check pool queue sanity in ipc

fangyi-zhou:
It is unsafe to assume the pool queue provided by the user is safe, sanity check is necessary before any pool queue operations.


fangyi-zhou:
We could do checks for RPC pool queues as well.

api/inc/pool_queue_exports.h
```diff
@@ -105,6 +105,16 @@ UVISOR_EXTERN int uvisor_pool_init(uvisor_pool_t * pool, void * array, size_t st
  * Return 0 on success, non-zero otherwise. */
 UVISOR_EXTERN int uvisor_pool_queue_init(uvisor_pool_queue_t * pool_queue, uvisor_pool_t * pool, void * array, size_t stride, size_t num);
 
+/* Sanity check for a pool.
+ * Returns true for a sane pool initialised from given parameters, false
```
Patater: Please use international English.

```diff
- initialised
+ initialized
```

core/system/src/pool_queue.c
```diff
@@ -419,3 +419,160 @@ uvisor_pool_slot_t uvisor_pool_queue_find_first(uvisor_pool_queue_t * pool_queue
...
+        return false;
+    }
+
+    /* Check stride */
+    if (pool->stride != stride) {
+        return false;
+    }
+
+    /* Cannot check sanity if lock is held by others */
+    if (pool->spinlock.acquired) {
```
Patater: Should we use try_lock instead to avoid race conditions in reading the spinlock's acquired state?

core/system/src/pool_queue.c
```diff
@@ -419,3 +419,160 @@ uvisor_pool_slot_t uvisor_pool_queue_find_first(uvisor_pool_queue_t * pool_queue
...
+        return false;
+    }
+
+    /* Check pool pointer */
+    if (queue->pool != pool) {
+        return false;
+    }
+
+    /* Cannot check sanity if lock is held by others */
+    if (pool->spinlock.acquired) {
```
Patater: Consider using try lock instead.

core/system/src/ipc.c
```diff
@@ -214,6 +214,15 @@ void ipc_drain_queue(void)
         return;
     }
 
+    if (!uvisor_pool_queue_is_sane(send_queue,
+                                   send_queue->pool,
+                                   send_array,
+                                   sizeof(*send_array),
+                                   UVISOR_IPC_SEND_SLOTS)) {
+        /* This shouldn't happen in a non-malicious box. */
```
Patater: This can happen in a non-malicious box if the spinlock is held by user code. Consider locking around the outside of the sanity check instead of inside the sanity check, so you can tell the difference between insane data structures and not able to get the lock.

core/system/src/ipc.c
```diff
@@ -288,6 +297,16 @@ void ipc_drain_queue(void)
             continue; /* Try the next send IO. */
         }
 
+        if (!uvisor_pool_queue_is_sane(recv_queue,
+                                       recv_queue->pool,
+                                       recv_array,
+                                       sizeof(*recv_array),
+                                       UVISOR_IPC_RECV_SLOTS)) {
+            /* This shouldn't happen in a non-malicious box. */
```
Patater: Consider locking around the outside of the check, instead of inside. When locking inside, this _can_ happen in a non-malicious box.

Patater (changes requested):

core/system/src/pool_queue.c
fangyi-zhou: Shall we lock inside or outside?

api/inc/pool_queue_exports.h
fangyi-zhou: Fixed

core/system/src/ipc.c
fangyi-zhou: Locking outside would make the sanity check API exported incomplete. We could use different return values to indicate whether the data structure is corrupted or we cannot lock.

core/system/src/pool_queue.c
Patater: Outside seems best

core/system/src/ipc.c
Patater: Changing the return value is also acceptable.

fangyi-zhou: Any guidance on error code?

Patater: Add something in https://github.com/ARMmbed/uvisor/blob/master/api/inc/halt_exports.h, so we can use it from test code as well.

api/inc/pool_queue_exports.h
```diff
@@ -105,6 +105,16 @@ UVISOR_EXTERN int uvisor_pool_init(uvisor_pool_t * pool, void * array, size_t st
  * Return 0 on success, non-zero otherwise. */
 UVISOR_EXTERN int uvisor_pool_queue_init(uvisor_pool_queue_t * pool_queue, uvisor_pool_t * pool, void * array, size_t stride, size_t num);
 
+/* Sanity check a pool.
+ * Returns 0 for a sane pool initialized from given parameters, non-zero
```
Patater: non-zero error code otherwise

core/system/src/pool_queue.c
```diff
@@ -419,3 +420,170 @@ uvisor_pool_slot_t uvisor_pool_queue_find_first(uvisor_pool_queue_t * pool_queue
...
+    /* Check allocated count */
+    if (pool->num_allocated + num_free != num) {
+        return false;
+    }
+
+    return true;
+}
+
+int uvisor_pool_sanity_check(uvisor_pool_t * pool, void * array,
+                         size_t stride, size_t num)
```
Patater: Please ensure parameters are aligned when wrapped.

Patater (changes requested):

core/system/src/pool_queue.c
```diff
@@ -419,3 +420,170 @@ uvisor_pool_slot_t uvisor_pool_queue_find_first(uvisor_pool_queue_t * pool_queue
...
+}
+
+int uvisor_pool_queue_sanity_check(uvisor_pool_queue_t * queue,
+                               uvisor_pool_t * pool, void * array,
+                               size_t stride, size_t num)
+{
+    bool visited[num];
+    int status = UVISOR_ERROR_OUT_OF_STRUCTURES;
+    /* Cannot check sanity if lock is held by others */
+    if (!uvisor_spin_trylock(&queue->pool->spinlock)) {
```
Patater: It seems we are locking too early, over too much. Please lock minimally, just around the accesses or cluster of accesses where locking is needed.

core/system/src/pool_queue.c
fangyi-zhou: Fixed

api/inc/pool_queue_exports.h
fangyi-zhou: Fixed

core/system/src/pool_queue.c
fangyi-zhou: Locking will not be effective, as a malicious user can always modify the internals ignore the lock

Patater: The locking is to prevent us from seeing the pool or queue inconsistently and mistakenly judging it to be defective.

Patater:
This sort of checking is all very expensive. Are we checking the minimal amount to ensure that a box will be limited to only shooting itself in the foot and not uVisor itself or other boxes?

fangyi-zhou:
It's checking more than expected. I agree that the check is very costly.

Alternatively,
We could have a variable to act as a latch in `uvisor_ipc_t`. At IPC request, it is set to `true`, then `ipc_drain_queue` only takes place if set to `true`.

PR #498, opened 2017-08-30
Fix wrong usage of uvisor_pool_queue_try_enqueue

mikisch81:
In `drain_message_queue`, `uvisor_pool_queue_try_enqueue` return value was checked for true/false when this API actually returns the slot number.
This caused a false negative assertion which broke the number-store example.
@Patater @fangyi-zhou @alzix


fangyi-zhou:
Thanks for the catch. I think I forgot to change the second enqueue in my commit.

PR #501, opened 2017-09-01
ipc: Use secure alias when using user pointer in ipc_deliver

fangyi-zhou:
When copying the message from send box to receive box, the secure alias
should be used as the kernel accesses user data. Otherwise a BusFault
will occur when secure gating is not disabled.


Patater:
It actually appears to be a BusFault, and it only happens when `DISABLE_GATING` is not enabled.

fangyi-zhou:
Updated commit message

PR #502, opened 2017-09-01
	ipc: Return success immediately if waiting on empty token

fangyi-zhou:
Waiting on an empty token will bypass the allocated check because
bitwise and won't fail. It will then cause an infinite loop because the
condition will never fulfil.

Instead, waiting on nothing should return success immediately

Fixes 60048c4d3740350bd3254351a00dad0f367a580d

### DIFF
--- a/api/src/ipc.c
+++ b/api/src/ipc.c
@@ -144,7 +144,7 @@ static int ipc_waitfor(int (*cond)(uint32_t have, uint32_t expect), uint32_t wai
 /* We have at least one of what we expect. */
 static int any(uint32_t have, uint32_t expect)
 {
-    return have & expect;
+    return (expect == 0) || have & expect;
 }
 
 /* We have at least all of what we expect (and maybe more). */


api/src/ipc.c
```diff
@@ -109,6 +109,9 @@ static int ipc_waitfor(int (*cond)(uint32_t have, uint32_t expect), uint32_t wai
 {
     uint32_t completed_tokens;
     bool condition_met;
+    if (!wait_tokens) {
+        return UVISOR_ERROR_INVALID_PARAMETERS;
+    }
```
Patater: What makes returning an error better than returning success?

One could expect waiting for no tokens to always succeed, as all tokens you wanted to wait for are complete already.

fangyi-zhou: Makes sense. Fixed.

fangyi-zhou: oops. git fail

danny4478:
I think the real issue here is that the 2 functions, all() and any() have different behavior.
When the <expect> parameter is 0:
- any() returns 0.
- all() returns 1.
The correct fix in my opinion, is to fix any() to return 1 in case expect is 0.
@patater, what do you think?

api/src/ipc.c
```diff
@@ -109,6 +109,9 @@ static int ipc_waitfor(int (*cond)(uint32_t have, uint32_t expect), uint32_t wai
 {
     uint32_t completed_tokens;
     bool condition_met;
+    if (!wait_tokens) {
+        return 0;
```
Patater: Should we also set `*done_tokens` to 0?

fangyi-zhou: Yes. Fixed

Patater:
@danny4478 

I'd expect the return error code to be 0 (no error) for both any and all, if the `wait_tokens` are 0. I'd also expect `*done_tokens` set to 0 in both cases. What's the reason you'd expect a 1 returned from any?

Patater:
I had misunderstood which functions @danny4478 was referring to. It seems the best place to fix this is in the `any` function, not `ipc_waitforany`.

fangyi-zhou:
I've moved the check to `ipc_waitforany` and `ipc_waitforall` due to the need to return different values.

Patater:
Sorry for the misunderstanding.

I think what Danny was suggesting was that we only need to do the following.

```C
static int any(uint32_t have, uint32_t expect)
{
-    return have & expect;
+    return (expect == 0) || (have & expect);
}
```

fangyi-zhou:
The current implementation saves the time on the lock, but I'm open to change to what you proposed.

danny4478:
Currently wait_tokens of 0 gives different results for ipc_waitforany() and ipc_waitforall().
I don't think this is what we want.

The modification to any() function as @Patater wrote is indeed what I was suggesting, and I think is a must anyway.
I think it can come together with the original change @fangyi-zhou suggested in the first commit to save lock time, and together they make a full solution.

Patater:
I like Danny's solution. The lock time is not significant.

PR #504, opened 2017-09-03
WIP: uVisor stack guard for v7-M

danny4478:


core/linker/default.h
```diff
@@ -69,7 +77,7 @@ MEMORY
   FLASH_S  (r x) : ORIGIN = (FLASH_ORIGIN + SECURE_ALIAS_OFFSET + FLASH_OFFSET),
                    LENGTH = UVISOR_FLASH_LENGTH_MAX
   RAM_S    (rwx) : ORIGIN = (SRAM_ORIGIN + SECURE_ALIAS_OFFSET + SRAM_OFFSET),
-                   LENGTH = UVISOR_SRAM_LENGTH_USED - (STACK_SIZE + STACK_GUARD_BAND) - (STACK_SIZE_NP ? (STACK_SIZE_NP + STACK_GUARD_BAND) : 0)
+                   LENGTH = UVISOR_SRAM_LENGTH_USED - (STACK_SIZE + STACK_GUARD_BAND + STACK_GUARD_BAND) - (STACK_SIZE_NP ? (STACK_SIZE_NP + STACK_GUARD_BAND) : 0)
```
Patater: This line is a change to v8-M, since guard band is now counted twice.

To make this easier to read, create the length for v8 and v7 separately. v7 doesn't care about `STACK_SIZE_NP`.

core/linker/default.h
```diff
@@ -60,7 +64,11 @@ ENTRY(main_entry)
  *       stack from the rest of the protected memories. For this reason the
  *       symbol is arbitrarily small and cannot be defined by the
  *       platform-specific configurations. */
+#if defined (__ARM_FEATURE_CMSE) && (__ARM_FEATURE_CMSE == 3U)
 #define STACK_GUARD_BAND 32
+#else
+#define STACK_GUARD_BAND 512
+#endif
```
Patater: This means the stack guard band is now 512 on Kinetis. Probably not a change you want to make yet.

core/uvisor-config.h
```diff
@@ -40,7 +40,7 @@
  * is configuration-specific. See \ref UVISOR_SRAM_LENGTH_USED for more
  * information.
  */
-#define UVISOR_SRAM_LENGTH_PROTECTED 0x2000
```
Patater: This was 8 KiB. It's now 16 KiB? Is that really necessary?

danny4478: Actually it needs to be 0x3000, but then fail on:
#if (SRAM_ORIGIN % UVISOR_SRAM_LENGTH_PROTECTED) != 0
#error "The SRAM origin must be aligned to a multiple of UVISOR_SRAM_LENGTH_PROTECTED"
#endif

core/linker/default.h
danny4478: Shouldn't it being counted twice?
Before the change we had one guard which was counted also for v8-M.

danny4478: What we can do now is:
For v7-M: Use 2 guards of 512 bytes.
For v8-M: In order to keep unchanged, keep the 32 bytes guard (although not used) and remove it when dealing with v8-M.
Is that acceptable?

Patater: Sounds good

core/uvisor-config.h
Patater: Can we align it properly? Or is the assertion incorrect?

core/linker/default.h
```diff
@@ -36,17 +36,52 @@ ENTRY(main_entry)
...
+ *       In order to use SRD (sub region disabled) we set the guard to
+ *       be 1/8th of the total stack size */
+#define STACK_GUARD_BAND (TOTAL_STACK_SIZE >> 3)
+/* Default uVisor stack size
+ * Note: This is uVisor own stack, not the boxes' stack. */
+#define STACK_SIZE (TOTAL_STACK_SIZE - STACK_GUARD_BAND - STACK_GUARD_BAND)
+#else
 /* Default uVisor stack size
  * Note: This is uVisor own stack, not the boxes' stack. */
-#if !defined(STACK_SIZE)
```
mikisch81: If STACK_SIZE is defined then it is overriden now.

danny4478: AFAIU since STACK_SIZE can come only from uVisor core (and not from application), it is only the one who builds uVisor that can define the stack size.
Therefore I think it should be defined  only in the linker script so I have eliminated the "if !define..." line

core/linker/default.h
```diff
@@ -55,21 +90,14 @@ ENTRY(main_entry)
...
   FLASH_NS (r x) : ORIGIN = (FLASH_ORIGIN + FLASH_OFFSET),
                    LENGTH = UVISOR_FLASH_LENGTH_MAX
   FLASH_S  (r x) : ORIGIN = (FLASH_ORIGIN + SECURE_ALIAS_OFFSET + FLASH_OFFSET),
                    LENGTH = UVISOR_FLASH_LENGTH_MAX
   RAM_S    (rwx) : ORIGIN = (SRAM_ORIGIN + SECURE_ALIAS_OFFSET + SRAM_OFFSET),
-                   LENGTH = UVISOR_SRAM_LENGTH_USED - (STACK_SIZE + STACK_GUARD_BAND) - (STACK_SIZE_NP ? (STACK_SIZE_NP + STACK_GUARD_BAND) : 0)
+                   LENGTH = UVISOR_SRAM_LENGTH_USED - TOTAL_STACK_SIZE - (STACK_SIZE_NP ? (STACK_SIZE_NP + STACK_GUARD_BAND) : 0)
   STACK_S  (rw ) : ORIGIN = ORIGIN(RAM_S) + LENGTH(RAM_S),
                    LENGTH = UVISOR_SRAM_LENGTH_PROTECTED - LENGTH(RAM_S)
 }
```
danny4478: @Patater @alzix 
On this code STACK_S length equals to TOTAL_STACK_SIZE.
When filling the STACK_S with data (some lines below, in .stack section) we add alignments on top of TOTAL_STACK_SIZE and can actually exceed the STACK_S length we declared before.
We can subtract the alignment size from RAM_S length but then we end up with taking memory from RAM_S in favor of STACK_S.
Can you advise?

Patater: It makes sense to me to take memory from RAM_S to use for STACK_S.

core/linker/default.h
Patater: We still need to keep the ifdef here. See docs/core/PORTING.md. We can provide a default stack size here, but targets need to be able to override this value.

core/linker/default.h
danny4478: It means taking 0x1000 bytes from RAM which is a half of it.

On the other hand, we can leave the code as is since the linker protects us from overflowing memory sections. Thus, in case a problem occurs, we get notified by the linker.
Actually, now the align command does not add a single byte.

danny4478:
@mikisch81 @romkuz01 
Please review

alzix:
retest uvisor

core/linker/default.h
danny4478: From your mail I understood that we can it without the #ifdef.
Did I misunderstand you?

danny4478: I believe that not allowing to pre-define STACK_SIZE is the best solution.
Nevertheless, I added the #ifdef STACK_SIZE as you suggested.

core/linker/default.h
```diff
@@ -36,17 +36,72 @@ ENTRY(main_entry)
...
+* |        Guard        |       |
+* |  [STACK_GUARD_BAND] |       /
+* '---------------------'    --
+*/
+
+/* Default total stack size including guard bands
+ * Must be a power of 2 for MPU configurations.
+ * If STACK_SIZE is predefined, it should be defined so
+ * that TOTAL_STACK_SIZE will be calculated to a power of 2. */
+#if !defined(STACK_SIZE)
```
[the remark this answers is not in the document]
danny4478: But then the user does not know what is the actual net size of the stack.
That's why I prefer doing it the way I did.
So, it is your call. What do you prefer?

core/linker/default.h
danny4478: I am not sure this is possible since it has to be aligned with the application linker script.
If we switch between STACK_S and RAM_S then the application linker script will have to be changed.
@Patater @alzix Please check me up on this one

core/linker/default.h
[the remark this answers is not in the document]
Patater: They can know the actual stack size by looking at the map files or viewing the derivative `STACK_SIZE` during the porting process. This isn't a use case we need to make "easy". Porting involves a sophisticated sort of user.

We also have guard bands on the stack, so if this ends up being not enough in practice, it's caught.

core/linker/default.h
Patater: uVisor is just reserving a large chunk of RAM (that it may or may not use) from the application's perspective. So long as uVisor stays within that chunk, the application won't be affected.

danny4478: Is it indeed the case?

Inside STACK_S in **uvisor linker script** we reserve space but also have symbols, e.g. __uvisor_stack_start__.
Inside RAM_S in **uvisor linker script** we also import symbols, like .bss.

Inside **.uvisor.bss** section in **application linker script** we first have:
*__uvisor_bss_main_start = .;*
*KEEP(*(.keep.uvisor.bss.main))*
*__uvisor_bss_main_end = .;*
Then we have:
*__uvisor_bss_boxes_start = .;*
*KEEP(*(.keep.uvisor.bss.boxes))*
*__uvisor_bss_boxes_end = .;*
All these are going into RAM_S in the application.

So AFAIU, they must be aligned.
Am I missing so something?

Patater: The application linker script is making sure the space reserved is kept and properly communicated to uVisor.

The application has no knowledge of where uVisor keeps its stack, but it knows the uVisor stack must be within `__uvisor_bss_main_start` and `__uvisor_bss_main_end`. So long as uVisor places its stack within `__uvisor_bss_main_start` and `__uvisor_bss_main_end`, the application won't be affected.

danny4478: I don't understand what you suggest.
Do you suggest simplifying the calculation of STACK_S and RAM_S length?
If so, I would prefer doing this in a separate PR as it is not related to the stack protection we are adding.

Patater: At least a separate commit would be good. If it helps make the stack protection easier to add, it would fit just fine in this PR.

I'm fine with what you have as is now, and was just making an observation as to how the linker script could be improved.

docs/core/PORTING.md
```diff
@@ -214,7 +214,7 @@ This table lists symbols that you can specify.
 
 | Symbol                        | Description                    |
 |-------------------------------|--------------------------------|
-| `STACK_SIZE`                  | The size of uVisor's own stack |
+| `TOTAL_STACK_SIZE`            | The size of uVisor's own stack |
```
Patater:
```diff
- The size of uVisor's own stack
+ The size of uVisor's own stack, including guard bands
```

danny4478: Tests passed on both efm33gg & K64F

PR #506, opened 2017-09-06
armv8m: Fix an error to clear pending Systick

p03189:
To prevent SysTick exception becomes UNKNOWN from
both PENDSTSET and PENDSTCLR are written to one,
PENDNSVSET bit in the value which we are going to
setup has to be cleared.

### DIFF
--- a/core/system/src/core_armv8m/virq.c
+++ b/core/system/src/core_armv8m/virq.c
@@ -115,9 +115,9 @@ static void virq_load_systick_ns(uint8_t box_id)
     SysTick_NS->VAL = 0;
     SysTick_NS->CTRL = g_virq_system_exception_state[box_id].systick.periph.CTRL;
     if (g_virq_system_exception_state[box_id].systick.pending) {
-        SCB_NS->ICSR |= SCB_ICSR_PENDSTSET_Msk;
+        SCB_NS->ICSR = SCB_ICSR_PENDSTSET_Msk;
     } else {
-        SCB_NS->ICSR |= SCB_ICSR_PENDSTCLR_Msk;
+        SCB_NS->ICSR = SCB_ICSR_PENDSTCLR_Msk;
     }
     if (g_virq_system_exception_state[box_id].systick.active) {
         SCB_NS->SHCSR |= SCB_SHCSR_SYSTICKACT_Msk;


Patater:
Do you have a failing test case to reproduce the issue this attempts to fix?

p03189:
Yes, I have.

Patater:
Please explain in the commit message in more detail how the problem occurs and why it is significant.

Could you share your test case so we can regression test?

Thanks

p03189:
Hi @Patater,

I have updated my commit message, and this test is for your reference.
[uvisor_test](https://github.com/p03189/uvisor_test)

Patater:
Hi @p03189 

I couldn't reproduce a SecureFault with your example. I was only able to reproduce an out of memory error, due to Mbed OS not giving any backing memory to the RTX memory allocators (https://github.com/ARM-software/CMSIS_5/issues/177).

I had to change the example's main function (code below), and then it was able to run without any SecureFault for a couple hours.

Do you have a better example that reproduces the SecureFault?

Thanks

```C
int main(void) {

    //size_t count = 0;

    /* XXX Can't use osThreadNew, since Mbed OS doesn't support RTX allocators. */
#if RTX_APIS
    osThreadId_t result = osThreadNew((osThreadFunc_t)sub_thread, NULL, NULL);
    if (result == NULL) {
        printf("\r\n sub_thread creating is failed \r\n");
        for (;;);
    }
#else
    Thread *thread = new Thread(osPriorityNormal, 1024, NULL);
    if (thread == NULL) {
        printf("\r\n sub_thread creating is failed \r\n");
        for (;;);
    }
    thread->start(sub_thread);
#endif

    while (1) {
        printf("\r\n***** PUBLIC BOX *****\r\n");
    }

    return 0;
}
```

p03189:
Hi @Patater,

What will happen if you simultaneously write both  PENDSTSET and PENDSTCLR to one ? This patch is to clear pending systick_ns_irq regardless it is comes before virq_load_systick_ns() or not. The pending systick_ns_irq is handled after returning from SysTick_IRQn_Handler() in uVisor. Because I don't register any systick handler for secure box and VTOR_NS is unchanged, the pending systick_ns_irq goes to be handled by SysTick_Handler() in RTX5. After that, RTX5 try to store context to PSP in Sys_ContextSave(). That causes a secure fault because PSP is zero for secure box.

I think  you can reproduce this error by using gdb commands to set PENDSTSET to one before virq_load_systick_ns() and check ICSR if the pending SysTick_NS is cleared. For regression, I'm sorry that
I can't force a systick_ns_irq in a test.

Patater:
I like the patch. I think it's necessary to avoid programming the hardware into an ambiguous state, and I'll accept the patch for that reason.

However, I'm not following how this prevents the error you describe, where systick_ns_irq handling for RTX is in a secure box instead of the public box. It sounds like we need something a bit more to prevent SysTick_NS from going active (but ok to go pending) during execution of secure boxes that don't use SysTick_NS. Would you agree, or is this patch good enough to solve the problem in a way I'm not seeing?

p03189:
Hi @Patater,

According to your design, the active and pending state of SysTick_NS will be reset for next activated box, so I'm wondering why you think a pending systick_ns_irq for RTX is ok during execution of secure box. I guess what you wanna do is to provide a way for each secure box to register systick_ns_irq and switch the SysTick_NS handler in vector table during box switching process. This patch is to safely clear pending state for systick_ns_irq reset, not to prevent SysTick_NS from going active. If you need a policy to manage systick_ns_irq between boxes, it is better to create new patch for that.

Patater:
@p03189 How does this change alone fix your issue of the secure box handling SysTick_NS when it isn't supposed to if you also need a separate PR to prevent SysTick_NS from going active? It doesn't doesn't like we've completely gotten to the bottom of this bug, as there appear to be holes in our understanding.

p03189:
Hi @Patater,

You and I both don't need to prevent SysTick_NS from going active, we just save and reset SysTick_NS status so that it can be restored for the public box switched back next time. I found there is a bug when you reset SysTick_NS status, and I just fixed it. I don't need another PR, this patch is good enough to fix the bug in SysTick_NS status reset.

alzix:
Hi @p03189 ,
from what i see you are suggesting to reset the pending bit.
As per documentation 
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.100235_0002_00_en/pal1486148212346.html all writes of 0 to bits to ICSR register are ignored.
Thus the correct update of ICSR register should be performed with `=` rather then `|=` - this is true trough all the code.
The correct way of clearing Systick pending status is:
`SCB_NS->ICSR = SCB_ICSR_PENDSTCLR_Msk;`

p03189:
Hi @alzix ,

Your recommendation is good to me and It is verified on my target.
I have also updated my commit message, thanks for your help.

core/system/src/core_armv8m/virq.c
```diff
@@ -117,7 +117,7 @@ static void virq_load_systick_ns(uint8_t box_id)
     if (g_virq_system_exception_state[box_id].systick.pending) {
         SCB_NS->ICSR |= SCB_ICSR_PENDSTSET_Msk;
```
Patater: Might it be possible we set both `SCB_ICSR_PENDSTSET_Msk` and `SCB_ICSR_PENDSTCLR_Msk` with this write?

p03189: Hi @Patater 

I think you are right. patch is updated and verified.
Thank you.

Patater:
Thanks!

PR #507, opened 2017-09-07
Coverity work

mikisch81:
* Merge all of the infinite loops like `while (1)` to one function `uvisor_noreturn()`.
* Add a killpath model for the `uvisor_noreturn()` - Coverity will stop analyzing after this function.
  * This significantly reduces number of false positives in the analysis.

@Patater @alzix @danny4478 @orenc17A

### DIFF
--- a/core/uvisor.h
+++ b/core/uvisor.h
@@ -57,8 +57,11 @@ typedef void (*UnprivilegedBoxEntry)(void);
 #define assert(...)
 #else /*NDEBUG*/
 #define DPRINTF dprintf
-#define assert(x) if(!(x)){dprintf("HALTED(%s:%i): assert(%s)\n",\
-                                   __FILE__, __LINE__, #x);while(1);}
+#define assert(x) \
+    if (!(x)) { \
+        dprintf("HALTED(%s:%i): assert(%s)\n", __FILE__, __LINE__, #x); \
+        uvisor_noreturn(); \
+    }
 #endif/*NDEBUG*/
 
 #ifdef  CHANNEL_DEBUG
--- a/tools/coverity/models.c
+++ b/tools/coverity/models.c
@@ -1,0 +1,20 @@
+/*
+ * Copyright (c) 2017, ARM Limited, All Rights Reserved
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+void uvisor_noreturn(void) {
+    return __coverity_panic__();
+}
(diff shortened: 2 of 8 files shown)


mikisch81:
I will move the last commit to a new PR

core/uvisor.h
```diff
@@ -52,13 +52,22 @@
 /* unprivileged box as called by privileged code */
 typedef void (*UnprivilegedBoxEntry)(void);
 
+UVISOR_FORCEINLINE void uvisor_noreturn(void)
+{
+    volatile int var = 1;
+    while(var);
```
Patater: Consider marking this function as no return with `UVISOR_NORETURN`. This will allow the compiler to reason about the function and its use better.

Consider also using `for(;;)` instead of a spin on a volatile variable.

mikisch81: Why is there a preference to using `for(;;)` instead?
Using this method it will be easier to debug, you just need to change the variable on-the-fly to continue (instead of changing the code and re-compile).

alzix:
@Patater althouhg `__noreturn` is better from the compiler perspective, it limits our debug capabilities. 

function with `__noreturn` does not generate stack frame. 
from GCC documentation: "Do not assume that registers saved by the calling function are restored before calling the noreturn function."

In cases we would wish to break at this function, examine stack trace and optionally manually mitigate the error for debugging purpose.

All this will not be available with the suggested implementation. Do you have any 'pros' for it?

core/uvisor.h
Patater: One can also just change the PC.

Patater:
The debugging reasons seem convincing enough to not use `UVISOR_NORETURN`. The 'pros' would be better warnings about unused or unreachable code.

mikisch81:
@Patater @alzix 
So, what is the conclusion?

mikisch81:
@orenc17A FYI

alzix:
given the debugging considerations i would vote for volatile loop

tools/coverity/models.c
```diff
@@ -0,0 +1,3 @@
+void uvisor_noreturn(void) {
```
Patater: This needs our copyright header and license at the top.

mikisch81: of course..

Patater:
I'm fine with volatile loop, though it doesn't seem very necessary, really.

tools/coverity/models.c
Patater: Should start copyright from the year the file was born (only 2017).

Patater:
Did you want to replace "for(;;)" with `uvisor_noreturn` as well?

mikisch81:
@Patater squashed and rebased...
skp_4979

tools/coverity/models.c
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+void uvisor_noreturn(void) {
+    return __coverity_panic__();
```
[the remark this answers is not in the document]
mikisch81: @Patater 
No need to include header file for that.
`models.c` is used by `cov-make-library` which builds this model.
So this file is built by Coverity not by us.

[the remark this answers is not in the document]
Patater: OK

alzix:
Lgtm

PR #512, opened 2017-09-18
Goodies from Roman

Patater:
PR up for testing

### DIFF
--- a/core/debug/src/core_armv7m/mpu_armv7m/debug_armv7m.c
+++ b/core/debug/src/core_armv7m/mpu_armv7m/debug_armv7m.c
@@ -22,12 +22,12 @@
 static void debug_fault_mpu(void)
 {
     if (VMPU_SCB_MMFSR & 0x80) {
-        dprintf("* MPU FAULT\n\r");
+        dprintf("* MPU FAULT\n");
     }
     else {
-        dprintf("* No MPU violation found\n\r");
+        dprintf("* No MPU violation found\n");
     }
-    dprintf("\n\r");
+    dprintf("\n");
 }
 
 void debug_mpu_config(void)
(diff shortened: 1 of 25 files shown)


core/debug/src/core_armv7m/mpu_armv7m/debug_armv7m.c
```diff
@@ -36,22 +36,22 @@ void debug_mpu_config(void)
...
     int i;
 
-    dprintf("* MPU CONFIGURATION\n\r");
+    dprintf("* MPU CONFIGURATION\n");
 
     /* CTRL */
     ctrl = MPU->CTRL;
-    dprintf("\n\r");
-    dprintf("  Background region %s\n\r", ctrl & MPU_CTRL_PRIVDEFENA_Msk ?
+    default_putc('\n');
```
alzix: why are we calling `default_putc` directly? this way it will be harder for us to redirect/disaable logs.
I believe we should use the same macro/function for all prints

core/debug/src/core_armv7m/mpu_kinetis/debug_kinetis.c
```diff
@@ -47,20 +47,20 @@ static void debug_fault_mpu(void)
...
                         }
-                        dprintf("\n\r");
+                        default_putc('\n');
                     }
                 }
                 if(!found)
-                    dprintf("[none]\n\r");
-                dprintf("  Master port:      %d\n\r", (edr >> 4) & 0xF);
-                dprintf("  Error attribute:  %s %s (%s mode)\n\r",
+                    dprintf("[none]\n");
```
alzix: since we are modifying this line - can you consider dding brackets around it.

core/debug/src/core_armv7m/mpu_armv7m/debug_armv7m.c
Patater: Fixed to use `dprintf` instead of `default_putc`

core/debug/src/core_armv7m/mpu_kinetis/debug_kinetis.c
Patater: Fixed style in this file. It had a lot of style issues. It is very old code!

PR #516, opened 2017-09-26
Bad link is killing the docs

iriark01:
Please merge ASAP - the engine can't build this page without it


Patater:
It's more obvious where the link is going when laid out that way. I'm fine with this.

Manually merged after fix of commit description to follow commit style guidelines.

PR #517, opened 2017-09-27
Heading levels

iriark01:
Sorry again but we've got the wrong header levels. Please can you merge this ASAP so we can publish?


Patater:
Updated commit to match style guidelines and applied manually

PR #519, opened 2017-10-03
Set uVisor main stack guard protection for v8-M

theamirocohen:
@mikisch81 
@romkuz01 
please review

### DIFF
--- a/core/system/src/main.c
+++ b/core/system/src/main.c
@@ -76,11 +76,15 @@ UVISOR_NAKED void main_entry(uint32_t caller)
         "it    eq\n"
         "popeq {pc}\n"
 
-        /* Set the MSP. Since we are changing stacks we need to pop and re-push
+        /* Set the MSP and MSPLIM. Since we are changing stacks we need to pop and re-push
          * the lr value. */
         "pop   {r0}\n"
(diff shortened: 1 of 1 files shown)


core/vmpu/src/mpu_armv8m/vmpu_armv8m.c
```diff
@@ -304,6 +304,10 @@ void vmpu_arch_init(void)
                   (SCB_SHCSR_BUSFAULTENA_Msk) |
                   (SCB_SHCSR_MEMFAULTENA_Msk);
 
+    /* Set MSPLIM to the bottom of the stack to prevent stack overflow */
```
romkuz01: We need to consider setting MSP limit in core/system/src/main.c
Somewhere near the point where MSP is configured:

        /* Set the MSP. Since we are changing stacks we need to pop and re-push
         * the lr value. */
        "pop   {r0}\n"
        "ldr   r1, =__uvisor_stack_top__\n"
        "msr   MSP, r1\n"
        "push  {r0}\n"

theamirocohen: moved to main_entry function in main.c

PR #486, opened 2017-08-14
change behavior of vmpu_unpriv_access in v8

orenc17A:

PR #520, opened 2017-10-15
Add msplim_ns for box 0

theamirocohen:
requies changes from PR https://github.com/ARMmbed/mbed-os/pull/5321
and changes from https://github.com/ARMmbed/uvisor/pull/486 
and https://github.com/ARMmbed/uvisor/pull/496

### DIFF
--- a/api/src/uvisor-input.S
+++ b/api/src/uvisor-input.S
@@ -114,6 +114,9 @@ uvisor_config:
     /* Debug driver pointer */
     .long __uvisor_debug_driver
 
+    /* Stack limit for public box MSPLIM_NS, accessed by public_box_stack_limit */
+    .long __StackLimit
+
 /* uVisor mode of operation
  * Modes available: UVISOR_ENABLED, UVISOR_DISABLED, UVISOR_PERMISSIVE. */
 __uvisor_mode:
--- a/core/system/inc/linker.h
+++ b/core/system/inc/linker.h
@@ -121,6 +121,9 @@ typedef struct {
     UvisorLibHooks const * const lib_hooks;
 
     TUvisorDebugDriver const * const debug_driver;
+
+    /* Stack limit for public box MSPNS_LIM*/
+    uint32_t public_box_stack_limit;
 } UVISOR_PACKED UvisorConfig;
 
 extern UvisorConfig const __uvisor_config;
--- a/core/system/src/main.c
+++ b/core/system/src/main.c
@@ -161,6 +161,9 @@ void main_init(void)
     /* NS P stack pointer, for the RTOS and the uVisor-ns. */
     __TZ_set_MSP_NS(original_sp);
 
+    /* NS P limit, for the RTOS and the uVisor-ns. */
+    __TZ_set_MSPLIM_NS(__uvisor_config.public_box_stack_limit);
+
     /* S NP stack pointer, for the SDSs and the transition gateways. */
     __set_PSP((uint32_t) &__uvisor_stack_top_np__);
 #else /* defined(ARCH_CORE_ARMv8M) */


theamirocohen:
please review @danny4478 @alzix @romkuz01

api/src/uvisor-input.S
```diff
@@ -114,6 +114,9 @@ uvisor_config:
     /* Debug driver pointer */
     .long __uvisor_debug_driver
 
+    /* Stack limit for MSPNS_LIM */
```
romkuz01: Change MSPNS_LIM to MSPLIM_NS in the comment.

core/system/inc/linker.h
```diff
@@ -121,6 +121,9 @@ typedef struct {
     UvisorLibHooks const * const lib_hooks;
 
     TUvisorDebugDriver const * const debug_driver;
+
+    /* Stack limit for MSPNS_LIM */
```
romkuz01: Change MSPNS_LIM to MSPLIM_NS in the comment.

api/src/uvisor-input.S
theamirocohen: fix

core/system/inc/linker.h
theamirocohen: fix

api/src/uvisor-input.S
```diff
@@ -114,6 +114,9 @@ uvisor_config:
     /* Debug driver pointer */
     .long __uvisor_debug_driver
 
+    /* Stack limit for MSPLIM_NS */
+    .long __uvisor_stack_limit
```
alzix: IMHO the variable name is misleading as it holds not the uvisor stack limit but the public box stack limit.
Am i missing something?

core/system/inc/linker.h
```diff
@@ -121,6 +121,9 @@ typedef struct {
     UvisorLibHooks const * const lib_hooks;
 
     TUvisorDebugDriver const * const debug_driver;
+
+    /* Stack limit for MSPLIM_NS */
+    uint32_t * stack_limit;
```
alzix: 1. same here what stack limit is it used for?
2. Additional question - what will we do with this new field on ARMv7-M platforms?
3. Please note the naming conventions here - members are named exactly as in the `uvisor-input.S` except `__uvisor_` prefix

api/src/uvisor-input.S
```diff
@@ -114,6 +114,9 @@ uvisor_config:
...
+
```
alzix: i see you are missing the assignment here as you have assigned the value to the `__uvisor_stack_limit` from the mbed-os linker script.
Please consider assigning __StackLimit__ symbol directly without intermediate variable and without modifying mbed-os sources.

api/src/uvisor-input.S
theamirocohen: fixed

alzix:
@Patater - please approve

api/src/uvisor-input.S
```diff
@@ -114,6 +114,9 @@ uvisor_config:
     /* Debug driver pointer */
     .long __uvisor_debug_driver
 
+    /* Stack limit for publix box MSPLIM_NS, accessed by public_box_stack_limit */
```
Patater:
```diff
- publix
+ public
```

api/src/uvisor-input.S
```diff
@@ -134,6 +137,9 @@ __uvisor_debug_driver:
     .long 0
     .long 0
 
+__uvisor_public_box_stack_limit:
+    .long __StackLimit
```
Patater: Why do we need the public box stack limit twice? Why not just put `__StackLimit` where `__uvisor_public_box_stack_limit` is instead?

theamirocohen: We feel that the var name __StackLimit is confusing while having 4 different stucks, and we wish to keep uvisor-input and linker.h variables the same

api/src/uvisor-input.S
theamirocohen: fixed

PR #524, opened 2017-10-26
Fix uVisor stack memory region size in uVisor linker script

alzix:
Wrong define (`UVISOR_SRAM_LENGTH_PROTECTED` instead of `UVISOR_SRAM_LENGTH_USED `) was used for calculating uVisor STACK_S memory region size in the linker script.
As a result STACK_S region was bigger than expected by SRAM_OFFSET, which i platform dependent value.
Since uVisor should fit in to a fixed memory region (defined in `uvisor-input.S`) this mistake could potentially lead to stack corruption.

@orenc17A @Patater could you please take a look


core/linker/default.h
```diff
@@ -100,21 +100,35 @@ ENTRY(main_entry)
...
+#define TOTAL_STACK_SIZE_NP 0
 #endif /* defined (__ARM_FEATURE_CMSE) && (__ARM_FEATURE_CMSE == 3U) */
 #endif /* !defined(STACK_SIZE_NP) */
 
+/*
+ *   uVisor SRAM map:
+ * 
+ *                       |-------------UVISOR_SRAM_LENGTH_USED-------------------------|
+ *                       |                                                             |
+ *    |------------------+----------UVISOR_SRAM_LENGTH_PROTECTED-----------------------|
```
danny4478: The length of PROTECTED is not clear by this drawing.

core/linker/default.h
```diff
@@ -100,21 +100,35 @@ ENTRY(main_entry)
 #if !defined(STACK_SIZE_NP)
 #if defined (__ARM_FEATURE_CMSE) && (__ARM_FEATURE_CMSE == 3U)
 #define STACK_SIZE_NP 256 /* TODO: choose better default. */
+#define TOTAL_STACK_SIZE_NP STACK_SIZE_NP + STACK_GUARD_BAND
```
danny4478: Put the define inside () :  (STACK_SIZE_NP + STACK_GUARD_BAND)

core/linker/default.h
orenc17A: add a comment referencing the platform

alzix: @danny4478 modified - please check again

core/linker/default.h
alzix: done

core/linker/default.h
alzix: @orenc17A the scheme remains valid for all platforms - the sizes of the regions may vary - 
 `TOTAL_STACK_SIZE_NP`  equals `0` for ARMv7-M

danny4478 (approved):

Patater (approved):
Code looks fine. Commit organization is good. Commit style could be improved, but not a blocker.

